PR #8, opened 2017-01-05
Add optional Serde serialization support

zaeleus:
This adds the Serde serialization trait to the following:

  * `ExecutionError`
  * `GraphQLError`
  * `InputValue` (+ deserialization)
  * `RuleError`
  * `SourcePosition`
  * `Spanning<ParseError>`
  * `Value`

Serde support is enabled via the "serde" feature flag.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ expose-test-schema = []
 [dependencies]
 rustc-serialize = "^0.3.19"
 iron = { version = "^0.4.0", optional = true }
+serde = { version = "^0.8.21", optional = true }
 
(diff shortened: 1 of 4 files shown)


mhallin:
Thanks!

A nice next step after this would probably be to make `rustc_serialize` optional too, to avoid Serde users having to link to two serialization libraries. Except for in the Iron handler, I _think_ JSON serialization only is used by the `defaultValue` meta field, which _probably_ could use something like the `Debug` trait instead.

If you want, you can open a PR and add yourself to the `authors` list in Cargo.toml :)

PR #14, opened 2017-01-11
Add type alias for query variables

zaeleus:


zaeleus:
Rebased master to fix conflicts.

PR #26, opened 2017-01-27
Accept explicit null input values in queries

### LINKED ISSUE #23: Nullable Fields For Input Objects Should  Accept Null As A Value.
As suggested in [3.18](http://facebook.github.io/graphql/#sec-Types.Non-Null)  nullable types can both accept null and are optional. Furthermore, the table found in [Input Object](http://facebook.github.io/graphql/#sec-Input-Objects) section also says this as well. 

For example, I have the mutation:
 
```
mutation {
  createRma(rmaInput:{
          id: "test123"
	  status: false
	  companyName: null
  })
}
```

Represented in rust like so: 

```rust
graphql_input_object!(
     pub struct RmaInput {
         id: String,
         status: bool,
         company_name: Option<String>
    }
);
```

I get an error message that says "Unexpected \\"null\\""

While understand that having this field added is somewhat wasted typing, but it would be nice to explicitly state that is should be null.

zaeleus:
Explicit `null` values are valid for nullable input values. See
<https://facebook.github.io/graphql/#sec-Types.Non-Null>.

Fixes #23

### DIFF
--- a/src/executor_tests/variables.rs
+++ b/src/executor_tests/variables.rs
@@ -339,6 +339,17 @@ fn allow_nullable_inputs_to_be_omitted_in_variable() {
 }
 
 #[test]
+fn allow_nullable_inputs_to_be_explicitly_null() {
+    run_query(
+        r#"{ fieldWithNullableStringInput(input: null) }"#,
+        |result| {
+            assert_eq!(
+                result.get("fieldWithNullableStringInput"),
+                Some(&Value::string(r#"None"#)));
+        });
+}
+
+#[test]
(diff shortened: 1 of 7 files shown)


mhallin:
Nice!

This is probably part of an effort to support the "October 2016" GraphQL specification (https://github.com/facebook/graphql/releases/tag/October2016) where the `null` literal was introduced. I wonder if we're still compatible with the revised/clarified execution rules as outlined in https://github.com/facebook/graphql/pull/221 - I haven't read that in detail yet.

PR #43, opened 2017-04-25
Drop Rustc Serialization Support

### LINKED ISSUE #36: Drop rustc-serialize support
According to [Rust release milestone predictions](https://internals.rust-lang.org/t/rust-release-milestone-predictions/4591?u=dtolnay), rustc-serialize is scheduled to be deprecated in next month's beta release of Rust.

This would be a good time to let us know if there is any functionality you require from Serde before that happens.

TheServerAsterisk:
This will close #36. 

One notable change is the `GraphQlIronError` enum I added. The purpose of this enum was to capture any serde or IO errors and log the error instead of making the thread panic if serde failed or the user failed to put in a valid JSON object.


src/integrations/iron_handlers.rs
```diff
@@ -7,11 +7,19 @@ use iron::status;
...
+use std::fmt;
+use std::boxed::Box;
 
-use rustc_serialize::json::{ToJson, Json};
+use serde_json;
+use serde_json::Value as Json;
+use serde_json::error::Error as SerdeError;
 
-use ::{InputValue, GraphQLType, RootNode, Variables, execute};
+use ::{InputValue, GraphQLType, RootNode, Variables, execute as execute_query};
```
TheServerAsterisk: Changed the name of this in order to clarify the difference between the execute function that exists in the iron_handlers file vs the execute function that resides in the lib.rs file.

src/integrations/iron_handlers.rs
```diff
@@ -77,59 +85,77 @@ impl<'a, CtxFactory, Query, Mutation, CtxT>
...
+        let result = execute_query(query, None, &self.root_node, variables, &context);
         let content_type = "application/json".parse::<Mime>().unwrap();
-        let mut map = BTreeMap::new();
+        let mut map = HashMap::new();
 
         match result {
             Ok((result, errors)) => {
-                map.insert("data".to_owned(), result.to_json());
+                let response_data = serde_json::to_value(result)
+                    .expect("Failed to convert response data to JSON.");
```
TheServerAsterisk: Wasn't sure how you, @mhallin, wanted handed handle errors at this points especially since the response has probably been parsed already and should only fail if the developer messes up.

src/integrations/serde.rs
```diff
@@ -1,13 +1,39 @@
...
 use parser::{ParseError, Spanning, SourcePosition};
 use validation::RuleError;
+use serde_json::Value as Json;
+
+
+impl InputValue {
+    /// Converts serde_json::Value to juniper::InputValue
+    pub fn from_json(json: Json) -> InputValue {
+        match json {
+          Json::Number(num) => {
```
TheServerAsterisk: Since Serde doesn't have I64 etc but just a general number it is unclear whether u64 should be converted to an i64 as well.  Please note if you want me to handle this case differently.

src/integrations/iron_handlers.rs
```diff
@@ -77,59 +85,77 @@ impl<'a, CtxFactory, Query, Mutation, CtxT>
...
+        match json_data {
+            Json::Object(json_obj) => {
+                let mut query = None;
+                let mut variables = Variables::new();
+
+                for (k, v) in json_obj {
+                    if k == "query" {
+                        query = serde_json::to_string(&v)
+                            .ok()
+                            .map(|query| query.replace("\"", ""));
```
TheServerAsterisk: It appears that sede_json::to_string will convert the query value to something like this. "\"{hero{name}}\""

mhallin: This sounds like a case of double encoding - what happens if you send `{droid(id: "1001"){name}}`?

src/integrations/iron_handlers.rs
mhallin: Yeah, if things fail here the only way out is to fail - whether through a clean exit or a panic which Iron will handle anyway doesn't really matter.

src/integrations/serde.rs
mhallin: I think we've got two options: either add `u64` to `InputValue` and `Value`, or convert it to either `i64` or `f64` if it's larger than `i64::MAX`.

BTW, I think it would be cleaner if this code read like:

```rust
if let Some(num) = num.as_i64() {
    InputValue::int(num)
}
else if let Some(num) = num.as_f64() {
    InputValue::float(num)
}
```

TheServerAsterisk:
@mhallin I made the appropriate changes, let me know if you want anythings else to be altered.

src/integrations/iron_handlers.rs
```diff
@@ -77,59 +85,74 @@ impl<'a, CtxFactory, Query, Mutation, CtxT>
...
-
-        let mut query = None;
-        let mut variables = Variables::new();
-
-        for (k, v) in json_obj {
-            if k == "query" {
-                query = v.as_string().map(|s| s.to_owned());
+        let mut request_payload = String::new();
+        itry!(req.body.read_to_string(&mut request_payload));
+        let json_data =
```
zaeleus: This section can be simplified by using serde's derive macro for deserialization instead of manually pulling out values, e.g., https://github.com/zaeleus/juniper/blob/a321204cfc8d378fe3d8d9af2c817476cba1211d/examples/rocket/src/graphql.rs#L32-L67

src/integrations/serde.rs
```diff
@@ -1,13 +1,42 @@
...
 use ast::InputValue;
 use executor::ExecutionError;
 use parser::{ParseError, Spanning, SourcePosition};
 use validation::RuleError;
+use serde_json::Value as Json;
+
+
+impl InputValue {
+    /// Converts serde_json::Value to juniper::InputValue
+    pub fn from_json(json: Json) -> InputValue {
```
zaeleus: [Deserialization is already implemented for `InputValue`](https://github.com/mhallin/juniper/blob/032b03add334f230854223bb69dcde1568ecb075/src/integrations/serde.rs#L52-L119). Why not use it?

src/integrations/iron_handlers.rs
TheServerAsterisk: I also ended up taking your WrappedGraphQlResult idea and placed both it and the GraphQlQuery into the serde module. Let me know if you have any more suggestions on how to improve the code.

TheServerAsterisk:
Ahhh, I forgot serde_derive is only support for rust 1.15 and greater. So I can either drop serde_derive as a dependency and manually implement serialization or support for rust 1.13 and rust 1.14 can be dropped. If the former let me know @mhallin and I will make the appropriate changes.

theduke:
I'd vote for just dropping older Rust versions.

mhallin:
Yeah, I think this is where we drop 1.14 and older :) The reason I let Travis test on older Rust versions is to not _accidentally_ dropping support for older versions.

norcalli:
From my reading, this seems to be ready. And this would fix Travis failing for the serde 1.0 PR as well, since that was only failing on <=1.14. If this PR is left stale for too long, forks may appear that will cause fragmentation of the codebase.

PR #44, opened 2017-04-29
Remove panic in Parser::next

Mrmaxmeier:
Hey,
I noticed this panic while passing invalid GraphQL queries to `juniper`.
I've moved it to the `ParseError::UnexpectedEndOfFile` error instead of panicking.


mhallin:
Neat! This was a case I thought was unreachable :) Do you have an example of a query that panics the current implementation?

Mrmaxmeier:
The queries I tried were `"mutation"` and `"query"`.
I've just created a fuzz target for [cargo-fuzz](https://github.com/rust-fuzz/cargo-fuzz) and found some other interesting ones:
- `""`
- `"\t"`
- `"{..."`
- `"))))"`
- `"$ \""`

mhallin:
Nice finds, thanks a lot.

PR #55, opened 2017-06-24
Restructure repo, add custom derive and integration tests.

theduke:
This PR handles several things:

* it restrucutres the repo into a Cargo workspace to allow multiple crates (also enables extracting parser)
* it adds a juniper_codegen module with custom_derive ( enums and input_objects)
* it adds a juniper_tests crate with integration tests between the crates

I realize this is quite a large PR. 
I can split it off into one that re-structures the repo and one for the derives and tests, but there is not much development going on anyway so I thought let's just do one.

Derive  works like this:

```rust
extern crate juniper;
#[macro_use] extern crate juniper_codegen;

#[derive(GraphQLEnum)]
// optional: #[graphql(name="MyBrand")]
enum Brand {
  Bmw,
  #[graphql(name="FORD", description="hmm..", deprecated=" not cool anymore...")]
  Ford,
}

#[derive(GraphQLInputObject)]
// optional: #[graphql(name="MyData")]
struct Data {
  // camel case by default.
  my_value: i32,
  // Default can be any valid rust expression inside a string.
  #[graphql(name="someField", default="false")]
  no_good_name: bool,
}

```

Maybe `juniper` might be a better choice for the attribute name?

Also, I plan on adding more derives:

* Input objects: much nicer as a derive, since there is no custom resolver code anyway
* objects: very useful if no custom resolving is going on

I think custom derive is much more iditiomatic for those cases than the macros.

I would also suggest deprecating the enum (and input_object macros once I add it), since they don't offer any additional utility and are more cryptic.

I called the derive crate `codegen` rather than derive since it could contain more than just custom_derive stuff.

For example, once proc_macros land, it could implement support for GraphlQL schemas.


theduke:
Note: there is one test failing because `cargo test --all --features XX` does not seem to propagate the features.

mhallin:
This looks very nice, thanks for all the work :D

As Rusts attribute macro system gets more powerful, I hope we can move away from the `graphql_*!` macros completely - they feel like a stopgap solution for now. This PR sets us up for this nicely.

I'll take a look at the CI builds and try to get them working before merging this.

theduke:
@mhallin basically I just took the easy path of adding `--all` to cargo build/test.

To work with the features the CI specs need to specify seperate builds/test commands for each crate.

theduke:
@mhallin I got the tests to pass.

I had to move the examples into the juniper directory for now.
I'd recommend moving them back to the root dir and making them into a separate crate, but maybe that can be done in a separate PR.

You should probably merge the other PRs first though so they don't have to be rebased.

theduke:
@mhallin any thoughts on how to move this forward?

PR #57, opened 2017-07-03
Avoid boxing in MultiVisitor

srijs:
Get rid of all the boxing/heap allocations in `visit_all_rules`, and rely on a stack-allocated heterogenous linked list instead.

In my small GraphQL application (use-case is a high volume of rather simple queries), this brings me from about 17k to 20k rps. Working on getting some more detailed and relevant numbers, in the meantime I thought I'd already put this up for review.

Let me know what you think!

### DIFF
--- a/src/validation/mod.rs
+++ b/src/validation/mod.rs
@@ -14,7 +14,7 @@ pub use self::traits::Visitor;
 pub use self::visitor::visit;
 pub use self::context::{RuleError, ValidatorContext};
 pub use self::rules::visit_all_rules;
-pub use self::multi_visitor::MultiVisitor;
+pub use self::multi_visitor::{MultiVisitor, MultiVisitorNil};
 pub use self::input_value::validate_input_values;
 
 #[cfg(test)]
--- a/src/validation/multi_visitor.rs
+++ b/src/validation/multi_visitor.rs
@@ -4,26 +4,32 @@ use parser::Spanning;
 use validation::{ValidatorContext, Visitor};
 
 #[doc(hidden)]
-pub struct MultiVisitor<'a> {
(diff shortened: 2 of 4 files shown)


srijs:
As promised, here are some more detailed numbers.

Without the patch:
```
Running 30s test @ http://localhost:3000/graphql
  2 threads and 10 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   585.81us  405.66us  23.94ms   95.20%
    Req/Sec     8.51k     1.61k   12.04k    56.31%
  Latency Distribution
     50%  533.00us
     75%  692.00us
     90%  848.00us
     99%    1.23ms
  510122 requests in 30.10s, 70.54MB read
Requests/sec:  16947.29
Transfer/sec:      2.34MB
```

With the patch:
```
Running 30s test @ http://localhost:3000/graphql
  2 threads and 10 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   537.27us  219.46us   8.27ms   75.96%
    Req/Sec     9.14k     1.89k   12.84k    53.41%
  Latency Distribution
     50%  493.00us
     75%  655.00us
     90%  806.00us
     99%    1.17ms
  546657 requests in 30.10s, 75.59MB read
Requests/sec:  18161.19
Transfer/sec:      2.51MB
```

mhallin:
Nice, thanks!

As a comment on all these performance PRs; I think adding an LRU cache from Hash(query) to AST(query) for validated queries would be a great improvement for performance - parsing and validating is probably 90% of Juniper's overhead.

I've been thinking of implementing that, but it would require the API to expose some kind of global state object, or use `lazy_static!`. It might also require exposing some configuration parameters.

PR #76, opened 2017-08-07
Use Cow version of to_camel_case in codegen

srijs:
Porting over the improvements I made in #58 to the codegen module.


theduke:
TBH, I don't think it's worth complicating the code in codegen here.

This is only run once during compile time, so the extra allocation is pretty much a non-issue.

srijs:
Yep, looks like I've misinterpreted the usage of it in codegen. Agreed with your assessment. Closing.

PR #78, opened 2017-08-08
Reject types with invalid names

### LINKED ISSUE #72: Registry should reject types with invalid names
While the spec doesn't explicitly specify what a valid name of a type is, it's reasonable to assume that they should be valid GraphQL name tokens, i.e. `/[_A-Za-z][_0-9A-Za-z]*/` since they're written as identifiers in e.g. fragment type conditions. This seems to be in line with what the [reference implementation](https://github.com/graphql/graphql-js/blob/7ddd8dd44cb7f8a3a9f7a760d5ab76ac91802931/src/utilities/assertValidName.js) does. This causes tools like GraphiQL to stop working if an invalid type name is exposed.

Since Juniper's macros use `stringify!`, invalid names can be generated by using e.g. multiple modules:

```rust
// Generates a GraphQL type named "auth::User"
graphql_object!(auth::User: () |&self| {
});
```

It's probably not possible to do this validation check at compile time, but the registry could probably panic on invalid names.

srijs:
Sets up some basic infrastructure for name validation and hooks it up to the registry.

In terms of error messages I tried to stick to the [reference implementation](https://github.com/graphql/graphql-js/blob/7ddd8dd44cb7f8a3a9f7a760d5ab76ac91802931/src/utilities/assertValidName.js#L53).

As a next step we could explore if/how to leverage `Name::is_valid` from the codegen to provide compile-time errors for invalid names, perhaps via the upcoming [`compile_error!`](https://github.com/sgrif/rfcs/blob/sg-add-error-macro/text/0000-add-error-macro.md) macro, or simply via `panic!`-ing (would love your guidance on that, @theduke).

Fixes #72

### DIFF
--- a/juniper/src/executor.rs
+++ b/juniper/src/executor.rs
@@ -13,6 +13,7 @@ use schema::meta::{Argument, EnumMeta, EnumValue, Field, InputObjectMeta, Interf
 use schema::model::{RootNode, SchemaType};
 
 use types::base::GraphQLType;
+use types::name::Name;
 
 /// A type registry used to build schemas
 ///
@@ -21,7 +22,7 @@ use types::base::GraphQLType;
 /// into `Type` instances and automatically registers them.
 pub struct Registry<'r> {
     /// Currently registered types
-    pub types: HashMap<String, MetaType<'r>>,
+    pub types: HashMap<Name, MetaType<'r>>,
 }
 
 #[derive(Clone)]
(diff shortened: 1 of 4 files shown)


mhallin:
Looks great, thanks!

theduke:
@srijs thanks for this.

I think we should panic on invalid names, both in the macros and in the custom derives.

PR #79, opened 2017-08-08
Proposal: Logo

srijs:
I feel like something like that is still missing in the project 😉 

Let me know what you think!

rust-grapqhql


theduke:
A blend of the Rust and GraphQL logos, I love it!

Thanks for the work!

Two ideas and two  notes:

* Could you maybe do an alternative version where there is a bit of a gap between the GraphQL logo and the Rust gears? So make that Graphql part a bit smaller or the gear a bit larger. Not sure if that looks better, but it might.

* Maybe keep the gear in a rusty brown? (again, not sure if that would look better, but some variations to compare would be cool

---

* We need to check if the GraphQL logo is under any license or trademark, and if it's permissible to re-use it for this kind of derived work

* It would be nice to have an accompanying "Juniper" text in a matching font and color. Maybe we can look for a nice font to go with it.

srijs:
Hi!

In the following some variations on the logo along the lines of your suggestions. I've also put a simple word mark next to it, which currently uses the same font as is used in the GraphQL logo. We can iterate more on the font once we've settled for a pictorial mark.

Regarding licensing, I've already listed the two sources + licenses in the `README.md` that accompanies this PR (Rust logo under CC-BY and GraphQL logo under 3-clause BSD), but you're right that we need to expand on the attribution a bit to include a proper copyright for both. IANAL, but my understanding is that as a derivative work we are obligated to mention the license + proper attribution, but not much beyond that.

rust-grapqhql

theduke:
I share your opinion, I'd favor the second one from the right. Nice work!

What do you think, @mhallin ?

mhallin:
Yeah, the first one from the left and the second one from the right are my favorites. Thanks for the work, @srijs! :D

This will go well in hand with a `graphql-rust.github.io` site with examples and guides :)

srijs:
Great, I'll update the PR with variants for light and dark backgrounds.

How do you feel about the font? Should we stick to this one for now?

Also, while we're on the topic of branding, I'm a little confused as of lately -- Is the project name still `juniper`, or are you transitioning to the `graphql-rust` name?

theduke:
I'm fine with the font.

Regarding the name: the project is still called juniper, the organization name `graphql-rust` is just to have an umbrella organization for GraphQL related crates.

So Juniper it is! ;)

theduke:
@srijs do you have time to update the PR with the Logo we have settled on?

Then I could merge it and include the logo in the README.

srijs:
I'll try to get to that on the weekend! Apologies for the delay here :/

srijs:
@mhallin @theduke Updated the PR with the logo we decided on, as well as variants for dark backgrounds and monochrome!

theduke:
@srijs awesome, thank you so much!

theduke:
@srijs logo is now in the README ;)

Issue #69, opened 2017-07-29
Rename ResultExt::to_field_err?


theduke:
The method name [ResultExt::fo_field_err](https://docs.rs/juniper/0.8.1/juniper/trait.ResultExt.html#tymethod.to_field_err) is a bit confusing, since it implies an Err return type.

We could leave it as is, or rename it to `to_field_result` or `to_field_res`.

If we rename, it should definitely be before 1.0.

Thoughts, @mhallin ?

mhallin:
Sounds reasonable, but while we're on the topic of field errors... I feel that error handling is a bit unergonomic right now. For example, if you're doing stuff that can fail, you've got two options in practice:

* Import `juniper::ResultExt`, write e.g. `let txn = db.transaction().to_field_result()?;`
* Use `jtry!`, e.g. `let txn = jtry(db.transaction());`

Converting error types yourself feels like a shortcoming on Juniper's side; why should a user have to care about that? On the other hand, using the macro looks a bit weird when you're chaining operations that can fail: `jtry!(jtry!(File.open(...)).read(...))`.

So, I wonder if we could make the macros support any kind of `Result<T, E>` (or at least `E: Display`) and call `.to_field_result()` on the result automatically? Then we could maybe even remove the `ResultExt` trait and `jtry!` macro.

srijs:
IIRC both the `try` macro as well as the `?` operator provide type upcasting: https://github.com/rust-lang/rfcs/blob/master/text/0243-trait-based-exception-handling.md#exception-type-upcasting.

So an alternative could be to define a concrete type `FieldError` implementing `From`:

```
struct FieldError {
  message: String
}

type FieldResult<T> = Result<T, FieldError>;

impl<E> From<E> for FieldError where E: Display {
  fn from(err: E) -> FieldError {
    FieldError { message: err }
  }
}
```

This would enable people to use the `?` operator without `to_field_result` in field bodies:

```
field foo()  {
  let txn = db.transaction()?;
  ...
}
```

theduke:
Deprecating the macros is something we already talked about.

I'll write more here later, but just for now: https://github.com/graphql-rust/juniper/issues/40 is also relevant for this.

theduke:
Both ResultExt and the jtry macro were removed in `0.9`.

Issue #40, opened 2017-04-09
Complex Errors


theduke:
As I can see, right now, only a simple String error for the `message` field is supported.

I often need more complex errors with additional data, like this for example:

```json
"errors": [
  {
    "message": "Validation failed", 
    "code": "validation_err",
    "data": {
      "username": {"code": "min_length", "message": "must be at least 5 characters long", "meta": {"min": 5, "max": 15}}
    }
  }
]
```

Do you see a way to implement this?

It would require changing the type of `FieldResult`.

As an idea:

```rust
pub enum FieldError {
  Simple(String),
  Custom(String, json_value::Value),
}

type FieldResult<T> = Result<T, FieldError>;
```

The problem is the type of the extra data.
It would require either having FieldError be generic over a type that implements `Serialize` (for serde) or tying it to a value type like [serde_value](https://github.com/arcnmx/serde-value) or `json_value::Value`.

TheServerAsterisk:
Hmm, I think I have an idea on how to implement this if @mhallin doesn't disapprove I'll try to write a PR for it in the next couple of days.

mhallin:
Errors in GraphQL are tricky! I think a clean way of handling failing mutations is one of the less thought-out areas of the whole standard. On the other hand - it might be hard to standardize this on the GraphQL level.

What I've been doing - both in Juniper and when using other systems - is to have the mutation return two fields: a nullable result object and a nullable list of error objects which usually contain a field enum and a message.

This way circumvents the "standard" GraphQL error system for better or worse. An upside is that the errors are actually documented using the same introspective schema like the rest of the system, so you can actually model the error handling pretty rigidly.

That being said, all of these points only cover failing mutations so I'm interested in hearing your proposed solution.

TheServerAsterisk:
I've been doing some research to ensure my initial idea was a good one. However my idea doesn't agree with the current best practice which is further discussed [here](https://github.com/graphql/graphql-js/issues/560). More specifically the best practice appears to be include an error field within your response object. For example, in this [issue](https://github.com/facebook/graphql/issues/117#issuecomment-170180628) here it was mentioned that Facebook has something like: 

```
mutation {
   signUp(email: "hello@atfakeemail.com", password: "thisbetterbeencrypted") {
       # bool
       didSucceed, 
       errors {
         # fields 
      }
   }
}
```
[...]

PR #80, opened 2017-08-08
Field error refactor

mhallin:
An initial attempt at solving #69 and #40. It pretty much matches what I wrote here: https://github.com/graphql-rust/juniper/issues/40#issuecomment-320870842.

To ease construction of custom error objects, I built a small macro to construct `Value` instances from JSON syntax. This might be a bit out of place, in which case it can be removed from the PR. I also don't think it's going to be used for other purposes, but whatever :)

This obviously breaks all users of `.to_field_err` because of the renaming and places where you return `Err("string literal".to_owned())` from a field, but all instances of e.g. `let txn = executor.context().db.transaction()?` should continue to work.

### DIFF
--- a/juniper/src/executor.rs
+++ b/juniper/src/executor.rs
@@ -109,6 +109,21 @@ impl<T: Display> From<T> for FieldError {
 impl FieldError {
     /// Construct a new error with additional data
     ///
+    /// You can use the `graphql_value!` macro to construct an error:
+    ///
+    /// ```rust
+    /// # #[macro_use] extern crate juniper;
+    /// use juniper::FieldError;
+    ///
+    /// # fn sample() {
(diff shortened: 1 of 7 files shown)


rushmorem:
@mhallin Is there any particular reason you are not merging this in?

theduke:
Probably waiting for feedback.

I'll try to carve out some time to go over all MRs tomorrow.

mhallin:
Yeah, I personally don't have a use case for sending extra data with errors so I wanted some input from someone else if this looks reasonable and solves their problem.

theduke:
@mhallin I finally looked over the PR and it looks good to me. Should cover the use case of detailed errors well.

PR #81, opened 2017-08-10
Point to the correct `juniper_rocket` repository

rushmorem:


theduke:
Uh, big oversight.

Thanks

Issue #2, opened 2016-11-16
Interop with async? (Tokio/Futures)


jsgf:
Hi -

Have you put any thought into how this could be used with an async IO model, such as with Tokio? Right now it seems like it's all implicitly synchronous, but perhaps it's just a matter of making the various functions take/return `Future`s?

Thanks

theduke:
The big thing here is that this would also enable an implementation of something like dataloader, which Facebook has deemed the best practice for handling remote data fetching for a GraphQL query.

mhallin:
I've made a few attempts locally at integrating `futures-rs` into Juniper, but there are a lot of stumbling blocks that need to be overcome. There are also some nice properties the current synchronous implementation has that we'll probably lose, particularly around memory usage and heap allocations. This might not be a big issue, but I like the "you don't pay for what you don't use" mentality of Rust and C++ in general.

So, these are some issues that have prevented me from building this. There are probably more :)

* Many AST nodes, particularly identifiers, use references to the original query string to avoid copying. Rust statically guarantees that no AST nodes will outlive the query string, and it's easy to reason about when `execute` just parses and executes the query synchronously.
* `GraphQLType::resolve*` would need to return a `Box<Future>`, which means that _all_ fields will cause extra heap allocations, even if it's just `field like_count() -> i32 { self.like_count }`. Now that I think about it, maybe you could change `FieldResult` to something like `enum { Immediate(Value), Err(String), Deferred(Box<Future<Item=Value, Error=String>>) }`.
* The [core execution logic](https://github.com/mhallin/juniper/blob/master/src/types/base.rs#L291) would need to be transformed into something that joins multiple futures together and emits a result. No fundamental problem here, it was just a very difficult programming task :)
* The `futures-rs` crate changing around a bit while working on this. This shouldn't be the case anymore, but I would be weary releasing Juniper 1.0 while depending on a pre-1.0 release of `futures-rs`.

I've been dabbling with `futures-rs` and Tokio in another project that I'm working on, and the general feeling I get is that's it's pretty unergonomic to use. I've stumbled upon cases where I couldn't break out a big `and_then` callback to a separate function because of various reasons: either the type was "untypeable" - i.e. containing closures, `BoxFuture` requires `Send` despite the name, `Box<Future>` did not work because of lifetime issues, and even when switching to nightly and returning `impl Future` there have been cases where returning lifetimes have been an issue.

[...]

PR #83, opened 2017-08-13
Expose current executor's location

rushmorem:
`Executor` has a `push_error` method which is very useful for reporting
back errors. However, this method needs to know the location of the
error (`SourcePosition`). This commit adds a `location` method to the
executor.

### DIFF
--- a/juniper/src/executor.rs
+++ b/juniper/src/executor.rs
@@ -288,8 +288,7 @@ impl<'a, CtxT> Executor<'a, CtxT> {
         match self.resolve(info, value) {
             Ok(v) => v,
             Err(e) => {
-                let position = self.field_path.location().clone();
-                self.push_error(e, position);
+                self.push_error(e);
                 Value::null()
             }
         }
@@ -355,8 +354,19 @@ impl<'a, CtxT> Executor<'a, CtxT> {
         self.fragments.get(name).map(|f| *f)
     }
 
-    /// Add an error to the execution engine
-    pub fn push_error(&self, error: FieldError, location: SourcePosition) {
+    /// The current location of the executor
+    pub fn location(&self) -> &SourcePosition {
+        self.field_path.location()
+    }
+
+    /// Add an error to the execution engine at the current executor location
+    pub fn push_error(&self, error: FieldError) {
+        let location = self.location().clone();
+        self.push_error_at(error, location);
+    }
+
+    /// Add an error to the execution engine at a specific location
+    pub fn push_error_at(&self, error: FieldError, location: SourcePosition) {
         let mut path = Vec::new();
         self.field_path.construct_path(&mut path);
 
--- a/juniper/src/types/base.rs
+++ b/juniper/src/types/base.rs
@@ -390,7 +390,7 @@ fn resolve_selection_set_into<T, CtxT>(
                 match field_result {
                     Ok(v) => merge_key_into(result, response_name, v),
                     Err(e) => {
-                        sub_exec.push_error(e, start_pos.clone());
+                        sub_exec.push_error_at(e, start_pos.clone());
                         result.insert((*response_name).to_owned(), Value::null());
                     }
                 }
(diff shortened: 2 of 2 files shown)


mhallin:
This looks good, but I would like to know your use case for needing to call `push_error` manually. Is returning errors from fields not enough, or is it related to #84?

rushmorem:
It's not related to https://github.com/graphql-rust/juniper/issues/84. It's loosely related to https://github.com/graphql-rust/juniper/issues/40, which https://github.com/graphql-rust/juniper/pull/80 seeks to resolve. To me, these are all related in that they are trying to improve the ergonomics when processing multiple inputs.

For my specific use case, I'm building a domain registration API. A user might want to check availability of multiple domain names at the same time. For example:-

```graphql
{
  domains {
    check(domains: ["example.com", "example.local"]) {
      domain
      available
    }
  }
}
```

Like in the example above, some of the domains submitted may not be valid (`example.local` in this case). I want my servers to be fault tolerant. That is, instead of simply returning an error message saying `example.local` is not a valid domain name, I want it to return the correct result for `example.com` and an error for `example.local` like so:-

```json
{
  "data": {
    "domains": {
      "check": [
        {
          "available": false,
          "domain": "example.com"
        }
      ]
    }
  },
  "errors": [
    {
      "message": "example.local is not a valid domain name",
      "locations": [
        {
          "line": 3,
          "column": 5
        }
      ],
      "path": [
        "domains",
        "check"
      ]
    }
  ]
}
```

I couldn't figure of out any other way of achieving this besides using `push_error`.

srijs:
Wouldn't it be more idiomatic if you used the built-in mechanisms that GraphQL provides to do batch queries (perhaps combined with fragments to keep it DRY)?

```
fragment checkField on Check {
  domain,
  available
}

query {
  domains: {
    check(domain: 'example.com') {
      ...checkFields
    }
    check(domain: 'example.local') {
      ...checkFields
    }
  }
}
```

This would get you a `FieldResult` per domain check that you can expose the error through.

rushmorem:
TLDR: Using a `List` as input for a use case like this one is more user friendly and more efficient.

---

I may be wrong about this but IMHO batching like that is ideal for queries that actually do different things instead of doing the same thing on different inputs.

There are a number of drawbacks for batching a query like this versus simply accepting a `List` like I'm doing:-

It complicates client queries unnecessarily. Compare this:-

```graphql
{
  domains {
    check(domains: ["example.com", "example.local"]) {
      domain
      available
    }
  }
}
```

to this:-

```graphql
fragment checkField on Check {
  domain,
  available
}

query {
  domains {
    alias1: check(domain: "example.com") {
      ...checkFields
    }
    alias2: check(domain: "example.local") {
      ...checkFields
    }
  }
}
```

* The user will have to write code for generating unique aliases for each domain.
* The user will also have to handle those aliases in the returned response.
* The first query is easier to understand at a glance.

Hence, I think batching in this sense is overkill for such a simple use case.

There are also practical reasons why I wouldn't use it for a use case like this:-

* In the first query, since I'm simply accepting a `Vec<String>`, I can easily strip duplicates from the query before processing. This is impossible using the second approach. I will have to trust the user to send unique domains, otherwise I will waste system resources processing duplicate domains for the same request.
* Because I'm using the first approach, I process the domains concurrently. I wouldn't be able to do this otherwise. This wouldn't be much of a big deal if `juniper` [processed fields asynchronously](https://github.com/graphql-rust/juniper/issues/2).

mhallin:
I see, this seems reasonable :)

May I suggest a small API change though: since you probably always will write `executor.push_error(..., executor.location())`, we could change `push_error` to always use the current location, and then have another method like `push_error_at_location` that takes an explicit location?

rushmorem:
@mhallin That's a good idea. Let me do that.

rushmorem:
@mhallin Instead of `push_error_at_location` how about just `push_error_at` much like how `std` library methods (`split_at` et al) are named?

rushmorem:
@mhallin Done. I used `push_error_at`. Let me know if you still prefer `push_error_at_location` and want me to change to that.

mhallin:
This looks great, thanks a lot!

Issue #82, opened 2017-08-13
Preserve the order of requested fields?


rushmorem:
According to the [section 7.1](http://facebook.github.io/graphql/#sec-Serialization-Format):

> Serialization formats which can represent an ordered map should preserve the order of requested fields as defined by CollectFields() in the Execution section. Serialization formats which can only represent unordered maps should retain this order grammatically (such as JSON).
>
> Producing a response where fields are represented in the same order in which they appear in the request improves human readability during debugging and enables more efficient parsing of responses if the order of properties can be anticipated.

and [sub-section 7.1.1](http://facebook.github.io/graphql/#sec-JSON-Serialization):

> For example, if the query was { name, age }, a GraphQL server responding in JSON should respond with  { "name": "Mark", "age": 30 } and should not respond with { "age": 30, "name": "Mark" }.

We may be able to achieve this by using [OrderMap](https://crates.io/crates/ordermap) instead of `HashMap`. Has anyone tried this yet?

PR #85, opened 2017-08-24
Preserve the order of requested fields

LegNeato:
Fixes https://github.com/graphql-rust/juniper/issues/82

Note: I am new to Rust, please let me know if there is anything I can do better.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,16 @@ The repository was restructured to a multi crate workspace to enable several new
 
 ### New features
 
-* New juniper_codegen crate which provides custom derives: 
+* New juniper_codegen crate which provides custom derives:
   * `#[derive(GraphQLInputObject)]`
   * `#[derive(GraphQLEnum)]`
   * `#[derive(GraphQLObject)]`
+
+## Breaking changes
+
+* To better comply with the specification, order of requested fields is
+  now preserved.
+  ([#82](https://github.com/graphql-rust/juniper/issues/82)
 
 ## [0.8.1] – 2017-06-15
 
--- a/juniper/Cargo.toml
+++ b/juniper/Cargo.toml
@@ -25,6 +25,7 @@ expose-test-schema = []
 default = ["uuid"]
 
 [dependencies]
+ordermap = { version = "^0.2.11", features = ["serde-1"] }
 serde = { version = "^1.0.8" }
 serde_derive = {version="^1.0.8" }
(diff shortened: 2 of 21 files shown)


theduke:
Since the spec recommends it, I'm definitely in favor.

There are two open questions for me:

* what is the performance impact of linked-hash-map vs Hashmap?

* I'm not so comfortable with bringing in https://github.com/contain-rs/linked-hash-map as a dependency
  The README states:
  WARNING: THIS PROJECT IS IN MAINTENANCE MODE, DUE TO INSUFFICIENT MAINTAINER RESOURCES

LegNeato:
Re: perf, I couldn't get `cargo bench` to work locally...any pointers would be appreciated.

Re: `linked-hash-map`, I was hesitant as well.  https://github.com/bluss/ordermap looks better maintained but doesn't yet have serde support (https://github.com/bluss/ordermap/pull/28) which `juniper` needs. Also, `linked-hash-map` maintainers said they will merge changes to keep it compatible and building...which I think is all we really need for juniper in the future. I intend to keep an eye on `ordermap` and switch if/when it has serde support.

rushmorem:
Can't we switch to OrderMap and use [remote derive](https://serde.rs/remote-derive.html) for now?

LegNeato:
I just tried and it doesn't look like you can use remote derive for this, as just using

```rust
#[derive(Serialize, Deserialize)]
#[serde(remote = "OrderMap")]
```
Fails as the default derive isn't sufficent and 

```rust
#[serde(remote = "OrderMap")]
```
fails without `derive()`.

I'm a Rust n00b, so if there is a way to do this that I am missing please let me know!

LegNeato:
Also, I am not sure but https://github.com/bluss/ordermap/issues/19 might cause some issues here too. I'd prefer to merge this as-is (assuming perf is ok), as I don't think switching in the future will be that big of a pain for users (but I could be wrong!).

LegNeato:
I got them to do a release, so latest changes use `ordermap`.

bluss:
Nice to see that you get some utility out of ordermap. As a point of order, you'll should probably add Ordermap's 1.0-ing on your own 1.0 road map, since it's a “public dependency” when used like this.

OrderMap itself doesn't have an 1.0 road map, for the record, but the future is open ended.

mhallin:
Thanks for this PR! Bringing us closer to spec compliance :)

PR #91, opened 2017-09-10
url feature: Implement url::Url as a scalar type

LegNeato:
Essentially the same as https://github.com/graphql-rust/juniper/pull/89.


mhallin:
Very nice, thank you.

Issue #87, opened 2017-08-28
Implement GraphQLType for chrono::DateTime


theduke:
Since the chrono crate is so popular, it's a pain not to have it implement GraphQLType.

We should optionally implement it behind a feature flag, representing DateTime with a string representation.

Mange:
Might I suggest a float representation (UNIX timestamp with optonally fractional second) instead? It's smaller on the wire and trivial to "parse" (as there is no parsing).

If you are worried about different float representations or limits to 64-bit numbers, the number could be stringified first, letting clients pick a precision. Still easier to parse and validate.

LegNeato:
I noticed you did this for `uuid` as well...is this process scalable? Should there perhaps be a different subcrate with these integrations or instead encourage others to publish crates with GraphQL types for their specific dep rather than `juniper` maintaining them?

theduke:
It is not scalable, but I think we should do it for the most important / most widely used types.

The problem is how trait implementation orphan rules work in Rust.

You can only implement a trait for a type in a crate that defines the trait or defines the type.

So the definitions have to live in the main Juniper crate, which defines the GraphQLType trait.

The only other option is to use wrapper types and implement the trait on those wrappers, which is awkward.

theduke:
@Mange good point, and I personally also always prefer to use floats.

If we implement this, floats should probably be the default.

BUT, the current convention in JS land I seem to see is to mostly use iso8601 strings.

Is there a need for both to be available? (as in both floats and string representations)

LegNeato:
@theduke ah, totally forgot about the trait stuff! Makes sense to me.

passcod:
Chrono's DateTime includes a timezone, so representing it as a number loses information.

Mange:
@passcod that settles it for me.

LegNeato:
@theduke I hope you don't mind that I put up a PR for this. Feel free to discard it if you were thinking of something different.

PR #92, opened 2017-09-26
chrono feature: Implement GraphQLType for chrono::NaiveDate and chrono::DateTime

LegNeato:
Fixes https://github.com/graphql-rust/juniper/issues/87

### DIFF
--- a/juniper/Cargo.toml
+++ b/juniper/Cargo.toml
@@ -22,9 +22,10 @@ path = "benches/bench.rs"
 [features]
 nightly = []
 expose-test-schema = []
-default = ["url", "uuid"]
+default = ["chrono", "url", "uuid"]
 
 [dependencies]
+chrono = { version = "^0.4.0", optional = true }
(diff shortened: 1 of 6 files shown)


passcod:
I'd just like to note that JSON numbers (i.e. IEEE doubles) are not precise enough for nanosecond datetimes expressed as timestamps. They're precise enough for up to microsecond resolution.

theduke:
@passcod I'm aware of the tradeoff, do you personally see that as a problem though?

I'd reckon microsecond resolution would be sufficient for most use cases, which this intends to cover. Most times used in a context were GraphQL is used tend to be only second resolution anyway in my experience.

passcod:
Ah, no, I mean it should be noted in the documentation, so that _users_ are aware of the tradeoff.

LegNeato:
Added the additional types and serialization + docs.

mhallin:
This looks very nice, thank you!

PR #93, opened 2017-10-02
perf(schema): use FnvHashMap for type and directive names

srijs:
Changes `SchemaType` to use a `HashMap` backed by the [https://github.com/servo/rust-fnv](Fowler–Noll–Vo hash function). This hash function performs much better for smaller inputs (think short strings like most GraphQL type names) than the default hash function, SipHash. The downside is that it does not have the same cryptographic properties, which open it up to DoS vulnerabilities when the input can be manipulated from the outside.

In this change set, I have chosen to apply FNV only to the type and directive names tracked by the `SchemaType`, for two reasons:

1. Look-ups in `SchemaType::type_by_name` and co. have been showing up as the most frequent/expensive use of hashing for my workload.
2. I believe it is reasonable to assume that a malicious party would not have control over the schema type, therefore could not leverage FNV for a DoS attack (contrary to e.g. using it in the validation logic).

Here you can see how `SchemaType::type_by_name`, `::concrete_type_by_name` as well as both `::make_type` and `ValidatorContext::with_pushed_type` (these use `type_by_name` internally) show up in my profile:

schematype-hash-perf

In my (somewhat limited) benchmark, the change appears to lead to a worthwhile performance improvement.

Pre:

```
Running 30s test @ http://localhost:3000/graphql
  2 threads and 4 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   572.58us  336.26us  14.93ms   97.35%
    Req/Sec     3.54k   284.33     3.93k    77.08%
  Latency Distribution
     50%  526.00us
     75%  573.00us
     90%  663.00us
     99%    1.15ms
  211827 requests in 30.10s, 26.46MB read
Requests/sec:   7037.39
Transfer/sec:      0.88MB
```

Post:

```
Running 30s test @ http://localhost:3000/graphql
  2 threads and 4 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   550.45us  213.25us   8.73ms   97.53%
    Req/Sec     3.64k   169.32     3.93k    79.87%
  Latency Distribution
     50%  520.00us
     75%  554.00us
     90%  616.00us
     99%    0.96ms
  217928 requests in 30.10s, 27.23MB read
Requests/sec:   7240.40
Transfer/sec:      0.90MB
```

### DIFF
--- a/juniper/Cargo.toml
+++ b/juniper/Cargo.toml
@@ -25,6 +25,7 @@ expose-test-schema = []
 default = ["url", "uuid"]
 
 [dependencies]
+fnv = "1.0.3"
 ordermap = { version = "^0.2.11", features = ["serde-1"] }
 serde = { version = "^1.0.8" }
 serde_derive = {version="^1.0.8" }
--- a/juniper/src/executor.rs
+++ b/juniper/src/executor.rs
@@ -4,6 +4,8 @@ use std::borrow::Cow;
 use std::collections::HashMap;
 use std::sync::RwLock;
 
+use fnv::FnvHashMap;
+
 use GraphQLError;
 use ast::{Definition, Document, Fragment, FromInputValue, InputValue, OperationType, Selection,
           ToInputValue, Type};
@@ -24,7 +26,7 @@ use types::name::Name;
 /// into `Type` instances and automatically registers them.
 pub struct Registry<'r> {
     /// Currently registered types
-    pub types: HashMap<Name, MetaType<'r>>,
+    pub types: FnvHashMap<Name, MetaType<'r>>,
 }
 
 #[derive(Clone)]
@@ -519,7 +521,7 @@ where
 
 impl<'r> Registry<'r> {
     /// Construct a new registry
-    pub fn new(types: HashMap<Name, MetaType<'r>>) -> Registry<'r> {
+    pub fn new(types: FnvHashMap<Name, MetaType<'r>>) -> Registry<'r> {
         Registry { types: types }
     }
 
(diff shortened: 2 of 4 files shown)


Techcable:
Why not seahash? Isn't that [signifigantly faster](https://docs.rs/seahash/3.0.5/seahash/#benchmark)?

srijs:
# Juniper Benchmarks

## FnvHasher

```
running 2 tests
test introspection_query ... bench:   1,744,556 ns/iter (+/- 422,545)
test query_type_name     ... bench:      14,961 ns/iter (+/- 3,658)

test result: ok. 0 passed; 0 failed; 0 ignored; 2 measured
```

## SeaHasher

```
running 2 tests
test introspection_query ... bench:   1,794,380 ns/iter (+/- 554,186)
test query_type_name     ... bench:      15,851 ns/iter (+/- 3,547)

test result: ok. 0 passed; 0 failed; 0 ignored; 2 measured
```

## DefaultHasher

```
running 2 tests
test introspection_query ... bench:   2,058,095 ns/iter (+/- 358,108)
test query_type_name     ... bench:      18,215 ns/iter (+/- 6,773)

test result: ok. 0 passed; 0 failed; 0 ignored; 2 measured
```

srijs:
@theduke @mhallin Friendly ping on this one, would appreciate a review!

mhallin:
I think this looks good, thanks for the improvement! I think your assumptions about hash-collisions attacks are reasonable.

Without having done any deeper performance analysis, I'd think that Juniper's biggest overhead on larger queries is running the validation passes. Having a MFU/MRU cache of parsed and validated ASTs could probably help there, or implementing [Apollo's persisted queries](https://dev-blog.apollodata.com/persisted-graphql-queries-with-apollo-client-119fd7e6bba5).

srijs:
@mhallin agreed. Coincidentally, my work on removing the lifetime parameter from the AST for purposes of introducing async could prove useful for introducing a query cache!

PR #97, opened 2017-10-16
Add support for Z time zone and fractional seconds in date formats

OinkIguana:
I was having trouble getting my dates to parse, and realized it was because Z time zone was not being recognized. I changed it to use the built in "%+" format for parsing RFC3339 dates and added tests to check for this case.


theduke:
@OinkIguana cheers!

PR #98, opened 2017-10-18
Fixed case where .unwrap was being called instead of the Option just being propagated up the call chain

joshmarlow:
I ran into a case where Juniper was unwrapping a value and then placing it back into an Option.  It seems reasonable to just pass the Option up the call stack.

I've been playing with Juniper the last few weeks and it's really cool!  Thanks for building, maintaining and extending it!

### DIFF
--- a/juniper/src/types/base.rs
+++ b/juniper/src/types/base.rs
@@ -111,7 +111,7 @@ impl<'a> Arguments<'a> {
     {
         match self.args {
             Some(ref args) => match args.get(key) {
-                Some(v) => Some(v.convert().unwrap()),
+                Some(v) => v.convert(),
                 None => None,
             },
             None => None,


joshmarlow:
As near as I can tell, my change shouldn't have caused these failures but I'm happy to put in some effort to make it pass with some guidance :)

theduke:
@joshmarlow thanks!

Could you maybe add a test that validates the expected behaviour?

joshmarlow:
@theduke totally!  I'll try to do that this weekend.

theduke:
@joshmarlow are you still interested here?

If not, no problem, just say so.

Then I'll merge this and add some tests myself.

joshmarlow:
@theduke I am but the holidays through me off - I should be able to Sunday but I'll ping you if I can't find thetime.

joshmarlow:
@theduke I don't think I'll be able to get to a test here anytime soon, but there should no longer be any merge conflicts if you want to merge this.  Thanks!

PR #100, opened 2017-10-19
Switched to FnvHashMap and removed nightly only features to fix failing tests

matt-keibler:
* Tests needed to be updated from `std::collections::HashMap` to `fnv::FnvHashMap` to match the `Registry` constructor
* Nightly only features needed to be removed for stable branches to pass in CI


theduke:
@matt-keibler awesome, thanks.

I wanted to do the same thing this weekend, so the effort is much appreciated!

PR #102, opened 2017-10-28
Derive Clone for GraphQLRequest

thomas-jeepe:
This is just a useful thing for me because I need to clone the http request.

This shouldn't really influence anything else, every field already implements `Clone`


theduke:
Sure thing.

Cheers.

PR #110, opened 2017-11-18
Generate coverage report in travis build and upload to codecov

sagiegurari:
I think this would really help in both providing you and the community feedback of the code quality of juniper and help you see where the library is missing tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: rust
 cache: cargo
+sudo: true
 
 rust:
   - stable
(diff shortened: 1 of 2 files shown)


sagiegurari:
You can see your coverage now at: https://codecov.io/gh/graphql-rust/juniper
You are at 87.61% which is great

theduke:
@sagiegurari wow awesome, thanks.

We wanted to do this already but didn't get around to it.

PR #114, opened 2017-12-02
WIP: 0.9 release

theduke:
Work in progress refactorings towards a 0.9 release.


sagiegurari:
bitflags stopped support of older rust versions so cargo make minimal supported version is now 20

theduke:
@sagiegurari thanks for the heads up!

theduke:
@sagiegurari I'm trying to disable building the `juniper_rocket` crate on non-nightly.

I've added a task override here: https://github.com/graphql-rust/juniper/pull/114/commits/4cebcee8e6846d1085b97ca603e9a802ff24731a

This should work in theory, right? It seems like the condition is not properly checked for workspace members though.

sagiegurari:
I'm adding support now for this specific feature (skipping specific member based on condition). Hopefully should be available soon.

sagiegurari:
published, you can now remove the juniper_rocket/Makefile.toml and just add a Makefile.toml in your root workspace with the following and it should work:

````toml
[tasks.init]
condition = { channels = ["beta", "stable"] }
env = { "CARGO_MAKE_WORKSPACE_SKIP_MEMBERS" = "juniper_rocket" }
````

[skip member docs](https://github.com/sagiegurari/cargo-make#usage-workspace-support-skip-members)

For appveyor, you should also run a composite flow in test script to support this so instead of running

````sh
cargo make ci-flow
````

Use like in .travis.yml

````sh
cargo make workspace-ci-flow --no-workspace
`````

sagiegurari:
Added juniper_tests to be skipped in non nightly. seems it is failing on windows on features issue.
maybe you can dig in more to see what exactly the issue is there.

theduke:
@sagiegurari hmm the issue seems to be the `cargo build --all-features`, because the `yansi` crate has a `nightly` feature which activates a feature.

No idea why this works on linux and not on windows though.

sagiegurari:
that what puzzled me as well unless they have some sort of code per platform using cfg there.
anyway so for now, in my code i'm skipping this module for non nightly but can be configured differently if you wish like running build without all features only for that specific module.

PR #117, opened 2017-12-05
windows build fix

sagiegurari:
**please do not merge until both linux and windows builds finish successfully.**

The changes include

* remove the --all-features from the juniper tests
* skip juniper rocket for beta/stable rust channel.
* both appveyor and travis will invoke the exact same flow.

PR #120, opened 2017-12-06
windows build

sagiegurari:
see #117 

The changes include

* remove the --all-features from the juniper tests
* skip juniper rocket for beta/stable rust channel and for windows also for gnu
* both appveyor and travis will invoke the exact same flow.

Fix was done in cargo-make to resolve the issue of failing member builds not failing the workspace level build.

### DIFF
--- a/juniper_rocket/Cargo.toml
+++ b/juniper_rocket/Cargo.toml
@@ -16,8 +16,8 @@ serde_derive = {version="1.0.2" }
 serde_json = { version = "1.0.2" }
 juniper = { version = "0.9.1" , path = "../juniper"}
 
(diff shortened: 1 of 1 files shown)


sagiegurari:
ring failed to compile so a specific member build failed which cause the workspace build to fail.
So its actually a good use case to see that the workspace level build works good now for both linux and windows.

theduke:
haha I pushed to your branch and Github decided just to close it, and now I can't re-open the mr.

theduke:
Re-opened with your commit cherry-picked over at https://github.com/graphql-rust/juniper/pull/129

PR #121, opened 2017-12-17
Update Rocket to latest version

LegNeato:
Rocket requires nightly, but on nightly 2017-12-13 and after
the version juniper's integration was using wouldn't
build. This is solved by updating Rocket to `0.3.4` which
according to the [changelog](https://github.com/SergioBenitez/Rocket/blob/v0.3.4/CHANGELOG.md#version-034-dec-14-2017):

"Codegen was updated for 2017-12-13 nightly."

### DIFF
--- a/juniper_rocket/Cargo.toml
+++ b/juniper_rocket/Cargo.toml
@@ -16,8 +16,8 @@ serde_derive = {version="1.0.2" }
 serde_json = { version = "1.0.2" }
 juniper = { version = "0.9.1" , path = "../juniper"}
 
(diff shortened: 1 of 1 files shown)


theduke:
@LegNeato thanks yet again!

LegNeato:
Ah, good point...I'm still getting the hang of cargo 😄 . I don't think the nightly change is backwards compatible, as Rocket's changelog also says:

`Minimum required rustc is 1.24.0-nightly 2017-12-13`

Issue #116, opened 2017-12-03
FromForm trait impl for GraphQLRequest should url_decode parameters


theduke:
When parameters are submitted via `GET` request, they are URL escaped. The current implementation of FromForm for GraphQLRequest does not decode the parameters, which results in the error like:

```
{
  "errors": [
    {
      "message": "Unknown character \"%\"",
      "locations": [
        {
          "line": 1,
          "column": 1
        }
      ]
    }
  ]
}
```

reported by @idubrov here: https://github.com/graphql-rust/juniper_rocket/issues/2

PR #122, opened 2017-12-17
url_decode parameters in Rocket integration

LegNeato:
Also as part of this change I fixed and enabled
the http integration tests for Rocket.

Fixes https://github.com/graphql-rust/juniper/issues/116.

### DIFF
--- a/juniper_rocket/src/lib.rs
+++ b/juniper_rocket/src/lib.rs
@@ -65,6 +65,7 @@ use juniper::RootNode;
 /// See the `http` module for more information. This type can be constructed
 /// automatically from both GET and POST routes by implementing the `FromForm`
 /// and `FromData` traits.
+#[derive(Debug, PartialEq)]
 pub struct GraphQLRequest(http::GraphQLRequest);
 
 /// Simple wrapper around the result of executing a GraphQL query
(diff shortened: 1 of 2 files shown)


juniper_rocket/src/lib.rs
```diff
@@ -107,30 +107,46 @@ impl<'f> FromForm<'f> for GraphQLRequest {
...
-                        .map_err(|err| err.description().to_owned())?);
-                },
-                _ => if strict {
-                    return Err(format!("Prohibited extra field '{}'", key).to_owned());
-                },
+                "query" => {
+                    if query.is_some() {
+                        return Err("Query parameter must not occur more than once".to_owned());
+                    } else {
+                        let decoded = value.url_decode().map_err(|_| ()).unwrap();
```
theduke: We need to properly return an error here instead of unwrapping. 
I also don't get the map_err?

juniper_rocket/src/lib.rs
```diff
@@ -107,30 +107,46 @@ impl<'f> FromForm<'f> for GraphQLRequest {
...
+                        query = Some(decoded);
+                    }
+                }
+                "operation_name" => {
+                    if operation_name.is_some() {
+                        return Err(
+                            "Operation name parameter must not occur more than once".to_owned(),
+                        );
+                    } else {
+                        let decoded = value.url_decode().map_err(|_| ()).unwrap();
```
theduke: Same as above.

juniper_rocket/src/lib.rs
```diff
@@ -148,7 +164,13 @@ impl FromData for GraphQLRequest {
     type Error = String;
 
     fn from_data(request: &Request, data: Data) -> FromDataOutcome<Self, Self::Error> {
-        if !request.content_type().map_or(false, |ct| ct.is_json()) {
+        // We cannot use `request.content_type()` here.
+        // https://github.com/SergioBenitez/Rocket/issues/505
```
theduke: I'd like to follow up on https://github.com/SergioBenitez/Rocket/issues/505 to trace the root cause here.

LegNeato:
Thanks for the review, I'm planning on getting to this in the coming week!

LegNeato:
Ok, added unit tests and now handle the decode errors. I'm not sure how I can feed non-UTF8 data in to trigger a decode error in tests...feel free to give a rust n00b pointers!

LegNeato:
Build failure is https://github.com/briansmith/ring/issues/598

LegNeato:
Friendly poke 👉 @theduke

juniper_rocket/src/lib.rs
LegNeato: Turns out this was a bug in Rocket. The fix is out in rocket v3.6. Let's land this as-is and after https://github.com/graphql-rust/juniper/pull/127 lands I'll post follow-up  to revert to `request.content_type()`

LegNeato: https://github.com/graphql-rust/juniper/pull/127 landed so reverting this.

theduke:
@LegNeato sorry for the big delay.

Thanks for the tests!
I'll merge once the pipeline succeeds.

PR #126, opened 2018-01-13
Add a master changelog

LegNeato:
This makes it so people using git dependencies know what has changed. It also
gives a spot to make a running changelog so when we do a release we can just
copy and paste.

### DIFF
--- a/changelog/master.md
+++ b/changelog/master.md
@@ -1,0 +1,9 @@
+## [master]
+
+## Changes
+
+### `__typename` for unions
+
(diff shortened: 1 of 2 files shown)


theduke:
Thanks again.

I'd like to keep a separate changelog for juniper_rocket though, could you put the rocket info in `juniper_rocket/changelog/master.md`?

Also, you wrote 0.3.4 instead of 0.3.6 in the changelog. ;)

LegNeato:
@theduke thanks for the speedy review! Latest commit has comments addressed.

theduke:
@LegNeato cheers. 👍

Issue #125, opened 2018-01-08
Build failure in `juniper_rocket` with recent nightlies


LegNeato:
Latest nightlies broke `ring`, which broke `rocket`, which breaks `juniper_rocket`.

We need to update `rocket` when`ring` is updated. See https://github.com/briansmith/ring/issues/598 and https://github.com/SergioBenitez/Rocket/issues/513.

PR #127, opened 2018-01-13
Update rocket in juniper_rocket to work with latest nightlies

LegNeato:
`juniper_rocket` now requires nightly >= 2018-01-12. See
https://github.com/SergioBenitez/Rocket/issues/513#issuecomment-357407524.

Fixes https://github.com/graphql-rust/juniper/issues/125.


LegNeato:
Note this isn't strictly required as the latest nightly makes the change backwards compatible, which should unbreak nightly for us. That being said, rocket bumped their nightly dependency so we might as well track them as the whole point of the crate is to interface with rocket.

PR #129, opened 2018-01-13
windows build

theduke:


theduke:
I'll merge once the pipeline succeeds.

sagiegurari:
Looks like codecov returned some error while downloading their bash script.
probably were momentarily down or something

++curl -s https://codecov.io/bash
/dev/fd/63: line 1: syntax error near unexpected token `newline'
/dev/fd/63: line 1: `<!DOCTYPE html>'

sagiegurari:
In other builds that 'succeeded' I also see codecov issues:
 X> Failed to upload

So looks like the codecov website is having temporary issues.

theduke:
@sagiegurari yeah I saw that. 

I think codecov should never fail the build.
Could you update the cargo-make definitions to ignore failures of the codecov task?

sagiegurari:
many go with that approach, but I personally don't like it because

* you will never know that coverage stopped working. manual monitoring never works
* you want to know if a PR of even your code change has whole blocks of code that are not tested.

If you really want, you can add force=false to the coverage task:
https://github.com/sagiegurari/cargo-make#ignoring-errors

something like in your main makefile:

````toml
[tasks.workspace-coverage-pack]
force = true
````

sagiegurari:
@theduke looks like the codecov website issue was resolved and builds are passing.
if you do want coverage to be optional, I can do a PR next week for that, but personally I prefer builds to fail if anything goes wrong.

theduke:
I agree that it's nice to see when it fails (even though you would notice anyway because the bot comment with the coverage change does not appear).

But let's leave it like this for now and reconsider if a similar problem comes up again.

PR #134, opened 2018-01-19
Provide type info to the concrete_type_name

idubrov:


LegNeato:
@idubrov Looks like you missed changing `src/macros/union.rs` and tests are failing.

idubrov:
Fixed. Sorry about that.

Issue #124, opened 2018-01-03
Query introspection in resolvers


weiznich:
## Short version:
I've tried to write a small wrapper layer for a database using diesel. A straight forward implementations leads to N+1 query problems.

## Long version:

Given the following database schema:
```rust
table! {
    heros {
        id -> Integer,
        name -> Text,
        species -> Integer,
    }
}

table! {
    species {
        id -> Integer,
        name -> Text,
    }
}

joinable!(hero -> species(species));
allow_tables_to_appear_in_same_query!(hero, species);
```
It should be possible to map this directly to the following graphql schema.
```graphql
type Character {
  id: Int!
  name: String!
  species: Species!
}

type Species {
  id: Int!
  name: String!
  heros: [Character!]!
}

type Query {
  heros(id: Int, name: String): Character
  species(id: Int, name: String): Species
}
```
Implementing this graphql api in a straight forward way using juniper leads to the following code
```rust
struct DbConnection(Mutex<SqliteConnection>);

#[derive(Queryable, Debug, GraphQLObject)]
struct Character {
    id: i32,
    name: String,
    species: Species,
}

#[derive(Queryable, Debug)]
struct Species {
    id: i32,
    name: String,
}


graphql_object!(Species: DbConnection |&self| {
    field id() -> i32 {
        self.id
    }
    field name() -> &String {
        &self.name
    }
    field heros(&executor) -> FieldResult<Vec<Character>> {
        let conn = &*executor.context().0.lock()?;
        // This results in N+1 queries
        heros::table.filter(heros::species.eq(self.id)).load(conn).map_err(Into::into)
    }
});

struct Query;

graphql_object!(Species: DbConnection |&self| {
    field heros(&executor, id: Option<i32>, name: Option<String>) -> FieldResult<Vec<Character>> {
        let conn = &*executor.context().0.lock()?;

       let q = heros::table.inner_join(species::table)
           .select((heros::id, heros::name, (species::id, species::name)));

       match (id, name) {
            (Some(id), Some(name)) => q.filter(heros::id.eq(id).and(heros::name.eq(name)).load(conn).map_err(Into::into),
            (Some(id), None) => q.filter(heros::id.eq(id)).load(conn).map_err(Into::into),
            (None, Some(name)) => q.filter(heros::name.eq(name)).load(conn).map_err(Into::into),
            (None, None) => q.load(conn).map_err(Into::into),
       }
    }

[...]

PR #136, opened 2018-01-26
Add a basic lookahead functionality

weiznich:
I've tried a bit around how to implement a look ahead feature as described in #124.
This should be seen as WIP. I've only opened this PR to start some discussion about this design.


fungl164:
Hi,

Was looking at your implementation and had a few questions:
1) Would it make more sense to define ```LookAheadSelection.arguments``` and ```LookAheadSelection.childs``` as ```Option<Vec<?>>``` in order to avoid unnecessary creation of objects and arrays?
2) While on the topic, would it make more sense to change ```childs``` to ```children``` as the more appropriate term?
3) Could you include a simple example on how you would use this along with diesel as described on #124?
 
Many thanks for the contribution!

weiznich:
> Would it make more sense to define LookAheadSelection.arguments and LookAheadSelection.childs as Option<Vec<?>> in order to avoid unnecessary creation of objects and arrays?

As far as I understand the [documentation of `Vec::new`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.new) this is not necessary because the constructor does not allocate anything.  So I don't see any reason to use `Option<Vec<T>>` in place of `Vec<T>`

> While on the topic, would it make more sense to change childs to children as the more appropriate term?

I'm really bad in naming things, so just suggest a better name and I will change it

> Could you include a simple example on how you would use this along with diesel as described on #124?

I have same code using this api, that provides a somehow working solution for a specific database. It is  currently quite hacky, so I need to do some cleanup before showing it somewhere.
Also before merging this I need to write some documentation. (I will start on this after we have figured out a api for this feature.)

fungl164:
@weiznich all great points. thnxs! Looking forward to your updates... : )

weiznich:
> I have same code using this api, that provides a somehow working solution for a specific database. It is currently quite hacky, so I need to do some cleanup before showing it somewhere.

Just a little update about that: I've not forgotten about this PR, but I found that writing diesel code that interacts with the lookahead functionality involves quite a lot of boilerplate code. So I've decided to try to write some generic code that is able to be used in place of the boilerplate code, but this needs some time. 
Hopefully I will have something to show in the next few weeks.

fungl164:
This would be great! Thnxs!!

weiznich:
So finally I got something to [show](https://github.com/weiznich/wundergraph). Please note that the look ahead functionality is not visible in the example code, because it is only used inside the library.

weiznich:
@fungl164 @theduke 
Are there any comments on the design here? Anything improvements that needs to be done?
Otherwise I will try to add some documentation to this so this could be merged in near future?

fungl164:
@weiznich this is great! just downloaded your code and tried it.  Seems to be working right out-of-the-box.  Have not had much time to look into the details; however, I did notice the SQL queries pull all the fields on the first select, but not on the subsequent ones.  Not sure if by design or oversight.  I'll try to spend more time on it in the next week or so when I get a chance.  Great stuff, btw!!

fungl164:
@weiznich just wondering if it would be useful to use a proper pluralizer/inflector (e.g. https://github.com/whatisinternet/Inflector) for models when running them through the mutation macros for example.

weiznich:
@fungl164 Thanks your suggestions.

> just wondering if it would be useful to use a proper pluralizer/inflector (e.g. https://github.com/whatisinternet/Inflector) for models when running them through the mutation macros for example.

I'm not sure if that's possible because I need the names currently as string literals. A maybe better solution would be to provide a way to pass custom names in there instead.

> I did notice the SQL queries pull all the fields on the first select, but not on the subsequent ones. 

What do you exactly mean here?
On each request for a table all columns are loaded. (For references to other tables only the foreign keys. As soon as you request a nested table the foreign key will be resolved to the nested object.).  Later juniper will forward only that ones that are requested.

fungl164:
> What do you exactly mean here?

If you look at the GraphQL query (below), there is no request for 'hair_color'.  However, when looking at the logs (also below), the first SELECT statement includes it as SQL field (even though it was not requested in GraphQL).  

Interestingly enough, the subsequent SELECT statements in the logs include only the exact fields as requested on the GraphQL query.   So, I was just wondering why the first SELECT did not behave the same way (e.g. exact field extraction/selection in SQL as those submitted via the GraphQL query).  Hope this makes sense...

In any case, it all works fine in the end so thanks for the wonderful work you've put in.  I just didn't know if this was the intended behavior, an optimization or a just a plain oversight in the implementation...

screen shot 2018-04-17 at 8 13 25 pm
screen shot 2018-04-17 at 8 12 29 pm

weiznich:
> If you look at the GraphQL query (below), there is no request for 'hair_color'. However, when looking at the logs (also below), the first SELECT statement includes it as SQL field (even though it was not requested in GraphQL).

This is totally expected. The current strategy is to load all fields of the table and then let juniper do the selection stuff. Only fields that represents other tables are handled different, because I think loading them would be become quite expensive. (As soon as I have landed all those PR's (this one and 2 or 3 to diesel) required to make wundergraph working I need to write some blog post about this :wink:)

@theduke @mhallin 
I've added some documentation to the newly added code and fixed the failing test cases. I would be happy to get a hopefully final review on this.

PR #140, opened 2018-02-08
Change pretty json string to compact json string

sackery:
Cause of serde_json::to_string_pretty() is unefficient.


theduke:
True that. Probably a relic from debugging.

Thanks.

PR #144, opened 2018-02-18
Support std::time types via humantime

tailhook:
Hi,

I've noticed two things in current juniper:
1. It doesn't support stdlib's SystemTime
2. Even if I want to use chrono's types, the `time` crate it depends on doesn't work on wasm (and I do use juniper on wasm32-unknown-unknown)

Also, I've recently added timestamp parsing and formatting to `humantime`, so I feel it's a good match here. Here are [benchmarks](https://users.rust-lang.org/t/whats-everyone-working-on-this-week-7-2018/15587/17) for humantime against chrono. And [binary size/dependency tree comparison](https://github.com/sebasmagri/env_logger/pull/67#issue-169754040) (this is on different use case so take it with a grain of salt).

It looks like timestamps in the form of rfc3339 are more or less standard in graphql community. But I'm less sure for durations. While durations like humantime represents are good for configuration files they are less convenient for calculations. So we may favor millisecond representation for javascript. I've put Duration support in different commit which we can remove it.

What do you think?

### DIFF
--- a/juniper/Cargo.toml
+++ b/juniper/Cargo.toml
@@ -40,6 +40,7 @@ serde = { version = "1.0.8" }
 serde_derive = {version="1.0.2" }
 
 chrono = { version = "0.4.0", optional = true }
+humantime = { version = "1.1.0", optional = true }
 serde_json = { version="1.0.2", optional = true }
 url = { version = "1.5.1", optional = true }
 uuid = { version = "0.5.1", optional = true }
--- a/juniper/src/integrations/humantime.rs
+++ b/juniper/src/integrations/humantime.rs
@@ -1,0 +1,40 @@
+use std::time::{SystemTime};
+
+use humantime::{parse_rfc3339_weak, format_rfc3339};
+
+use Value;
+
+graphql_scalar!(SystemTime as "DateTime" {
+    description: "DateTime represented as a RFC3339 string (in UTC)"
+
+    resolve(&self) -> Value {
+        Value::string(format_rfc3339(*self).to_string())
+    }
+
+    from_input_value(v: &InputValue) -> Option<SystemTime> {
+        v.as_string_value().and_then(|s| {
+            // Uses permissive parser for input
+            parse_rfc3339_weak(s).ok()
+        })
+    }
+});
+
+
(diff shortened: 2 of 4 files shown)


theduke:
Out of curiosity, what's your use case for using Juniper on wasm?

theduke:
Also, the only issue I have with this is the extra dependency.
I'm definitely not comfortable merging this with a beta dependency, so we would at least need to wait for a stable release.

But regardless, I prefer to be conservative with dependencies.
The preferable option here for me would be to only implement this for unix timestamps (as float), but that is problematic as well.

I'll think about it. ;)

tailhook:
> Out of curiosity, what's your use case for using Juniper on wasm?

A very early work-in-progress [code is here](https://github.com/mglawica/kokkupanek). Which boils down to:

1. We have an orchestration system, which needs embedded scripting.
2. We use webassembly/[wasmi](https://crates.io/crates/wasmi) for scripting [verwalter](http://verwalter.readthedocs.io/en/latest/)
3. We use rust as a "scripting" language
4. And we need to send some "user actions" to that embedded language. So we use graphql/juniper for it.

(I use "scripting" here very loosely: i.e. I need a sandboxed language that can be reloaded at runtime)

> I'm definitely not comfortable merging this with a beta dependency, so we would at least need to wait for a stable release.

Sure. My idea was to make few pull requests to find out whether current API is good enough and what I've done fits some use cases well. The plan is to release humantime shortly. As you can see the [pull request](https://github.com/sebasmagri/env_logger/pull/67) to `env_logger` is also waiting for a release.

theduke:
I'm sorry, @tailhook , and thanks for the PR, but I don't think this functionality belongs in Juniper in this form.

There is no standard for parsing time durations, and implementing a default implementation for SystemTime is also questionable to me. (Of course the same thing can be said to some degree for the chrono implementations, but still...).

Also we really need to be careful with default implementations and also dependencies.
The default ones (uuid, chrono, ...) are already a bit of a pain due to version updates.

If someone needs this functionality, it is not that hard to implement this yourself for your project with a wrapper type.

PR #145, opened 2018-02-18
Fix merging parallel inline fragments

carlosdp:
Currently, this:

```
query Query {
 user {
   ... Fragment
 }
}

fragment Fragment {
 a
 ... on User {
  b
 }
}
```

returns

```
{
  "user": {
    "b": <value>
  }
}
```

overriding the previous parallel fragment definition. This issue was fixed for pre-defined parallel fragments in https://github.com/graphql-rust/juniper/commit/c5552419780a8d4fccf5c34c5c24bc976829fd4e, the merge logic just hadn't been applied to inline fragments.

This adds that logic (a 1 line change) and adds a test.

### DIFF
--- a/juniper/src/executor_tests/executor.rs
+++ b/juniper/src/executor_tests/executor.rs
@@ -203,12 +203,22 @@ mod merge_parallel_inline_fragments {
     use types::scalars::EmptyMutation;
 
     struct Type;
+    struct Other;
 
     graphql_object!(Type: () |&self| {
         field a() -> &str { "Apple" }
         field b() -> &str { "Banana" }
         field c() -> &str { "Cherry" }
         field deep() -> Type { Type }
+        field other() -> Vec<Other> { vec![Other, Other] }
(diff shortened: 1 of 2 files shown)


carlosdp:
There's still a bug when doing something like this:

```
fragment Fragment {
  deep {
    a
  }
  ... on Type {
    deep {
      b
    }
  }
}
```

where only `a` shows now. Looking into it.

carlosdp:
This still has a bug when merging lists, working on it.

carlosdp:
Fixed.

theduke:
@carlosdp good catch, thanks.

I'm always amazed that so many users don't open an issue but just submit a fix. 

Tests seem to be failing due to a nightly issue, i'll investgate.

LegNeato:
Anything more to do here or are we good to merge?

theduke:
I have re-started the build and will merge once it succeeds.

Issue #150, opened 2018-02-22
Chronos NaiveDate Serialisation / Deserialisation


sporto:
I was trying to use `NaiveDate` as a scalar for a bookings application and was surprise to find out that I need to pass a string like `2018-01-01T16:39:57-08:00` for it to work. 

If I understand correctly `NaiveDate` is what you would use when you want a date independent of the timezone. When using a `Date` column type in postgres, using NaiveDate in a struct seems like a good fit.

So when using NaiveDate as a scalar I expected NaiveDate to serialize to `2018-01-01`. And be able to deserialize from that.

Maybe I'm wrong on my expectation of NaiveDate. Is there a type better suitable for only dates? Or maybe NaiveDate should only serialize to the date part.

Thanks

theduke:
Fully agreed.
This was an oversight when this was implemented.
Will be a small fix: https://github.com/graphql-rust/juniper/blob/master/juniper/src/integrations/chrono.rs

sporto:
Done in https://github.com/graphql-rust/juniper/pull/151

PR #151, opened 2018-02-23
Change serialization of Chronos NaiveDate

sporto:
https://github.com/graphql-rust/juniper/issues/150

`NaiveDate` should serialize / deserialize from "2018-01-02" instead of a full ISO date.

### DIFF
--- a/juniper/src/integrations/chrono.rs
+++ b/juniper/src/integrations/chrono.rs
@@ -19,7 +19,6 @@ use Value;
 
 #[doc(hidden)]
 pub static RFC3339_FORMAT: &'static str = "%Y-%m-%dT%H:%M:%S%.f%:z";
-static RFC3339_PARSE_FORMAT: &'static str = "%+";
 
 graphql_scalar!(DateTime<FixedOffset> as "DateTimeFixedOffset" {
     description: "DateTime"
@@ -82,9 +81,7 @@ graphql_scalar!(NaiveDateTime {
 
 #[cfg(test)]
 mod test {
-    use super::RFC3339_PARSE_FORMAT;
     use chrono::prelude::*;
-    use Value;
 
     fn datetime_fixedoffset_test(raw: &'static str) {
         let input = ::InputValue::String(raw.to_string());
@@ -136,8 +133,12 @@ mod test {
         datetime_utc_test("2014-11-28T21:00:09.005+09:00");
     }
 
-    fn naivedate_test(raw: &'static str, y: i32, m: u32, d: u32) {
-        let input = ::InputValue::String(raw.to_string());
+    #[test]
+    fn naivedate_from_input_value() {
+        let input = ::InputValue::String("1996-12-19".to_string());
+        let y = 1996;
+        let m = 12;
+        let d = 19;
 
         let parsed: NaiveDate = ::FromInputValue::from_input_value(&input).unwrap();
         let expected = NaiveDate::from_ymd(y, m, d);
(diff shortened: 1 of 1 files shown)


juniper/src/integrations/chrono.rs
```diff
@@ -157,17 +150,16 @@ mod test {
...
+        naivedate_test("1996-12-19", 1996, 12, 19);
     }
 
-    #[test]
-    fn naivedate_from_input_value_with_fractional_seconds() {
-        naivedate_test("1996-12-19T16:39:57.005-08:00", 1996, 12, 19);
-    }
+    fn naivedate_serialization() {
+        let date = NaiveDate::from_ymd(2000, 1, 2);
+        // let value = date.serialize().unwrap();
```
sporto: @theduke can you tell me what needs to be called in order to serialize a `NaiveDate`. So convert this to an `InputValue` using the `resolve` function. Thanks

LegNeato: Something like https://github.com/graphql-rust/juniper/blob/master/juniper/src/integrations/url.rs#L23 ?

juniper/src/integrations/chrono.rs
```diff
@@ -157,17 +151,14 @@ mod test {
...
-    }
-
-    #[test]
-    fn naivedate_from_input_value_with_z_timezone() {
-        naivedate_test("1996-12-19T16:39:57Z", 1996, 12, 19);
+    fn naivedate_serialization() {
+        let date = NaiveDate::from_ymd(2000, 1, 2);
+        let value = date.resolve();
+        let expected = Value::string("2000-01-02".to_string());
+        assert_eq!(value, expected);
```
sporto: @LegNeato there is a test above that tests that given an input value it parses correctly and returns a NaiveDate. This https://github.com/graphql-rust/juniper/blob/master/juniper/src/integrations/url.rs#L23 looks similar.

I would like a test that given a NaiveDate it tests that it serialises correctly. Something that calls the `resolve` function to serialize. I haven't found an example test that do this.

I'm the app I'm building I'm getting a panic when trying to serialize NaiveDate, unsure what is wrong, but knowing how to test the serialization will help to find the issue.

juniper/src/integrations/chrono.rs
```diff
@@ -157,17 +151,17 @@ mod test {
...
-        naivedate_test("1996-12-19T16:39:57.005-08:00", 1996, 12, 19);
-    }
+    fn naivedate_serialization() {
+        use types::base::GraphQLType;
 
-    #[test]
-    fn naivedate_from_input_value_with_z_timezone() {
-        naivedate_test("1996-12-19T16:39:57Z", 1996, 12, 19);
+        let date = NaiveDate::from_ymd(2000, 1, 2);
+        let value = date.resolve(&(), None, &executor::Executor);
```
sporto: I figured out what to pass for the first two args, but don't know about the third.
Any help on this? Thanks

sporto: @LegNeato @theduke can I get your help on this please? Thanks

LegNeato: @sporto I did a PR to your PR with some integration tests. Feel free to structure them another way.

LegNeato:
Please update the doc comment at the top of the file and add this to the changelog.

This is a potentially breaking change if someone is using a non-rust client or storing results...so we should probably bump to `0.10`. Thoughts @theduke?

sporto:
@LegNeato this PR is ready for review. Thanks

LegNeato:
@sportoi sadly I don't have commit privs so can't review and merge. We'll have to wait for @theduke or @mhallin.

theduke:
@sporto thanks and sorry for the delay.

PR #159, opened 2018-03-19
Switch to indexmap 1.0

LegNeato:
`ordermap` changed names to `indexmap` and released a stable 1.0.

### DIFF
--- a/juniper/Cargo.toml
+++ b/juniper/Cargo.toml
@@ -35,7 +35,7 @@ default = [
 juniper_codegen = { version = "0.9.2", path = "../juniper_codegen"  }
 
 fnv = "1.0.3"
-ordermap = { version = "0.2.11", features = ["serde-1"] }
+indexmap = { version = "1.0.0", features = ["serde-1"] }
 serde = { version = "1.0.8" }
 serde_derive = {version="1.0.2" }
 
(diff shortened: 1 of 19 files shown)


sagiegurari:
looks like cargo nightly version is broken

sagiegurari:
open issue to cargo rust-lang/cargo#5212

theduke:
I'll re-trigger the build tomorrow and then merge.

sagiegurari:
builds won't pass on nightly until both rustc and cargo are fixed.
you could use nightly of February or make allow failures in build for nightly for builds to pass.

theduke:
The PR has been merged, no? 
If so so let's hope that tomorrows nightly is fixed.

sagiegurari:
unfortunately, cargo fix keeps failing in their build and as for rustc, they currently don't know the reason the incremental builds keep failing.
you could disable the incremental builds via env var to workaround that issue, but the cargo issue is still not solved.

LegNeato:
There's no rush on this, we can sit back until nightly is fixed.

theduke:
@LegNeato thanks again!

I don't want to delay this any further so I'll merge right now, but can you do another PR updating the changelog?

PR #164, opened 2018-04-01
Add actix-web to readme

user453584:
Add actix-web to readme.

Might be nice in the future to add an example like this one for reference as well: https://github.com/actix/actix-web/tree/v0.4.0/examples/juniper

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ your Schemas automatically.
 
 * [rocket][rocket]
 * [iron][iron]
-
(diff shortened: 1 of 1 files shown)


theduke:
Hey, sorry for the big delay.

Integration with Rocket and Iron is covered in the examples in this repo and in the book, so in this case it might be confusing without a direct link to the example in Actix.

theduke:
@user453584 I've added a link to the actix-web example in the README.

Thanks.

PR #165, opened 2018-04-06
Add support for BigDecimal with a feature flag

ambaldwin21:


theduke:
Hey @ambaldwin21 , thanks for the PR.

I'm afraid I'm sceptical about accepting this.
The other integrations we have are with popular and well-established crates.
BigDecimal is rather young and not widely used.
Accepting it as an integration would mean that we are stuck with it.

For now, if this is something you need often, you could publish a `juniper_bigdecimal` crate with a wrapper type `BigDecimal(::bigdecimal::BigDecimal)` that implements the required Juniper traits.

sporto:
It seems like Diesel has adopted bigdecimal for the numeric PG type. As numeric is widely used and Diesel is very likely to be used with Juniper, maybe this integration should be re considered.

LegNeato:
Moving discussion to https://github.com/graphql-rust/juniper/issues/235

PR #170, opened 2018-04-29
Add Compile Time Check For "Invalid" Names

piperRyan:
This doesn't remove any run time checks. RegExp can be found in the util file if I overlooked a case let me know!

### DIFF
--- a/juniper_codegen/Cargo.toml
+++ b/juniper_codegen/Cargo.toml
@@ -16,6 +16,8 @@ proc-macro = true
 [dependencies]
 syn = "0.11"
 quote = "0.3"
+regex = "0.2.10"
+lazy_static = "1.0.0"
 
 [badges]
 travis-ci = { repository = "graphql-rust/juniper" }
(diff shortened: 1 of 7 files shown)


theduke:
@piperRyan thanks for the PR, but I'm afraid we can't really do things like that.

While the checks you are adding are in line with the conventions, the spec does not enforce any of those constraints, and neither can/should we.

There could potentially be some strict mode that validates these checks, but it definitely needs to be optional.

LegNeato:
Personally I think just going the spec regex route will suffice

theduke:
Jup, validating according to the spec (http://facebook.github.io/graphql/October2016/#Name) would be fine.

piperRyan:
@theduke updated regex to be aligned with specs. I am playing around with the older macros to see if I can get the compile_error! macro to work alongside it, but my understanding of macros is still a bit weak so it might be some time before I can get it right assuming one can do so.  In meantime let me know if there is any other concerns with this PR.

piperRyan:
Yup! I would just be need to update the [master.md](https://github.com/graphql-rust/juniper/blob/master/changelog/master.md) file correct?

LegNeato:
Yep! I want to revamp how it is done eventually but for now just stick it in there.

piperRyan:
Done! Let me know if you want me to make any further changes.

LegNeato:
Thanks again for the PR and updating it a bunch of times with the requested changes! 🍻 🎉

PR #179, opened 2018-05-08
Deserialize large integers as InputValue::float (fixes #178)

### LINKED ISSUE #178: Floating point numbers issue
Hi, I came across the issue that all numbers without decimal part are treated as i32, which is a problem for handing floating point numbers. I.e. if I need large floating point number without a decimal part it's encoded as an integer by browsers:

```
>>> JSON.stringify({"x": 2.0 ** 35})
{"x":34359738368}
```

This means that you can't decode that json as `InputValue` reliably. As it will fail with the error like in #52.

I think treating everything that is > i32::max as floating point is fine, for InputValue? Or you you see any shortcommings of this approach?

tailhook:


LegNeato:
This looks good, but can you foresee any issues with non-browser JSON clients? I guess before we just threw an error so this is materially better...

LegNeato:
Also, can you add the change to the changelog?

tailhook:
> This looks good, but can you foresee any issues with  non-browser JSON clients? I guess before we just threw an error so this  is materially better...

I don't think there should be any issues. Python has the distinction between types. I.e. it can encode something as `1.0` but this works perfectly with both old and new version.

> Also, can you add the change to the changelog?

Added. But I'm not sure it's clear enough.

PR #180, opened 2018-05-10
(juniper_iron) fallible context factory

andy128k:


andy128k:
@theduke is there any reason not to break compatibility while juniper is versioned as 0.x?

theduke:
No, just mentioning it for reference and as a reminder to myself.

andy128k:
@LegNeato I've added this to changelog.
I was not sure about date and version, so I've put next minor version and a date of a commit.

PR #181, opened 2018-05-14
[WIP] Add impl derive for graphql objects

ForsakenHarmony:
https://gist.github.com/ForsakenHarmony/8d42d4963dec52d80853e48ff7401ef1#file-proc_macro-rs

### DIFF
--- a/juniper/src/schema/meta.rs
+++ b/juniper/src/schema/meta.rs
@@ -786,9 +786,9 @@ fn clean_docstring(multiline: &[&str]) -> Option<String> {
                     .map(|ch| ch.is_whitespace())
                     .unwrap_or(false)
                 {
-                    ln.trim_right() // skip trimming the first line
+                    ln.trim_end() // skip trimming the first line
                 } else if ln.len() >= trim_start {
-                    ln[trim_start..].trim_right()
+                    ln[trim_start..].trim_end()
                 } else {
                     ""
                 };


theduke:
Love it. Exactly what I wanted to add.

Sadly, this will have to wait until proc macros hit stable.
We could feature gate it behind a nightly feature though.

theduke:
I will do a more thorough review soon, but some short hints:

* we can probably use https://crates.io/crates/proc-macro-hack to make it work on stable

It would be nice to specify the context only once.
Something in the vain of:

```rust
#[graphql_object]
impl GraphQLObject for MyType<Context=SomeType> {
}
```

ForsakenHarmony:
proc-macro-hack is for function like macros?
this is an attribute proc macro

and sure that would be something to think about

ForsakenHarmony:
welp, I just upgraded this to the latest `syn` and `quote` as well

ForsakenHarmony:
@theduke ?

LegNeato:
Does proc_macro being stabilized change this? Of course, it will still have to be gated because we don't want to bump our rust requirement that high yet so maybe not...

ubnt-intrepid:
I think it is not good to specify the context type as a bounds of associated types.
The syntax is different to the bindings of associated type which is usually used in the impl and hence not intuitive.

I thinks it is prefer to be written as follows:

```rust
#[graphql]
impl MyType {
    type Context = ();
    ...
}
```

In order to avoid confusion, we should clarify the specification of the attribute-style macro to be created...

ForsakenHarmony:
Good idea

theduke:
@ForsakenHarmony are you motivated to finishing this up?

ForsakenHarmony:
If you tell me what exactly you want

theduke:
A few things left to do:

* rebase 
* handle the new custom scalar generic 
* remove the nightly cfg stuff
* add a few tests similar to the other custom derive tests in `integration_tests/juniper_tests/codegen`
* specify context with `type Context = *`
* require `impl GraphQLObject for X` (see below)

I can absolutely help out with any of those, you just say if and what you are able to tackle. ( it's great that you started this in the first place!)

=== 
I think it would be nice to make it clear in the code that you are implementing GraphQLObject, so it could look something like this:

```rust
#[graphql_object]
impl GraphQLObject for MyType {
  type Context = MyCtx;

  ...
}

```

ForsakenHarmony:
I'll take a look

Swoorup:
Does this allow graohql queries/mutation to be split across multiple files?

ForsakenHarmony:
Nope

theduke:
@ForsakenHarmony what was contained in your recent push, and what's still TODO?

ForsakenHarmony:
it was just a rebase, pretty much everything, because you didn't respond before

ForsakenHarmony:
Oh right, this is a thing

But I guess my ide would also show an error for the rest of the functions if it was a trait implementation

theduke:
Ah, I just noticed something else due to the above issue:

It would be quite confusing to have both a `graphql_object!` macro and a `gql_object` proc macro.

We should find a name that distinguishes the two clearly.

Since with edition 2018 macros are not global and can be imported, my preference would be: `impl_object`

```rust
#[juniper::impl_object]
impl Query{
  ...
}
```

ForsakenHarmony:
what's the `new custom scalar generic on GraphQLObject`?

I'll probably do that and the `type Context` but can't say if I'll have time before the weekend

ForsakenHarmony:
well, it might be better if you just implement it, don't have too much time and am not entirely sure what the best way is

theduke:
Continued in https://github.com/graphql-rust/juniper/pull/333 due to Github being buggy.

ForsakenHarmony:
seems like you accidentally force pushed master or smth, making this empty

did that once as well

theduke:
Wasn't an accident,  I squashed your commits together.
Apparently Github can't handle it if the PR is in the master branch and the original base commit is lost.

Jedenfalls nochmal danke. ;)

Issue #191, opened 2018-05-30
CI broken on 1.21.0 due to cargo-make


LegNeato:
Looks like `cargo-make` dropped support for 1.21.0:

https://github.com/sagiegurari/cargo-make/commit/8a2372305ddbf53d53bc9e3253f0baf2a1b1b9e3

This is causing CI to fail.

PR #192, opened 2018-05-30
Pin the version of `cargo-make` in CI.

LegNeato:
Newer versions of `cargo-make` drop support for rust 1.21.0. We
still support 1.21.0, so we'll pin cargo-make until we drop 1.21.0
support.

Fixes https://github.com/graphql-rust/juniper/issues/191


LegNeato:
Note the use of `vers` here as `version` wasn't added until later.

sagiegurari:
sorry about that. it was due to rand upgrade to 0.5.0 which in turn wasn't backwards compatible and didn't support that rust version anymore.
the current stable version is 1.26.1 so it was reasonable enough for me to be ok with that as it means that I currently support the latest 7 rust versions (counting nightly and beta)

LegNeato:
No worries!

sagiegurari:
for linux only you can use prebuilt versions.
for example
wget from https://bintray.com/sagiegurari/cargo-make/download_file?file_path=cargo-make_v0.11.0

Issue #199, opened 2018-06-20
Performance issues


weiznich:
I'm currently in the process of cleaning up the implementation of wundergraph. As part of this I've ported [this graphql orm](https://github.com/hasura/graphql-backend-benchmarks) benchmark to a diesel/wundergraph/juniper/actix_web stack.
This lead to the following results:
![image](https://user-images.githubusercontent.com/1674512/41662707-36d0ecda-7491-11e8-89d1-9ac1af0bd5b9.png)
The corresponding query is
```graphql
query tracks_media_all {
  Tracks {
    id
    name
    media_type_id {
      name
    }
  }
```
which will be translated to the following sql by wundergraph:
```sql
select * from tracks;
select * from media_types where id in (list_of_media_ids_from_tracks);
```

So I started to dig into this. 
After several tries I've used perf to profile the execution of that query. As a result of that I've found out that more then 50% of the execution time is spend in [this function](https://github.com/graphql-rust/juniper/blob/master/juniper/src/types/base.rs#L325). The main issue seems that some `IndexMap` in there gets reconstructed several times and that those reconstructions do not preallocate the corresponding size.

Code: [Wundergraph benchmark](https://github.com/weiznich/wundergraph/tree/infer_schema%2Bbench/wundergraph_bench) (Needs the [postgres variant of the chinook database](https://github.com/xivSolutions/ChinookDb_Pg_Modified/tree/pg_names))

LegNeato:
Thanks for the detailed report! I'll try to repro

weiznich:
I've tried to dig into this a bit further.
The current state are [this commit to juniper](https://github.com/weiznich/juniper/commit/ebe6dd65f49b26f0ca7371776d5ca47a5b6c4824) and [this commit to wundergraph](https://github.com/weiznich/wundergraph/commit/5917fe5b34e2fbc32ee59c585d7ed3bb7c5554c0). Both contain changes that are targeting juniper. That said I'm not sure if everything I've done there is sane or if something will break...
Using this changes it was possible to get on par with hasura at least for 50 and 100 requests per second.

LegNeato:
Obvious thing, but are you testing in release mode with LTO?

weiznich:
> Obvious thing, but are you testing in release mode with LTO?

Yes that's with all optimization enabled. (I enabled even `target-cpu=native`)

That said here are some more information about the test case:
[...]

PR #202, opened 2018-06-24
Some performance improvements

weiznich:
* Replace the IndexMap in the serialized object with a plain
  `Vec<(String, Value)>` because linear search is faster for few
  elements

* Some general tweaks to skip some allocations

Addresses #199

### DIFF
--- a/juniper/src/value.rs
+++ b/juniper/src/value.rs
@@ -1,8 +1,7 @@
-use indexmap::IndexMap;
-use std::hash::Hash;
-
 use ast::{InputValue, ToInputValue};
 use parser::Spanning;
+use std::iter::FromIterator;
+use std::vec::IntoIter;
 
 /// Serializable value returned from query and field execution.
 ///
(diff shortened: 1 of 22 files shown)


LegNeato:
@theduke

weiznich:
The ci fails because something on the master branch is brocken for the `juniper_iron` test case.

juniper/src/value.rs
```diff
@@ -22,7 +21,115 @@ pub enum Value {
...
+        for<'a> &'a str: PartialEq<K>,
+    {
+        self.key_value_list
+            .iter()
+            .any(|(key, _)| (key as &str) == f)
+    }
+
+
+    /// Get a iterator over all field value pairs
+    pub fn iter(&self) -> impl Iterator<Item = &(String, Value)> {
```
LegNeato: Hmmm, looks like this uses impl trait, which didn't stabilize until [1.26](https://blog.rust-lang.org/2018/05/10/Rust-1.26.html). Supporting impl trait would mean we would have to bump our min rust version.

theduke:
Regarding versions, I'd like to stick to the `stable + last 2 versions` rule we have right now, so impl trait would have to wait for 1.28.

Also, we'd really need some solid benchmarks comparing the two implementations.

This might be a good oppurtunity to set up a nice [criterion](https://github.com/japaric/criterion.rs) based benchmark suite.
It also might be an opportunity to contribute to/ improve the indexmap crate.

theduke:
Either way, as a first step,  it might be a cool idea to export a wrapper type from a submodule so we can easily change the implementation without a lot of hassle.

weiznich:
(I'm back from vacation, so here are some answers.)

> Regarding versions, I'd like to stick to the stable + last 2 versions rule we have right now, so impl trait would have to wait for 1.28.

It should be possible to do this without `impl trait`. I will fix the code in the next few days.

> Also, we'd really need some solid benchmarks comparing the two implementations.

I've developed this change as part of optimizing wundergraph. For this I used the [following benchmark](https://github.com/hasura/graphql-backend-benchmarks). See the wundergraph repo (wundergraph-bench) for commits [without](https://github.com/weiznich/wundergraph/commit/174e3760545fffa615402a14644a73df59dd7f32) and [with](https://github.com/weiznich/wundergraph/commit/6f44e5421d611e7f33a3ef1664754f11ff6a063d) this change. The relevant graphql query is this one:
```graphql
query tracks_media_all {
  Tracks {
    id
    name
    media_type_id {
      name
    }
  }
}
```

Benchmark results without this change:
41662707-36d0ecda-7491-11e8-89d1-9ac1af0bd5b9

Benchmark results with this change:
signal-2018-06-24-202927

> This might be a good oppurtunity to set up a nice criterion based benchmark suite.

My benchmark is unfortunately something that involves a quite complex interplay of several components so it seems not to be appropriated to be used as benchmark for juniper alone. 

> It also might be an opportunity to contribute to/ improve the indexmap crate.

This is not something that could be improved in the indexmap crate.
Basically this PR changes 3 things:

1. It introduces a wrapper type for the object enum kind. This is a purely cosmetic change to give us room to change the internal presentation later.
2. It forces every construction of that wrapper type to preallocate a fixed number of elements. This eliminates the need of continuously reallocating space for more elements as soon as we start to push more field value pairs into the object. (We know how many fields we need in every place so we could just allocate the right number of slots). This contributes ~50% of the performance improvements.
3. It replaces the use of `IndexMap<String, Value>` with a plain `Vec<(String, Value)>`. On the first look this change is making things worse because now we need to to a linear search instead of a hash lookup if we want to know if some field exists. But in my observation for almost all cases that I've seen there are only a few fields (< ~50) so a linear search is faster than doing the hash lookup. This contributes the other ~50%.

dvic:
What is the status of this PR? I believe the `impl trait` has been removed from this PR, right? Are there any remaining issues/todo's before this can be merged?

weiznich:
The usage of `impl Trait` has been already removed in 3b3d0411b8d2c07db7647fb7cfa8fa98298ea78d.
This pr is missing a final review/decision from the juniper team.

LegNeato:
Ah, didn't see this was updated, thanks for the reminder! I'm merging this as-is as I think the change is good and straightforward. I'm slightly concerned about:

1. This is basically implementing a vector-backed arena. It might be better to use one from crates.io, as there are no shortage of them with various perf and mem tradeoffs. Or better yet, have a pluggable interface as @theduke mentioned.

2. Down the road it might not be obvious why we need the arena. I'm worried without some sort of validation or perf test we could regress.

In any case, I don't think those are blockers for landing this.

PR #204, opened 2018-06-25
Don't make assumptions what features the rocket app might use

csharad:
Even after doing `default-features = false` and cherry picking features on `juniper`, `juniper-rocket` was still pulling in default features.
This disables the default features on `juniper-rocket`.


LegNeato:
This appears to be failing tests

csharad:
Even without the changes, the test is failing. Also, it is failing on `juniper_iron` whereas I made changes to `juniper_rocket`.

LegNeato:
Oh, ok. Sorry for the delay, thanks for the patch! 🍻

Issue #196, opened 2018-06-12
Update to the June 2018 GraphQL spec


LegNeato:
A new spec is out (https://github.com/facebook/graphql/releases/tag/June2018). It looks like a lot of our error handling questions / behaviors have been clarified among other things.

theduke:
We should come up with a checklist of what we have to implement.

LegNeato:
https://github.com/graphql-rust/juniper/pull/219 switched from `data` to `extensions` for custom error data.

dvic:
@LegNeato I think it would be useful to have a test suite based on https://github.com/graphql-cats/graphql-cats. That project seems to cover many scenario's, I think also beyond the June 2018 spec, but at least then we have an idea of what is supported and what is not. 

I'm not sure what the implementation would be exactly (a simple dynamic schema executor based on the scenarios'?) but the project is useful at least as a point of reference :)

theduke:
@dvic I already made an issue for this a while ago.

https://github.com/graphql-rust/juniper/issues/245

dvic:
@LegNeato Ah nice, next time I'll make sure to do a search ;) Maybe I have this weekend time to have a look at it to see what we can do with it.

tyranron:
@LegNeato I guess this has been fixed in #631?

LegNeato:
Yes, this should be fixed.

PR #205, opened 2018-07-08
Allow custom errors to be returned from queries, mutations

### LINKED ISSUE #40: Complex Errors

csharad:
Fixes #40.

This is not a breaking change, as `FieldResult` still works. For use cases which desire for custom error handling that return additional response fields ergnomically, those error types should implement `IntoFieldError` trait which converts your error type to the FieldError. This maintains that in the response, `message` field is always returned along with other optional user desired metadata.

You may implement your error as:
```
enum CustomError {
  NotFound
}

impl IntoFieldError for CustomError {
  fn into_field_error(self) -> FieldError {
    FieldError::new("Not Found", graphql_value!({ "type": "NOT_FOUND" })
  }
}
```

And use it likewise:
```
field add_likes(&executor, id: i32) -> Result<Like, CustomError> {
  Err(CustomError::NotFound)
}
```

### DIFF
--- a/changelog/master.md
+++ b/changelog/master.md
@@ -28,3 +28,9 @@
   `#[graphql(description = "my description")]`.
 
   [#194](https://github.com/graphql-rust/juniper/issues/194)
+
+* Introduced `IntoFieldError` trait to allow custom error handling
+  i.e. custom result type. The error type must implement this trait resolving
+  the errors into `FieldError`.
+
+  [#40](https://github.com/graphql-rust/juniper/issues/40)
--- a/juniper/src/executor/mod.rs
+++ b/juniper/src/executor/mod.rs
@@ -193,6 +193,21 @@ pub type ExecutionResult = Result<Value, FieldError>;
 /// The map of variables used for substitution during query execution
 pub type Variables = HashMap<String, InputValue>;
 
+/// Custom error handling trait to enable Error types other than `FieldError` to be specified
+/// as return value.
+///
+/// Any custom error type should implement this trait to convert it to `FieldError`.
+pub trait IntoFieldError {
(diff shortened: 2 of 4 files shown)


LegNeato:
Thanks for the change! Have you seen https://github.com/graphql-rust/juniper/issues/196 / https://github.com/facebook/graphql/releases/tag/June2018? There are a lot more requirements for errors in the newest spec. Does the newest spec obsolete this PR?

csharad:
@LegNeato I read the error section of the specification and one of the change it outlines is to use `extensions` field to map out additional metadata to the error. We are currently using `data` field in its place which needs to be refactored and that could be done in another PR.

As for this PR, it only deals with mapping our own error to the `FieldError`. So, I think the spec does not have a problem with it.

LegNeato:
Great, thanks! 🍻 . Sorry this took so long to merge.

PR #207, opened 2018-07-10
Consistent error serializing for GraphqlError

tailhook:
Graphql spec requires every error contain `{"message": "field"}`. Every
error in juniper except ones fixed here are reported consistently with
this style. This commit fixes the representations of the three errors left.

While technically this commit can be treated as breaking, I think this issue is an oversight rather than intentional and should be fixed as a bug.

What do you think?


LegNeato:
Looks great, thanks! I'm not worried about it being a breaking change as long as we are closer to the spec.

theduke:
@tailhook :+1:

PR #212, opened 2018-07-19
Implement GraphQLType for Arc

tailhook:
I don't see any reasons why GraphQLType is implemented for Box bot not implemented for Arc. Also added ToInputValue.

I don't add `FromInputValue` for same reason that serde is not implemented for `Rc` and `Arc` by default: one my expect same objects to be deserialized to the same Arc which isn't true. I'm not entirely sure of whether this precaution is still useful for graphql though.

### DIFF
--- a/changelog/master.md
+++ b/changelog/master.md
@@ -34,3 +34,7 @@
   the errors into `FieldError`.
 
   [#40](https://github.com/graphql-rust/juniper/issues/40)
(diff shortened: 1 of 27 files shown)


theduke:
Thanks again. ;)

theduke:
@tailhook never mind, did it myself.

PR #216, opened 2018-08-03
Add integration crate for the warp framework

tomhoule:
I think this still needs more testing but coming up with this integration was very smooth and it feels like idiomatic warp (I've only been programming with warp since its first public release, so yesterday :)).

I'll keep working on polishing this over the weekend (docs, tests) but feedback would still be appreciated.

This notably does not have support for batch requests. Should it be implemented in this PR or can it be done separately?


juniper_warp/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+///     .and(warp::post(graphql_filter));
+/// # }
+/// ```
+pub fn make_graphql_filter<Query, Mutation, Context>(
+    schema: juniper::RootNode<'static, Query, Mutation>,
+    context_extractor: BoxedFilter<(Context,)>,
+) -> BoxedFilter<(impl warp::Reply,)>
+where
+    Context: Send + Sync + 'static,
+    // CtxRef: AsRef<Context> + Send + 'static,
```
tomhoule: this should be deleted

juniper_warp/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+    Query: juniper::GraphQLType<Context = Context, TypeInfo = ()> + Send + Sync + 'static,
+    Mutation: juniper::GraphQLType<Context = Context, TypeInfo = ()> + Send + Sync + 'static,
```
tomhoule: I played with the design quite a bit, maybe not all of these bounds are needed anymore

LegNeato:
This is looking great! I'm just coming up to speed, but if you are not aware we have an integration test harness you likely want to leverage. Here's Rocket using it: https://github.com/graphql-rust/juniper/blob/master/juniper_rocket/src/lib.rs#L453

tomhoule:
I wasn't aware of that, thanks for the pointer!

juniper_warp/examples/warp_server/src/main.rs
```diff
@@ -0,0 +1,45 @@
...
+    let state = warp::any().map(move || Database::new());
+    let graphql_filter = juniper_warp::make_graphql_filter(schema(), state.boxed());
+
+    warp::serve(
+        warp::get(
+            warp::path("graphiql")
+                .and(juniper_warp::graphiql_handler("/graphql"))
+                .or(homepage),
+        ).or(warp::path("graphql").and(graphql_filter))
+        .with(log),
```
tomhoule: with a few more days of warp experience, this doesn't look like the right way to do it, I'll have to improve this. Should this be a single file example by the way? I haven't worked with those but this one is meant to be run and used in a browser

tomhoule:
Small update on this: I spent some time setting up the framework integration harness, and GET query parameters handling is a bit weird when testing a filter directly, I may have to change the tests to spawn a whole server to test against instead

tomhoule:
Ok now all tests are passing locally (I haven't seen on CI yet). There is the small matter of error handling - I introduced `failure` as a dependency for juniper_warp since that's what I was able to implement a fast solution with, but we might want to have a custom error enum.

tomhoule:
@LegNeato I think this is ready for review.

There is an issue with rust 1.21 and 1.22 in CI - some warp dependencies don't seem to compile with these versions. If latest stable is the oldest supported compiler for warp, we should probably have the same requirement for juniper-warp. I haven't looked into how that could be implemented yet. What do you think?

I'll squash before merging.

juniper_warp/src/lib.rs
```diff
@@ -0,0 +1,363 @@
...
+///     .map(|auth_header: String, app_state: Arc<AppState>| {
+///         let user_id = auth_header; // we believe them
+///         ExampleContext(app_state, user_id)
+///     })
+///     .boxed();
+///
+/// let graphql_filter = make_graphql_filter(schema, context_extractor);
+///
+/// let graphql_endpoint = warp::path("graphql")
+///     .and(warp::post(graphql_filter));
```
tomhoule: this style of method filter is going to be deprecated, I'll change this next time I touch the code

Cargo.toml
```diff
@@ -5,4 +5,6 @@ members = [
   "juniper_tests",
   "juniper_iron",
   "juniper_rocket",
+  "juniper_warp",
+  "juniper_warp/examples/warp_server",
```
LegNeato: Is this necessary? I don't believe we needed it for the other integrations.

juniper_warp/src/lib.rs
LegNeato: 👍

juniper_warp/src/lib.rs
```diff
@@ -0,0 +1,363 @@
...
+        fn get(&self, url: &str) -> TestResponse {
+            use percent_encoding::{percent_encode, DEFAULT_ENCODE_SET};
+            let url: String = percent_encode(url.replace("/?", "").as_bytes(), DEFAULT_ENCODE_SET)
+                .into_iter()
+                .collect::<Vec<_>>()
+                .join("");
+
+            // if url.len() > 200 {
+            //     panic!("{:?}", url);
+            // }
```
LegNeato: Delete.

juniper_warp/src/lib.rs
```diff
@@ -0,0 +1,363 @@
...
+                .to_str()
+                .expect("invalid content-type string")
+                .to_owned(),
+        }
+    }
+
+    impl HTTPIntegration for TestWarpIntegration {
+        fn get(&self, url: &str) -> TestResponse {
+            use percent_encoding::{percent_encode, DEFAULT_ENCODE_SET};
+            let url: String = percent_encode(url.replace("/?", "").as_bytes(), DEFAULT_ENCODE_SET)
```
LegNeato: This is fine for now, but I think we'll have the harness send valid encoded URLs so all these integrations don't have to deal with it themselves.

LegNeato:
No need to squash before merging, I can do that.

As for rust version, I'm not sure. We could not include `juniper_warp` in the workspace and test it separately on CI using only the most recent stable rust versions. We could probably support that at the CI config level, but we are also using `cargo make` so maybe that is a better spot for the split. It feels weird to bump juniper's requirement for the warp integration, so let's not do that.

I'm happy to poke at the build stuff if you don't want to...just let me know!

juniper_warp/src/lib.rs
LegNeato: Oh, I wasn't clear...I was just musing out loud! I don't think there is a bug in warp or you changes here, but it is actually an issue with the integration test harness. See https://github.com/graphql-rust/juniper/pull/230#issuecomment-414464667 for the same thing with Hyper, and I had to fix it hackily in the iron integration a bit ago in https://github.com/graphql-rust/juniper/pull/208.

sagiegurari:
you can take a look at the juniper rocket makefile.toml as an example

juniper_warp/src/lib.rs
tomhoule: Oh alright, sorry for the misunderstanding :)

tomhoule:
I'll try to get this done some time this weekend.

Just a note so we don't forget: we need some kind of thread pool, or document that this needs to be done by the user, since warp is async but juniper is not (yet).

LegNeato:
@tomhoule Awesome, no rush! The Hyper integration (https://github.com/graphql-rust/juniper/pull/230) includes a way to pass in a`futures_cpupool::CpuPool` so I would imagine we want to do the same. Not sure if there is a generic "pool" facade to be had or if one would even be necessary.

tomhoule:
tokio currently has [its own threadpool implementation](https://github.com/tokio-rs/tokio/tree/master/tokio-threadpool) that they claim is [much faster](https://github.com/tokio-rs/tokio/tree/master/tokio-threadpool#why-not-futures-cpupool) than futures-cpupool. But mainly [it is going to merge](https://github.com/tokio-rs/tokio/issues/211) with [the thread pool that will be built-in in the next major version of futures](https://rust-lang-nursery.github.io/futures-api-docs/0.3.0-alpha.3/futures/executor/struct.ThreadPool.html). This is probably what we will want to use eventually, but it's probably easier to use an external crate that does not depend on a particular version of futures (other than 0.1.*) or tokio at first.

What do you think of making the thread pool implicit in `make_graphql_filter` (the filter makes one by default), and allow passing one in for users who want more control? The advantage of the implicit default thread pool is that the juniper filter appears as purely async from that angle (and less boilerplate).

LegNeato:
FYI, we ended up dropping support for rust 1.21 to move our dep on `regex` to 1.0 in master...so we are fine requiring 1.22 here as well. Sorry for the churn and making you muck around with the build stuff!

tomhoule:
I'll push my latest changes (small improvements and some cargo make things that may not be necessary anymore) and see if I can make the Makefile.toml simpler, or even remove it.

When depending on juniper through a `path` instead of the latest version (0.9.2) I get a test failure with batched queries (which aren't implemented), so that still needs to be done but I think it's the last missing feature.

tomhoule:
So apparently even 1.23 is too old for warp. I tried fiddling with cargo make in various ways (mostly custom scripts to skip juniper_warp depending on the rust version - the scripts I put under the `condition_script` key in the Makefile.toml don't seem to run), and haven't gotten it to work. 

There is a cargo-make issue (https://github.com/sagiegurari/cargo-make/issues/110) about this. For now the only way I could make it work would be separating the test runs for juniper_warp from the rest of the crates, and invoking them directly from the `.travis.yml`. Do you have ideas on how to solve this?

tomhoule:
Oh and I should probably mention that I started using this code on a small project and it seems to be working fine so far.

LegNeato:
Let's just put multiple runs in `.travis.yml`, thanks!

tomhoule:
CI is passing now. I noticed https://github.com/graphql-rust/juniper/pull/239 , should we wait for that to be merged in case the new tests fail for juniper_warp?

LegNeato:
I just merged that, so rebase and then we will get this merged!

sagiegurari:
New cargo-make version 0.14.0 now has built in support for rust min/max/equal version condition criteria. so that might resolve your previous issues.

tomhoule:
Just found time to work on this a bit - the new http integration tests are passing now :)

juniper_warp/scripts/rustc_version_is_supported.sh
```diff
@@ -0,0 +1,9 @@
+# Exits successfully if the Rust version that is running is compatible with juniper_warp.
```
LegNeato: We don't need this file anymore, right?

tomhoule: Oh definitely, my bad. I'll also try to see if we can revert to using cargo make with the new feature.

LegNeato: We can do that in a follow-up, just delete this file for now.

tomhoule: done

LegNeato:
🎉🎉🎉🍻 🎉 🍻 

I am publishing a new juniper + all the integration crates next week...I'll publish` juniper_warp` and add you as an owner at that time.

PR #219, opened 2018-08-10
Use 'extensions' as field for error details

dvic:
The June 2018 spec has reserved `extensions` as the field for error details. This PR updates the serializer to use this name.

From http://facebook.github.io/graphql/June2018/#sec-Errors:

> GraphQL may provide an additional entry to errors with key *extensions*. This entry, if set, must have a map as its value. This entry is reserved for implementors to add additional information to errors however they see fit, and there are no additional restrictions on its contents.


dvic:
@LegNeato done 👍

LegNeato:
Awesome, thanks so much! 🍻

PR #222, opened 2018-08-13
Fix bad merge

LegNeato:


LegNeato:
I'm not sure why I can't merge this :-(

dvic:
@LegNeato does @tailhook have write permissions? If not, then that is probably the problem: 

*At least 1 approving review is required by reviewers **with write access.***

ForsakenHarmony:
also only people with write access can actually merge afaik

PR #224, opened 2018-08-16
Add support for skipping fields in GraphQL objects

LegNeato:


djc:
Drive-by review: I was looking for something like this, would be great to add it. The code changes seem simple and sensible. Probably would be nice to also mention this in the guide somewhere. (Nit: I think it would be easier to review if the `keyed_item_value()` changes were in a separate commit.)

PR #230, opened 2018-08-20
WIP: Hyper integration

### LINKED ISSUE #119: Add hyper example/integration
I would be great to have a juniper crate (or example) that integrates with "vanilla" hyper

dvic:
Fixes #119

### DIFF
--- a/juniper_hyper/src/lib.rs
+++ b/juniper_hyper/src/lib.rs
@@ -6,15 +6,17 @@ extern crate hyper;
 extern crate juniper;
 #[macro_use]
 extern crate serde_derive;
+#[cfg(test)]
+extern crate reqwest;
 extern crate serde_json;
 #[cfg(test)]
 extern crate tokio;
 extern crate url;
 
-use futures::future;
-use futures::{Future, Stream};
+use futures::{future, Future};
 use futures_cpupool::CpuPool;
 use hyper::header::HeaderValue;
+use hyper::rt::Stream;
 use hyper::{header, Body, Method, Request, Response, StatusCode};
 use juniper::http::{
     GraphQLRequest as JuniperGraphQLRequest, GraphQLResponse as JuniperGraphQLResponse,
@@ -317,58 +319,44 @@ impl Error for GraphQLRequestError {
 
 #[cfg(test)]
 mod tests {
-    use futures::future;
-    use futures::{Future, Stream};
+    use futures::{future, Future};
     use futures_cpupool::Builder;
     use hyper::service::service_fn;
-    use hyper::Client;
     use hyper::Method;
-    use hyper::{header, Body, Request, Response, Server, StatusCode};
+    use hyper::{header, Body, Response, Server, StatusCode};
     use juniper::http::tests as http_tests;
     use juniper::tests::model::Database;
     use juniper::EmptyMutation;
     use juniper::RootNode;
+    use reqwest;
+    use reqwest::Response as ReqwestResponse;
     use std::sync::Arc;
-    use tokio::runtime::Runtime;
-
     use std::thread;
     use std::time;
+    use tokio::runtime::Runtime;
 
     struct TestHyperIntegration;
 
     impl http_tests::HTTPIntegration for TestHyperIntegration {
         fn get(&self, url: &str) -> http_tests::TestResponse {
-            let client = Client::new();
-            let url = format!("http://localhost:3001/graphql{}", url);
-            let uri: hyper::Uri = url.parse().expect(&format!("url {} is invalid", url));
-            let resp = make_test_response(
-                client
-                    .get(uri.clone())
-                    .wait()
-                    .expect(&format!("failed GET {}", uri)),
-            );
-            resp
+            let url = format!("http://127.0.0.1:3001/graphql{}", url);
+            make_test_response(reqwest::get(&url).expect(&format!("failed GET {}", url)))
         }
 
         fn post(&self, url: &str, body: &str) -> http_tests::TestResponse {
-            let url = format!("http://localhost:3001/graphql{}", url);
-            let client = Client::new();
-            let uri: hyper::Uri = url.parse().unwrap();
-            let mut req = Request::new(Body::from(body.to_string()));
-            *req.method_mut() = Method::POST;
-            *req.uri_mut() = uri.clone();
-            make_test_response(
-                client
-                    .request(req)
-                    .wait()
-                    .expect(&format!("failed POST {}", url)),
-            )
+            let url = format!("http://127.0.0.1:3001/graphql{}", url);
+            let client = reqwest::Client::new();
+            let res = client
+                .post(&url)
+                .body(body.to_string())
+                .send()
+                .expect(&format!("failed POST {}", url));
+            make_test_response(res)
         }
     }
 
-    fn make_test_response(response: Response<Body>) -> http_tests::TestResponse {
+    fn make_test_response(mut response: ReqwestResponse) -> http_tests::TestResponse {
         let status_code = response.status().as_u16() as i32;
-
         let content_type = String::from_utf8(
             response
                 .headers()
@@ -377,14 +365,7 @@ mod tests {
                 .unwrap_or(vec![]),
         ).expect("Content-type header invalid UTF-8");
 
-        let body = response
-            .into_body()
-            .concat2()
-            .map(|chunk| {
-                String::from_utf8(chunk.iter().cloned().collect::<Vec<u8>>())
-                    .expect("body is not valid utf-8")
-            }).wait()
-            .unwrap();
+        let body = response.text().unwrap();
 
         http_tests::TestResponse {
             status_code,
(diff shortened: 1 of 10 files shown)


dvic:
I have a few remaining questions before I can wrap-up this PR:

* In `juniper/src/http/mod.rs`, I updated the http tests so that valid URL's are sent instead of modifying the urls in the tests (hyper does not accept unescaped query parameters, the iron test suite also had this problem). Is this OK?
* Should the integrations return a `HTTP 400 BAD REQUEST`? I know that with the GraphQL playground the UI breaks if a non-200 error is returned (it does not parse/display the returned parse error, for example). For the hyper integration I'm returning currently `422 Unprocessable Entity` for execution errors in order to distinguish from JSON parse errors (which should be `400 Bad Request`?) In any case, I think it's a good idea to document this and add a test case so that all framework integrations exhibit consistent behavior.

Remaining TODO's:
* update README to mention this module
* cleanup/document code

djc:
I happen to have written my own hyper integration this week, but it looks like yours is quite a bit more complex than mine. I'm not sure that makes sense for the example. In particular, I'm not sure it makes sense for example code to use a `CpuPool`; and I would expect this code to use `juniper::graphiql::graphiql_source` rather than embedding GraphiQL sources directly.

LegNeato:
@djc is your version public so I can check it out?

dvic:
@djc the embedded source is indeed for https://github.com/prisma/graphql-playground. Wrapping `RootNode` in an `Arc` upfront is the way to go I think because `RootNode::new` is quite [expensive](https://github.com/graphql-rust/juniper/blob/90b89f00eef551f86fba8df4f5649f9bd88b8dbd/juniper/src/schema/model.rs#L97) to perform on each request.

dvic:
@LegNeato I'm not sure why my test hangs on nightly on TravisCI, do you think this is a TravisCI issue or is there something wrong with my test? (I use nightly and for me the test passes without problems).

djc:
So I guess one question I have is, are these integrations (for hyper/iron/rocket/warp) intended to be example code so you can copy them into your own project and polish them as you see fit? Or are they intended to be libraries that you can import some stuff from into your own project? The Rocket integration seems only re-export some types from Juniper itself, whereas the Iron integration actually exposes Iron `Handler`s you can use in your project.

The customization going on in this PR makes me think that the latter approach (integrations as example code) makes more sense than the former, and is maybe more in line with the current state. But if so, the example code should optimize for readability and be as simple as possible rather than being optimized to cover all bases.

In either case, putting the source to the playground in the integration as is done here doesn't seem to make much sense. It could either live next to the GraphiQL source in Juniper proper, or not be included in the project at all. (By my more minimalist intuition, having only GraphiQL seems good enough.)

LegNeato:
@djc The goal of the integration crates is to make it as simple as possible to use Juniper with the various http servers. Ideally we want people to only worry about their graphql schema / data fetching while exposing that logic via their http server lib of choice is a trivial copy and paste of a couple of lines.

That being said, we don't want to be so prescriptive you can't plug in juniper to an service you have already defined or one where you have different non-graphql endpoints. So I think the abstraction here and in the examples is correct...we want to show people how to use it and integrate it.

LegNeato:
> In juniper/src/http/mod.rs, I updated the http tests so that valid URL's are sent instead of modifying the urls in the tests (hyper does not accept unescaped query parameters, the iron test suite also had this problem). Is this OK?

Yep!

> Should the integrations return a HTTP 400 BAD REQUEST? I know that with the GraphQL playground the UI breaks if a non-200 error is returned (it does not parse/display the returned parse error, for example). For the hyper integration I'm returning currently 422 Unprocessable Entity for execution errors in order to distinguish from JSON parse errors (which should be 400 Bad Request?) In any case, I think it's a good idea to document this and add a test case so that all framework integrations exhibit consistent behavior.

I agree we should define consistent behavior here, though it is not clear what it should be. I do think JSON parse errors should be 400 Bad Request, but it is unclear that an error returned from `execute` should be 422 or 400 or 500.

Currently `juniper_rocket` does:
* `execute` returning error: 400 Bad Request (https://github.com/graphql-rust/juniper/blob/master/juniper_rocket/src/lib.rs#L146)
* json parsing error: 400 Bad Request (https://github.com/graphql-rust/juniper/blob/master/juniper_rocket/src/lib.rs#L288)
* Internal failures: 500 Internal server failure (https://github.com/graphql-rust/juniper/blob/master/juniper_rocket/src/lib.rs#L283)
* A custom way to provide an error outside of the normal flow: 400 Bad Request (https://github.com/graphql-rust/juniper/blob/master/juniper_rocket/src/lib.rs#L191)

Currently `juniper_iron` does:
* `execute` returning error: 400 Bad Request (https://github.com/graphql-rust/juniper/blob/master/juniper_iron/src/lib.rs#L288)
* json parsing error: 400 Bad Request (https://github.com/graphql-rust/juniper/blob/master/juniper_iron/src/lib.rs#L277 and https://github.com/graphql-rust/juniper/blob/master/juniper_iron/src/lib.rs#L378)

So it looks like we are using `400 Bad Request` for everything and then using the message to say specifically what happened. Probably best to just match what is there.

juniper_hyper/src/lib.rs
```diff
@@ -20,48 +21,133 @@ use std::error::Error as StdError;
...
+) -> impl Future<Item = Response<Body>, Error = Err>
+where
+    CtxT: Send + Sync + 'static,
+    QueryT: GraphQLType<Context = CtxT> + Send + Sync + 'static,
+    MutationT: GraphQLType<Context = CtxT> + Send + Sync + 'static,
+    QueryT::TypeInfo: Send + Sync,
+    MutationT::TypeInfo: Send + Sync,
+    Err: Send + Sync + 'static,
+{
+    pool.spawn_fn(move || {
```
djc: Doesn't this technically violate the [spec guarantee](https://graphql.org/learn/queries/#multiple-fields-in-mutations) that mutation operations are always executed sequentially?

dvic: This is spawning a whole GraphQL request on the pool, so the mutations are not executed in parallel (that part is handled by Juniper of course).

dvic:
@LegNeato I will look into it in more detail, the rust nightly version is the same so I'm not sure what is going on. I'll keep you posted.

dvic:
It still works on my Mac nightly but I was able to reproduce the issue on a docker container, using: 
`docker run --rm -ti -v $PWD:/src rustlang/rust:nightly bash`

It's very weird, sometimes the tests pass all, sometimes it blocks on the first client call, sometimes on the second, etc.

I'm not sure if I can figure out, with my current knowledge of Rust/Hyper, the exact cause of this issue. Is there maybe another way to test Hyper handlers? (i.e., not having to startup an actual server).

LegNeato:
Hmm, definitely sounds like a race to a deadlock. I'm curious, does running with `cargo test -- --test-threads=1` fix it? I think this might actually point to a real issue so still want to dig in a bit more...hyper tests their stuff fine on travis/docker.

dvic:
@LegNeato The problem remains when using `cargo test -- --test-threads=1`.

LegNeato:
So I debugged this a bit but haven't found a smoking gun. One thing is that it hangs even on the first request and even when doing one request, so it isn't an interaction with multiple requests in flight.

When it fails it never even gets to the hyper server code. I saw the test pass / not hang when increasing the bind/server start timeout. So my current thinking is in docker it takes a lot of time to spin up the server / bind, and the `futures::lazy` is never called because it is dropped before. Not sure though, and some tests I did that I expected to fix it did not. Sorry, I'm not super familiar with futures so have generally been doing printf/commenting out debugging.

dvic:
@LegNeato Ah probably it had to do with the fact that I was blocking using the hyper client, which causes a deadlock because that future is being run on the same (tokio) runtime? Anyways, I replaced the http client with `reqwest` and now it's working on nightly, but beta is failing for some other (unrelated) reason:
```
The following packages have unmet dependencies:
 cmake3 : Depends: libjsoncpp0 (>= 0.6.0~rc2) but it is not installable
E: Unable to correct problems, you have held broken packages.
[cargo-make] ERROR - Error while executing command, exit code: 100
[cargo-make] WARN - Build Failed.
[cargo-make] ERROR - Error while executing command, exit code: 1
[cargo-make] WARN - Build Failed.
[cargo-make] ERROR - Error while executing command, exit code: 1
[cargo-make] WARN - Build Failed.
```

dvic:
In addition, I have applied the following changes:
* removed playground endpoint as @djc suggested
* switched to HTTP 400 (to better align with the iron and rocket integrations)
* updated the README files
* used more sane defaults in the hyper example (no hardcoded pool size, serve `graphiql` at root)

@LegNeato Let me know if I forgot something or anything else needs fixing.

LegNeato:
Nope, looks great! Thanks so much 🎉.

Do you want to publish to crates.io and give me permissions or shall I publish when I update juniper and give you permissions?

dvic:
@LegNeato You can go ahead and publish when you update juniper, thanks for all your help! 👍

PR #231, opened 2018-08-23
Upgrade juniper_codegen dependencies

djc:


LegNeato:
Thanks for this PR! I was dragging my feat on this an love that you came up with a PR. 🎉

djc:
So I changed the initial upgrade commit to keep the structure as before (because you're correct that is the minimal change), but then also added a commit that changes the code to the (simpler, in my mind) direct extend structure again. Feel free to skip that commit if you disagree it's an improvement. Also fixed some typos I found during the fixes.

djc:
Replaced 1.21 with 1.23 for Travis CI and added a note to the changelog about it. I couldn't find any mention of the minimum supported Rust version in README, so I left that for now -- would be happy to do a follow-up, though.

PR #238, opened 2018-08-31
[juniper_rocket] Expose GraphQLResponse fields

maxdeviant:
This PR exposes the fields of `GraphQLResponse` to make it easier to create custom responses.

##### Example

Here we're building a custom Rocket response so we can set some `Access-Control-*` headers:

```rust
fn execute_graphql_request<'a>(
    connection: DatabaseConnection,
    request: GraphQLRequest,
) -> Response<'a> {
    let context = Database::new(connection);
    let schema = Schema::new(Query, Mutation);
    // Prior to this change, this fails to compile with an error:
    // "constructor is not visible here due to private fields".
    let GraphQLResponse(status, json) = request.execute(&schema, &context);
    Response::build()
        .raw_header("Access-Control-Allow-Origin", "*")
        .raw_header("Access-Control-Allow-Methods", "OPTIONS, GET, POST")
        .raw_header("Access-Control-Allow-Headers", "Content-Type")
        .header(ContentType::new("application", "json"))
        .status(status)
        .sized_body(Cursor::new(json))
        .finalize()
}
```

### DIFF
--- a/juniper_rocket/src/lib.rs
+++ b/juniper_rocket/src/lib.rs
@@ -121,7 +121,7 @@ impl<'a> GraphQLBatchResponse<'a> {
 pub struct GraphQLRequest(GraphQLBatchRequest);
 
 /// Simple wrapper around the result of executing a GraphQL query
-pub struct GraphQLResponse(Status, String);
+pub struct GraphQLResponse(pub Status, pub String);
 
 /// Generate an HTML page containing GraphiQL
 pub fn graphiql_source(graphql_endpoint_url: &str) -> content::Html<String> {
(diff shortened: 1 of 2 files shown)


maxdeviant:
@LegNeato Alright, I added an integration test. I based it off the existing tests for `juniper_rocket`, but it builds up a custom Rocket response.

I had to put the test in the `tests` directly so that it properly tests whether the fields of `GraphQLResponse` are visible.

maxdeviant:
@LegNeato No worries! I think "integration test" was a bit of a loaded term :sweat_smile: 

But the test is much simpler now :+1:

LegNeato:
Yeah, sorry, when you put a file in `tests` it is technically an integration test because it can only see `pub` stuff:

https://doc.rust-lang.org/book/second-edition/ch11-03-test-organization.html#integration-tests

Again, I should have been more explicit, thanks for the change and hanging in there!

Issue #242, opened 2018-09-05
Juniper panics when custom `FromInputValue` impl for struct returns `None`


stearnsc:
Hi, we're running into an error and I'm not sure if it's a bug in juniper or if we're doing something wrong.

I have a tagged enum I want to deserialize from a graphql input object, something like:
```
enum Foo {
  A(String),
  B(String),
}
```
which should deserialize from 
```
{
    "variant": "A"
    "value": "..."
}
```
To do so, we implemented `FromInputValue` and `GraphQLType` for `Foo` as so:
```
impl FromInputValue for Foo {
    fn from_input_value(v: &InputValue) -> Option<Self> {
        let obj = v.to_object_value()?;
        match (obj.get("variant"), obj.get("value")) {
            (Some(InputValue::String(variant)), Some(InputValue::String(value))) => match variant.as_str() {
                "A" => Some(Foo::A(value.to_owned())),
                "B" => Some(Foo::B(value.to_owned())),
                _ => None,
            },
            _ => None,
        }
    }
}

impl GraphQLType for Foo {
    type Context = ();
    type TypeInfo = ();

    fn name(_: &()) -> Option<&str> {
        Some("Foo")
    }

    fn meta<'r>(_: &(), registry: &mut ::juniper::Registry<'r>) -> ::juniper::meta::MetaType<'r> {
        let fields = &[
            registry.arg::<String>("variant", &()),
            registry.arg::<String>("value", &()),
        ];
        registry.build_input_object_type::<Foo>(&(), fields).into_meta()
    }
}
```
This appeared to work as expected, and ran in production for some time, until a change in the front-end code exposed a bug where "variant" was sent as "C" (an invalid value). Rather than returning a 400 error with a helpful message as we expected when juniper fails to deserialize a value, juniper instead panicked with the message "Argument missing - validation must have failed". I've confirmed that the error seems to occur any time `from_input_value` returns `None` for a type that has a non-zero number of "fields" as the second argument to `build_input_object_type`.

Is this expected behavior? If so, what's the recommended way to do input validation while avoiding panics?

Thanks!

LegNeato:
Sorry about the crash in production! Are you using an integration crate (like `juniper_iron` or `juniper_rocket`) or did you roll your own?

stearnsc:
No worries; just want to prevent it from happening again :). We're using `juniper_rocket`.

LegNeato:
[...]

PR #244, opened 2018-09-06
Handle InputValues that return `None`

LegNeato:
Fixes https://github.com/graphql-rust/juniper/issues/242.


LegNeato:
While fixing the original issue, this isn't done yet as it breaks the derives.

PR #249, opened 2018-09-13
Use raw url for Juniper logo in README

dvic:
I noticed on https://crates.io/crates/juniper that the logo is not rendering properly? I see the following warning:
`Cross-Origin Read Blocking (CORB) blocked cross-origin response https://github.com/graphql-rust/juniper/blob/master/assets/logo/juniper-dark-word.png with MIME type text/html. See https://www.chromestatus.com/feature/5629709824032768 for more details.`

When I replace the URL with `raw` the logo appears.


theduke:
Thanks!

Issue #235, opened 2018-08-28
Come up with create scalar integration policy / decide what (if anything) to do about smaller integrations


LegNeato:
See https://github.com/graphql-rust/juniper/pull/165#issuecomment-416403522.

There will always be a push and pull of wanting to put integrations in juniper vs keeping them out.

I think this will come up more and more as Juniper gets popular, especially once we go full async and show some benchmarks compared to other languages 😜 . We should probably have some "official" policy of what gets added or not.

Thinking out loud, perhaps we should have an `unstable-integrations` flag that we can put less popular crates behind. We would have no contract for keeping them release-to-release and deleting them would not impact semver but people can opt into using them at their own risk. Users would be acutely aware that if we delete the integration they need to copy the code into their own project and/or create their own integration crate.

sporto:
I would like to see most types supported in Diesel also supported in Juniper. E.g. BigDecimal. As Diesel is a very likely crate that people will use with Juniper, I think maintaining a close parity make sense.

weiznich:
Also see #232.

theduke:
The biggest challenge is the maintenance burden of version upgrades of dependencies.
We already had that situation with `uuid`.

We need a solid strategy for dealing with this, especially when Juniper approaches a 1.0.

With Uuid and chrono we have been able to get a way with having a version range, since the API surface we need is usually very shallow and rarely changes.
If it were to change though, we would need features like `uuid-06`, `chrono-07`, etc. Which are also a bit awkward long term because we can never get rid of them until a a new major version.

I'd propose to remain *very conservative* with adding integrations.
Only relatively stable and very widely used dependencies should be added. 
Everything else will be a big headache.

I currently don't see (m)any crates that would fit this definition.

## Making newtypes easier

Using the newtype pattern and defining wrapper types is admittedly pretty annoying, no doubt about that.
But it's also not that bad.
Especially since, at least in my experience, you end up needing custom types just for the API anyway, since they usually look subtly to not so subtly different from the types used in DB or application logic layers.
I think it's also usually a good pattern to do this, to enforce separation between your db, business logic and API code base.

We can make using wrapper types quite a bit easier though!
We could have a macro like `newtype_scalar!` ( or a custom derive) that makes implementing a wrapper type very effortless, with automatic `From` / `Into` implementations etc.

If we then have a good documentation page in the book for this pattern, it should be enough for most use cases where we feel uncomfortable with adding the dependency.

Issue #232, opened 2018-08-24
Support custom scalar types


weiznich:
I've toyed a bit around to support custom scalar values in juniper. 
The following is some sort of brain dump of the current state.

Adding support for new scalar types that are mapping to existing types seems to be quite easily possible, by reusing existing enum variants in `Value` and `InputValue`. Using this approach  it is for example possible to add support for a 16 bit integer type, because we could simply map that to the existing 32 bit integer. On the other hand adding support for types that are not representable by the existing enum variants is not possible. (An Example here would be adding support for 64 bit integers. They would potentially cause overflows in the 32 bit integer enum variant)

I see several ways to "solve" this: (I've not tried any of this so it may work or it may not work)
* Add more variants to both enums. So that there is a own variant for each supported type. 
This approach has several downsides: 
    * Only a limited set of types could be supported, so if someone want's to introduce a fancy type doing crazy things he will hit the same limit as today
    * We need to teach the parser when to use which integer type (I think this information should already exist somewhere in the schema)

* Use other types in the existing enum variants
Again there are downsides:
    * What's the correct type to uses? For integers: should we use `i64`, but what when someone want's to represent a `i128` and so on 
    * Again, this is not extensible by some crazy new user defined type

* Try to make the schema generic about the exact `Value`/`InputValue` type, provide a default `Value` type (maybe the existing one) and allow users to override this type
Downsides:
     * Again we need to use more knowledge from the schema in the parser
     * I would guess this involves at least rewriting bigger parts of the schema and the parser

I would love to here some other opinions on this before trying further to make this work in a nice way.

(This is basically the only remaining open point for wundergraph)

cc @theduke @LegNeato 

Related: #198

LegNeato:
Thanks for the investigation in this and #198 !

I don't know why, but this still seems "off" to me. It feels like leaking the implementation details / server language semantics into the GraphQL layer. For example, a JS client wouldn't care about i16 vs i32, yet it now needs to care because the server exposes both as scalar types.

Then again, custom scalar types are essentially the server telling the client what the requirements are and the invariants that hold for that field, so maybe it is the right thing to do.

I guess if we define these types in juniper consumers of the lib aren't forced to use them. They could just use the default `Int` and `Float` types and it would be as if they were never added.

So, I think I  adding these via a feature flag that defaults to on--similar to the `uuid`, `url`, etc integrations--would be a good way forward. That way, if juniper users want to stick to only the scalars defined in the spec and defined by their code they can.

[...]

PR #251, opened 2018-09-14
Introduce a abstraction for scalar values

weiznich:
# Disclaimer :wink:
So first I want to say sorry for opening such a big PR, but I see no way to do this in several smaller steps, because this basically needs to change/touch ~everything in juniper.

My main motivation for doing this is to finally solve the discussion about which scalar type should be supported and which not, because wundergraph will need to have some sort of custom scalar values that are not part of the graphql core standard.

Next thing: This is not finished yet, but I decided to open a PR early to start a discussion about the general idea. 
In detail at least the following things needs to be fixed before we could talk about merging this

- [x] Fixing all tests
- [x] Clean up some parts of the code (somewhere are debug `println!` calls)
- [x] Fixing all the other crates in the repo
- [x] Add some documentation
- [x] Add handling of the new generic parameter to some macros (They just set it internally currently, but similar to the `graphql_scalar!` macro there should be a way to specify the type of the used scalar value representation)
- [x] Rebase to master

cc #232 and #235 (Basically this would allow implementing all custom integrations outside of juniper itself, if I do not miss something)

# Description

Before this possible scalar values where hard coded to be representable
in one of the following types: `i32`, `f64`, `String` or `bool`. This
restricts what types of custom scalar values could be defined. (For
example it was not possible to define a scalar value that represents a
`i64` without mapping it to a string (that would be not efficient))

One solution to fix that example above would simply be to change the
internal representation to allow it represent a `i64`, but this would
only fix the problem for one type (till someone want's to support
`i128` for example…). Also this would make juniper not following the
graphql standard closely.

This commit tries to explore another approach, by making the exact "internal"
representation of scalar values swappable (in such a way that a down
stream crate could provide it's own representation tailored to their
needs). This allows juniper to provide a default type that only
contains the types described in the standard whereas other crates
could define custom scalars following their needs.

To that we need to change several things in the current implementation
* [Add some traits that abstract the behaviour of such a scalar value representation](https://github.com/graphql-rust/juniper/pull/251/files#diff-56c8a06849ae961474cb8b01fc58b015R11)
* Change [`Value`](https://github.com/graphql-rust/juniper/pull/251/files#diff-71d8685fbe7c3bd6ebda74205263ecbeR21) and [`InputValue`](https://github.com/graphql-rust/juniper/pull/251/files#diff-5ba3016ef50727e20384c208cb84451cR41) to have a scalar variant (with a
  generic type) instead of hard coded variants for the standard
  types. This implies adding a generic parameter to both enums that
  needs to be added in the whole crate.
* Change the parser to allow deciding between different types of
  scalar values. The problem is basically that the original parser
  implementation had no way to know whether a parsed integer number is
  a `i32` or a `i64` for example. To fix this we added some knowlege
  from the existing schema to the parser.
* Fix some macros and derives to follow the new behaviour

This also contains a unrelated change about the way `juniper_codegen`
resolves items from juniper. In detail the `_internal` flag is removed
the the resolution is replaced by an macro. This is a drive by change
because I was annoyed of adding new items to that list.


juniper/src/executor_tests/custom_scalar.rs
```diff
@@ -0,0 +1,386 @@
...
+            MyScalarValue::Int(i) => Some(i as f64),
+            _ => None,
+        }
+    }
+}
+
+impl From<MyScalarValue> for Option<String> {
+    fn from(s: MyScalarValue) -> Self {
+        match s {
+            MyScalarValue::String(s) => Some(s.clone()),
```
djc: This doesn't have to clone, right?

juniper/src/executor_tests/custom_scalar.rs
```diff
@@ -0,0 +1,386 @@
...
+
+     from_str(value: &str) -> Result<MyScalarValue, ParseError> {
+        value
+            .parse()
+            .map_err(|_| ParseError::UnexpectedToken(Token::Scalar(value)))
+            .map(|s: i64| s.into())
+    }
+});
+
+struct TestType;
```
djc: This should probably go in a nested `#[cfg(test)]` module?

juniper/src/executor_tests/custom_scalar.rs
djc: Never mind, this is only for tests.

juniper/src/lib.rs
```diff
@@ -126,15 +126,15 @@ mod macros;
...
+pub(crate) mod schema;
 mod types;
 mod util;
 mod validation;
 // This needs to be public until docs have support for private modules:
 // https://github.com/rust-lang/cargo/issues/1520
 pub mod http;
 pub mod integrations;
-// TODO: remove this alias export in 0.10. (breaking change)
+// // TODO: remove this alias export in 0.10. (breaking change)
```
LegNeato: Revert.

juniper/src/lib.rs
```diff
@@ -145,7 +145,7 @@ pub mod tests;
 #[cfg(test)]
 mod executor_tests;
 
-// Needs to be public because macros use it.
+// // Needs to be public because macros use it.
```
LegNeato: Revert.

juniper/src/integrations/uuid.rs
```diff
@@ -1,28 +1,36 @@
 use uuid::Uuid;
 
+use parser::ParseError;
+use value::{ParseScalarValue, ScalarValue};
 use Value;
 
-graphql_scalar!(Uuid {
+graphql_scalar!(Uuid where Scalar = <S> {
```
LegNeato: I really don't like this public API. Is it possible to default to `<S>` when there is not a `where Scaler =` in the macro?

juniper/src/macros/object.rs
```diff
@@ -137,11 +137,11 @@ automatically via the `?` operator, or you can construct them yourself using
 struct User { id: String }
 
 graphql_object!(User: () |&self| {
-    field id() -> FieldResult<&String> {
+    field id() -> FieldResult<&String, __S> {
```
LegNeato: I really don't like this public API. Is it possible to default `FieldResult` to the proper type in the simple case and introduce something like  `CustomScalarFieldResult` for the custom case?

juniper/src/executor_tests/enums.rs
```diff
@@ -45,28 +44,35 @@ where
...
+            result.get_field_value("aColor"),
+            Some(&Value::string("RED"))
+        );
     });
 }
 
 #[test]
 fn does_not_accept_string_literals() {
-    let schema = RootNode::new(TestType, EmptyMutation::<()>::new());
+    let schema: RootNode<DefaultScalarValue, _, _> =
```
LegNeato: These annotations now being required are unfortunate :-/. I guess there is no way to alias the default `DefaultScalarValue` case such that they are not necessary?

juniper/src/integrations/uuid.rs
weiznich: It already defaults to `<__S>` but in my opinion (at least in this case) it is more readable if there is no type parameter coming from somewhere.

In cases where someone want's to define a new scalar type based on an existing one, the `where Scalar = …` clause is not necessary, because the `from_str` function could be written in terms of the existing type (for example `<String as ParseScalarValue<_>>::from_str(value)`).

For cases where a custom scalar type is used the scalar type needs to be specified explicitly.

juniper/src/lib.rs
```diff
@@ -180,36 +179,38 @@ pub enum GraphQLError<'a> {
 }
 
 /// Execute a query in a provided schema
-pub fn execute<'a, CtxT, QueryT, MutationT>(
+pub fn execute<'a, S, CtxT, QueryT, MutationT>(
```
weiznich: It should here be easily possible to define a default type for the new generic type.

Something like:
```rust
pub fn execute<'a, S = DefaultScalarValue, CtxT, QueryT, MutationT>(
   document_source: &'a str,
    operation_name: Option<&str>,
    root_node: &'a RootNode<S, QueryT, MutationT>,
    variables: &Variables<S>,
    context: &CtxT,
) -> Result<(Value<S>, Vec<ExecutionError<S>>), GraphQLError<'a>>
where
    S: ScalarValue,
    for<'b> &'b S: ScalarRefValue<'b>,
    QueryT: GraphQLType<S, Context = CtxT>,
    MutationT: GraphQLType<S, Context = CtxT>,
```

I've not done that yet, because it is easier to debug/compile everything without that.

(Also maybe here the generic scalar type should be the last one in the list)

juniper/src/macros/object.rs
weiznich: Similar to the execute function it is possible to define type aliases with default types, so the definition of `FieldResult` would look like this: `type FieldResult<T, S = DefaultScalarValue> = Result<T, FieldError<S>>;`

(Sidenote: In this case this would not have been helpful, because the `graphql_object!` macros introduces internally a generic scalar type (`__S`). This macro needs to be rewritten in a similar manner like the `graphql_scalar!` macro)

weiznich: It should be much more clearer now. 
It is either `FieldResult<&String>` for using the `DefaultScalarValue` or `FieldResult<&String, ScalarType>` for using  a custom type or a generic parameter. In both cases this type is introduced in the header of the macro.

juniper/src/executor_tests/enums.rs
weiznich: It is gone. (At least if you use `DefaultScalarValue`)

juniper/src/integrations/uuid.rs
weiznich: > Is it possible to default to <S> when there is not a where Scaler = in the macro?

In theory this should be possible, but then there would be a generic parameters coming from somewhere. In my opinion that's noting we should do .
Currently the default is set to `DefaultScalarValue`. All other variants need some syntax to specify them. If someone has a better idea than `where Scalar = …`, that should be quite easy to change.)

weiznich:
I've pushed some updates fixing most things that where open in the initial version.

juniper/src/lib.rs
weiznich: So I was wrong. It is not possible to define default types for functions, only for types.

weiznich:
I've pushed another bunch of changes. Now everything should be building, all tests should be passing and there should be no merge conflict.

weiznich:
Remaining CI errors are looking like not to be caused by this change 🙈 

The 1.22/1.23 builder fails because some dependency uses recently stabilised
 features.
The stable/beta builder fails because it tries to compile `juniper_rocket` that requires nightly.

LegNeato:
I'll take a look at this later in the week, but CI should pass here...it's currently passing on master.

weiznich:
> I'll take a look at this later in the week, but CI should pass here...it's currently passing on master.

I totally agree with you that the ci should pass, but I do not see what needs to be fixed (beside of editing the ci config…)
There are the following errors:

* AppVeyor:
    * [nightly gnu target](https://ci.appveyor.com/project/theduke/juniper/build/1.0.321/job/mbuvcfjv4odwadx3#L42) some git failure, looks like that probably goes away by trying to rebuild.
* Travis:
    * [stable](https://travis-ci.org/graphql-rust/juniper/jobs/431639828#L4381) Tries to build juniper-rocket. This will never succeed on stable because of the dependency on rocket
    * [beta](https://travis-ci.org/graphql-rust/juniper/jobs/431639829#L4387) Tries to build juniper-rocket. This will never succeed on beta because of the dependency on rocket
    * [1.23](https://travis-ci.org/graphql-rust/juniper/jobs/431639833#L3341) Fails to build because crossbeam (dependency of wrap) seems to use a feature that requires a newer rust version
    * [1.22](https://travis-ci.org/graphql-rust/juniper/jobs/431639835#L3326) Fails to build because lazycell (dependency of wrap) seems to use a feature that requires a newer rust version

LegNeato:
Right, looks like something got messed up here in a merge? On`master`:

* `juniper_rocket` is [only built on nightly](https://github.com/graphql-rust/juniper/blob/master/juniper_rocket/Makefile.toml).
* `juniper_warp` has a [separate line in the travis config](https://github.com/graphql-rust/juniper/blob/master/.travis.yml#L24) and is only run on certain builders.

It looks like this PR branch doesn't behave similarly.

weiznich:
> Right, looks like something got messed up here in a merge/rebase?

Whatever happened I've edited the ci config in that way that `juniper_rocket` is only build on nightly and `juniper_wrap` is only build on >=stable.
Now the CI passes.

juniper/src/ast.rs
```diff
@@ -496,45 +541,48 @@ impl<'a> VariableDefinitions<'a> {
 mod tests {
     use super::InputValue;
     use parser::Spanning;
+    use value::DefaultScalarValue;
+
+    type TestValue = InputValue<DefaultScalarValue>;
```
LegNeato: Is this needed anymore now that you have a default type?

juniper/src/executor/look_ahead.rs
```diff
@@ -544,7 +569,7 @@ query Hero {
 
     #[test]
     fn check_query_with_argument() {
-        let docs = ::parse_document_source(
+        let docs = parse_document_source(
```
LegNeato: Why does this not need the turbofish and the other calls do?

juniper/src/macros/common.rs
```diff
@@ -0,0 +1,627 @@
...
+}
+
+
+#[doc(hidden)]
+#[macro_export]
+macro_rules! __juniper_parse_object_header {
+    (
+        callback = $callback:ident,
+        rest = <$($lifetime:tt),*> $name: ty $(: $ctxt: ty)* as $outname: tt
+        where Scalar = <$generic:tt $(: $bound:tt)*> $(| &$mainself:ident |)* {
```
LegNeato: I'd still like to make the `where Scalar` clause optional in macros and default insert a `where Scalar = <S>` when omitted. This will ease upgrade pain and means every `graphql_object!` call doesn't need to be updated, only those with custom scalars.

juniper/src/integrations/url.rs
```diff
@@ -1,28 +1,34 @@
...
 #[cfg(test)]
 mod test {
     use url::Url;
+    use value::DefaultScalarValue;
 
     #[test]
     fn url_from_input_value() {
         let raw = "https://example.net/";
-        let input = ::InputValue::String(raw.to_string());
+        let input: ::InputValue<DefaultScalarValue> = ::InputValue::scalar(raw.to_string());
```
LegNeato: Is `InputValue<DefaultScalarValue>` needed here or does type inference handle it?

juniper/src/macros/common.rs
weiznich: There is already [a default case](https://github.com/graphql-rust/juniper/pull/251/files/aa357f7935d0a361f17ded2ae5c69d64839f7c3e#diff-b5a47bf0125442e4083847d3907d57b5R101) without `where Scalar` clause. It defaults to using `DefaultScalarValue` because otherwise `FieldResult` and such things wouldn't work. (If we would default to a generic type `S` each field returning a result could not just return `FieldResult<T>` as in juniper <=0.10, but needs to return `FieldResult<T, S>` with some magic type `S`.)

juniper/src/ast.rs
weiznich: This could be replaced by `InputValue`, but the type annotation will be required in this case. (Because there is no way for rustc to see which type should be used for `S` here)

juniper/src/executor/look_ahead.rs
weiznich: Because this is the only test that explicitly uses a `DefaultScalarValue` in it's expected results.
All other tests don't use scalar values, so we must tell rustc which scalar value type should be used.

juniper/src/integrations/url.rs
weiznich: `: InputValue` is needed, type inference could not handle this otherwise.

juniper/src/macros/common.rs
LegNeato: Oh, I missed that....the macros are huge 😄. Let's change the integrations to omit the clause then (example: https://github.com/graphql-rust/juniper/pull/251/files/aa357f7935d0a361f17ded2ae5c69d64839f7c3e#diff-e1dde797a5a7df30690ff359d95079d0R25)

weiznich:
If we change the integrations to use the default clause only `DefaultScalarValue` will be supported. In my opinion that would be quite surprising. 
The current implementation applies to all scalar value representations out there. So not everybody needs to write the same interactions again and again.


Am 26. September 2018 23:12:12 MESZ schrieb Christian Legnitto <notifications@github.com>:

LegNeato:
Sorry, I'm not being clear. I'll put up a patch showing what I mean.

weiznich:
AppVeyor build failed with an unrelated network error. 
@theduke @mhallin Would be great to get a review on this.

weiznich:
@LegNeato @theduke @mhallin I've rebased everything and added a changelog entry, so this is ready for a hopefully final review.

LegNeato:
It's Monday...today we land! @theduke and @mhallin  can chime in after the merge if they think this isn't the proper direction.

PR #255, opened 2018-09-24
add `#[doc(html_root_url)]` to all published crates

ubnt-intrepid:


theduke:
Mhm, that's the first time I've seen this attribute.

If I understand correctly this causes rustdoc to use the specified url as the base for links when the crate is not also documented locally/in the same run?

https://rust-lang-nursery.github.io/api-guidelines/documentation.html#crate-sets-html_root_url-attribute-c-html-root says that it should contain the exact version though.

There is no redirect from something like `docs.rs/juniper/http` to `docs.rs/juniper/0.10.0/http` so we will need the specific version.

ubnt-intrepid:
> If I understand correctly this causes rustdoc to use the specified url as the base for links when the crate is not also documented locally/in the same run?

Yes. ([example](https://finchers-rs.github.io/finchers-juniper/finchers_juniper/fn.execute.html))

> There is no redirect from something like `docs.rs/juniper/http` to `docs.rs/juniper/0.10.0/http` so we will need the specific version.

It is exactly as you say...
I'll soon fix the specified URL with the accurate crate version.

LegNeato:
Curious, what issue does this solve? I haven't had to use this attribute for other projects.

ubnt-intrepid:
@LegNeato This attribute is intended to help create hyperlinks when the documentations produced by rustdoc refer to items in other crates.
It is not essential to make crate work properly for reasons other than documentation, but it's set on some major crates:

* [in `log`](https://github.com/rust-lang-nursery/log/blob/d711a6bf4891aedbafddad684b6b42ddc5193c51/src/lib.rs#L273)
* [in `hyper`](https://github.com/hyperium/hyper/blob/1370a6f8f06f9906ff75dec904ab9c6d763e37f0/src/lib.rs#L1)

Using this attribute is recommended in the API guideline (see the link mentioned by theduke).
Since it'll increase the process at the time of publishing crates, please reject this PR if you think that it is not necessary.

PR #256, opened 2018-09-30
Use only a single thread pool for juniper_hyper

jonhoo:
The previous implementation used a `futures_cpupool` for executing blocking juniper operations. This pool comes in addition to the thread pool started by hyper through tokio for executing hyper's futures. This patch uses [`tokio::blocking`](https://docs.rs/tokio-threadpool/0.1.7/tokio_threadpool/fn.blocking.html) to perform the blocking juniper operations while re-using the same thread pool as hyper, which simplifies the API.

### DIFF
--- a/juniper_hyper/Cargo.toml
+++ b/juniper_hyper/Cargo.toml
@@ -15,13 +15,13 @@ url = "1.7"
 juniper = { version = ">=0.9, 0.10.0" , default-features = false, path = "../juniper"}
 
 futures = "0.1"
-futures-cpupool = "0.1"
+tokio = "0.1.8"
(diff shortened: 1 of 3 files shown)


LegNeato:
Cool, thanks for the PR! 🍻 

I bet this is why we needed to use `reqwest` to solve the deadlocks in Travis...the futures cpupool and the hyper-handled pool would starve / deadlock each other on machines with few cores.

The `warp` integration likely needs similar changes as it uses  [`futures_cpupool`](https://github.com/graphql-rust/juniper/blob/master/juniper_warp/src/lib.rs#L177) as well. That is out of scope for this PR though.

PR #262, opened 2018-10-07
Update uuid to 0.7

Dylan-DPC:
Updates uuid from 0.4 - 0.6 to 0.7. None of the breaking changes in 0.7 affect the codebase.

### DIFF
--- a/juniper/Cargo.toml
+++ b/juniper/Cargo.toml
@@ -42,7 +42,7 @@ serde_derive = { version = "1.0.2" }
 chrono = { version = "0.4.0", optional = true }
 serde_json = { version="1.0.2", optional = true }
 url = { version = "1.5.1", optional = true }
-uuid = { version = "> 0.4, < 0.7", optional = true }
+uuid = { version = "0.7", optional = true }
 
 [dev-dependencies]
 bencher = "0.1.2"


Dylan-DPC:
That might not be a good idea. Since there are many API changes between 0.4-0.6 and 0.7. You might end up with some weird situations if you call a 0.7 function and user has 0.4 to 0.6 dependency installed. 

Also, the way cargo handles dependencies, it will most likely install the highest version of a dependency (0.7.*), so majority of the users will get a 0.7 version

LegNeato:
But from Juniper's point of view the API we use has been stable. If we just do 0.7 and a user has 0.6, they get two copies. If we do a range and they have 0.6, they only use one copy of 0.6 right? Or am I misunderstanding cargo's dep resolution?

Dylan-DPC:
So yeah, i did a demo of it. I created a dummy project dependent on uuid 0.5 and juniper 0.10.0 (uuid 0.4 to 0.6). For juniper, cargo picks 0.6 despite the project having a direct dependency on 0.5 and thus you get 2 copies instead of one. So my assumption was correct.

LegNeato:
Ah, cool, thanks for testing!

Issue #253, opened 2018-09-19
[FR] make opration_name public


trsh:
```
fn operation_name(&self) -> Option<&str> {
        self.operation_name.as_ref().map(|oper_name| &**oper_name)
    }
```

Sometimes u want todo extra stuff before execute or after. Like checking some access orsmth.

LegNeato:
Sure, would be happy to take a PR.

PR #265, opened 2018-10-16
Make operation_name public

cgm616:
See #253.


cgm616:
Where can I add an integration test? Is there a collection of them already that exists, or should I create a `juniper/tests` directory alongside `src` in that crate?

LegNeato:
I think the `tests` directory is the way to go. You'll need to use [`GraphQLRequest::new()`](https://github.com/graphql-rust/juniper/blob/fd636e07df99744d9964570991bba3ff283f6642/juniper/src/http/mod.rs#L46) and then call `operation_name()`. If it is private (that is, we undo this change), the integration test should then fail to compile.

Let me know if you run into any issues!

LegNeato:
@cgm616 Thank you very much! As soon as CI goes green I will merge.

LegNeato:
🎉

PR #195, opened 2018-06-08
Add support for using doc comments as descriptions

LegNeato:
Fixes https://github.com/graphql-rust/juniper/issues/194.

Issue #194, opened 2018-06-07
Doc comments and custom derives


weiznich:
Custom derives should parse doc comments and use them as description. This is currently not done at least for `GraphQLEnum` and `GraphQLInputObject`. 

(Maybe the same would be possible for `deprecated` annotations?)

PR #269, opened 2018-10-23
Infer graphql "deprecation" from #[deprecated(note = "...")] in derive (and macros)

kestred:
Follow on to https://github.com/graphql-rust/juniper/issues/194.

### Derive
Updates the `GraphQLObject` and `GraphQLEnum` derives to use the first `#[deprecated]` annotation with a `note = ...` to supply the deprecation text, which can still be overridden with `#[graphql(deprecation = ...)]` like in https://github.com/graphql-rust/juniper/pull/195.

### Macros
Additionally, this PR adds support for `#[doc = ...]` and `#[deprecated(note = ...)]` to the `graphql_object!` and `graphql_interface!` macros, which has some benefits:

  * Lets your `field name(args...)` line-up so that they're easier to scan (even with a deprecation)`
  * Allows mutliple doc strings be combined akin to the [collapse_docs](https://doc.rust-lang.org/rustdoc/the-doc-attribute.html) behavior.
    (including any `const SOME_DESC: &'static str` if you want w/ `#[doc = SOME_DESC]`) 
  * Makes it easier to refactor a `#[derive(GraphQLObject)]` into a `graphql_object!`, e.g. a contrived example

    ```rust
    #[derive(GraphQLObject)]
    struct Plus {
         left: i32,
         right: i32,

         #[deprecated(note = "This has been deprecated because it doesn't handle overflow")]
         result: i32,
     }

    // after converting to use macro
    graphql_object!(Plus: () |&self| {
         field left() -> i32 { self.left }
         field right() -> i32 { self.right }

         #[deprecated(note = "This has been deprecated because it doesn't handle overflow")]
         field result() -> i32 { self.result }
         field checked_sum() -> Option<i32> { self.left.checked_sum(self.right) }
    })
    ```

### DIFF
--- a/juniper_codegen/src/derive_enum.rs
+++ b/juniper_codegen/src/derive_enum.rs
@@ -67,6 +67,9 @@ impl EnumVariantAttrs {
         // Check doc comments for description.
         res.description = get_doc_comment(&variant.attrs);
 
+        // Check rust "deprecated" for deprecation.
+        res.deprecation = get_deprecated_note(&variant.attrs);
+
         // Check attributes for name and description.
         if let Some(items) = get_graphql_attr(&variant.attrs) {
             for item in items {
--- a/juniper_codegen/src/derive_enum.rs
+++ b/juniper_codegen/src/derive_enum.rs
@@ -44,7 +44,7 @@ impl EnumAttrs {
                     continue;
                 }
                 panic!(format!(
-                    "Unknown attribute for #[derive(GraphQLEnum)]: {:?}",
+                    "Unknown enum attribute for #[derive(GraphQLEnum)]: {:?}",
                     item
                 ));
             }
@@ -57,7 +57,7 @@ impl EnumAttrs {
 struct EnumVariantAttrs {
     name: Option<String>,
     description: Option<String>,
-    deprecation: Option<String>,
+    deprecation: Option<DeprecationAttr>,
 }
 
 impl EnumVariantAttrs {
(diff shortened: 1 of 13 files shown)


kestred:
Oh; I didn't realize the reason was optional; I'll definitely update to support the bare deprecation

kestred:
@LegNeato: Changelog, docs, etc added--- build failure is only due to __spurious network error__.

juniper_codegen/src/derive_enum.rs
```diff
@@ -90,13 +93,24 @@ impl EnumVariantAttrs {
                     continue;
                 }
                 if let Some(AttributeValue::String(val)) =
-                    keyed_item_value(&item, "deprecated", AttributeValidation::String)
+                    keyed_item_value(&item, "deprecation", AttributeValidation::String)
```
LegNeato: This should be `deprecated`, right? And if so, it looks like repeated code and can be removed, as the next block is doing `match keyed_item_value(&item, "deprecated", AttributeValidation::String`

kestred: Partially yes.

This check appears to be unnecessary for backwards compatibility, however I believe the reason that the line is in the `EnumVariantAttrs` impl is because I was using the same code block to handle it for both that and `ObjectFieldAttrs`.

In the __current version__ (specifically on master), the attribute name is inconsistent between enum variants which use `deprecated` and object fields which use `deprecation`.

This change, makes both support the same set of attributes in a backwards-compatible way.
Happy to remove it if that is undesirable.

If the use of `deprecation` on object fields is itself considered deprecated, I can remove this one on EnumAttrs, and if desired can also remove the same one on ObjFieldAttrs.

LegNeato: Ah, yeah. Sorry, forgot about that previous support.

PR #271, opened 2018-10-28
Upgrade `juniper_warp` to `warp` 0.1.8

LegNeato:
This unbreaks the build, as the newer warp deprecates some of the filters we
were using and we had `deny_warnings` on.


LegNeato:
Build failures are https://github.com/rust-lang/rust/issues/55376

PR #273, opened 2018-11-01
Just use doc comments instead of the doc attribute in the changelog e…

weiznich:
This just looks more like rust code in this way.

(And yes this is just working)

### DIFF
--- a/changelog/master.md
+++ b/changelog/master.md
@@ -57,13 +57,13 @@
     }
 
     // New alternative syntax for field descriptions
-    #[doc = "Description"]
+    /// Description
     field my_field() -> { ... }
 
     // New alternative syntax for argument descriptions
     field my_field(
-      #[doc = "The number of starfish to be returned. \
-               Can't be more than 100."]
+      /// The number of starfish to be returned.
+      /// Can't be more than 100.
       arg: i32,
     ) -> {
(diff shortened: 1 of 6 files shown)


theduke:
The tests are there for a reason... ;)

Also, the object macro docs should still show the `#[doc]` variant.

There are valid situations where you'd want to use that instead.

weiznich:
I've fixed the tests. (There were just some whitespace issues 😉 ). I've also included a hint to the docs that doc attributes are also supported.

The CI failures seems to be caused by a dependency that does not support the minimal supported rustc version.

kestred:
Oh this actually works?!  😅

weiznich:
@kestred Doc comments are basically identical to using the doc attribute in this case. See [the documentation](https://doc.rust-lang.org/beta/rustdoc/the-doc-attribute.html) for more information.

weiznich:
@theduke @LegNeato any news on this?

LegNeato:
I'm trying to get CI back up using precompiled binaries from https://github.com/sagiegurari/cargo-make/pull/127.

LegNeato:
Actually, I see all the tests pass on other builders so I'll just land while I wrestle with CI.

PR #275, opened 2018-11-07
Pin cargo make version

LegNeato:
0.14 of cargo make does not support our min rust version

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ env:
     - CARGO_MAKE_RUN_CODECOV=true
 
 script:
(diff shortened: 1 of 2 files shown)


sagiegurari:
This is good as in 0.15.0 min rust version again will have to go up :(
but it will come with some nice features such as rustup integration

LegNeato:
@sagiegurari One of these days I am going to send you a patch to get binaries automatically uploaded to bintray for all platforms we care about 😄

sagiegurari:
@LegNeato please do!!!

Issue #218, opened 2018-08-10
Officially support wasm32-unknown-unknown


LegNeato:
We appear to work with `wasm32-unknown-unknown`. It would be nice to build with it in CI, as well as have an example where it is used in `node`. I'd be curious as to the perf, as rust/wasm is generally faster than js but most js graphql server implementations are async.

theduke:
The really interesting use case here for me would be `offline` support.

As in, writing your GraphQL backend only once, with some for of compatability layer for database access. 
Compiling the backend down to wasm and then using it in the browser in offline mode would be a pretty rad feature.

LegNeato:
Oh yeah, that would be cool. If the DB was https://realm.io/products/realm-database/ your changes would get transparently synced when you went back online too.

theduke:
@LegNeato guess this can be closed?

LegNeato:
Yeah, I think we are ok for now. I still want to have a node example but that can come later.

PR #278, opened 2018-11-16
Build wasm32-unknown-unknown in CI

willglynn:
`cargo check --wasm32-unknown-unknown` is sufficient to demonstrate that the `juniper` and `juniper_codegen` crates can compile for WebAssembly, which should help avoid accidental regressions.

This is a small step towards the goals in #218.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,14 @@ rust:
   - 1.23.0
   - 1.22.0
 
+matrix:
+  include:
+    - language: rust
+      rust: stable
+      name: "check wasm32 support"
+      install: rustup target add wasm32-unknown-unknown
+      script: cargo check --target wasm32-unknown-unknown --package juniper --package juniper_codegen
+
 env:
   global:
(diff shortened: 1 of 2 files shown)


willglynn:
Added!

I'm new to Azure Pipelines and I wasn't quite sure what approach to take here, so I settled for the one that seemed least invasive. The result was a "Check WebAssembly target" step added to every job, but which runs only for the stable toolchain and is skipped everywhere else.

I also considered expanding the matrix such that "install stable + `cargo check --target=wasm32-unknown-unknown`" would be a separate job on all three platforms. However – as I understand it at least – this would require either a) making the regular build process conditional on whether we're building normally or checking WebAssembly, or b) adding a second unrelated job template and expanding it separately from the `azure-pipelines.yml` in the root. Adding a two-line step to the existing jobs and conditionally executing it seemed like the better option.

LegNeato:
Sorry this took so long, thanks for the PR! 🍻

PR #281, opened 2018-11-23
Use pre-compiled cargo-make

LegNeato:


sagiegurari:
I'll be releasing the next patch with windows binary hopefully next week

Issue #282, opened 2018-11-28
Can't use ID tuple struct inside graphql_object macro


BrendanBall:
Hi, I'm new to rust and trying to implement a basic graphql server. According to the rust book you use tuple structs as follows:
```
struct Color(i32, i32, i32);
let black = Color(0, 0, 0);
```
And the `juniper::ID` type is defined as `pub struct ID(String);`. However when trying to instantiate an ID inside the graphql_object macro it results in the error 
```
error[E0423]: expected function, found struct `ID`
  --> src/schema.rs:23:21
   |
23 |                 id: ID("1".to_owned()),
   |                     ^^ did you mean `ID { /* fields */ }`?
```
e.g modified code from https://github.com/actix/examples/blob/master/juniper/src/schema.rs
```
graphql_object!(MutationRoot: () |&self| {
    field createHuman(&executor, new_human: NewHuman) -> FieldResult<Human> {
        Ok(Human{
            id: ID("1234".to_owned()),
            name: new_human.name,
            appears_in: new_human.appears_in,
            home_planet: new_human.home_planet,
        })
    }
});
```
I haven't found any existing issues related to this. How am I meant to use the ID type inside of the macro?

theduke:
Hey, the inner String of the `ID` struct is private, as you can see here, so you can't directly create it. (see https://docs.rs/juniper/0.10.0/src/juniper/types/scalars.rs.html#17).

ID has a `From` implementation for String though, so you can construct it like this:

```rust
  id: "1234".to_string().into(),
  // OR:
  id: ID::from("1234".to_string()),
```

theduke:
If you want to contribute, @BrendanBall , you could do a PR with the `new` method added.

BrendanBall:
Thanks, it seemed as though I could instantiate `ID` normally outside of the macro which is what confused me. I can add the new method.

theduke:
Cheers.

And it doesn't have anything to do with the macro, it wouldn't work outside of it either. ;)

PR #283, opened 2018-11-29
Add new method to ID  scalar

BrendanBall:
As suggested in https://github.com/graphql-rust/juniper/issues/282#issuecomment-442854714


BrendanBall:
The build failed for rust 1.22 and 1.23, is this expected? It doesn't seem to be anything to do with the code I changed.

LegNeato:
Yep, that's expected. I'm going to take a crack at fixing it this week. Thanks for the PR, looks great!

PR #284, opened 2018-11-29
Edit: WIP - run CI without cargo-make

theduke:
While cargo-make is a cool project for more complicated build setups, 
 at this point I don't really think it provides much of a benefit for us.

Also it requires 1.30 now so we'd need to install stable in addition to the tested version in all tests.

All we really need to do is (in the workspace root):

```bash
cargo check --verbose --all-features
cargo test --verbose --all-features
+ code coverage
```

The only moderately complicated part is code coverage so I'd vote for just dropping cargo-make.

(haven't pushed yet)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,8 @@ rust:
   - nightly
 
   # Prevent accidentally breaking older Rust versions
-  - 1.23.0
-  - 1.22.0
+  - 1.28.0
+  - 1.27.0
 
 env:
   global:
--- a/_build/azure-pipelines-template.yml
+++ b/_build/azure-pipelines-template.yml
@@ -11,9 +11,9 @@ jobs:
       nightly:
         rustup_toolchain: nightly
       minimum_supported_version_plus_one:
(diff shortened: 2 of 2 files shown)


theduke:
Necessary now because some deps are failing on 1.23

LegNeato:
Hmmm, I haven't looked into this too much but don't different crates in the workspace have different requirements? FWIW `cargo-make` only requires a newer rust to build, not run of course...which is why I added support for them building artifacts and started on https://github.com/graphql-rust/juniper/pull/281

theduke:
Nah we don't do anything custom with it, just the default workflows which just run build and test.

LegNeato:
`juniper_rocket` uses it to run on nightly and skip on non-nightly CI jobs:

https://github.com/graphql-rust/juniper/blob/master/juniper_rocket/Makefile.toml

Also, we use the stock coveralls upload support. 🤷‍♂️

theduke:
Closing this for now.

PR #291, opened 2018-12-11
Fix example link in juniper_rocket README

cschneid:
It appears that the official repo was combined back into the main juniper repo, but the link didn't get updated.


LegNeato:
Awesome, thanks!

PR #298, opened 2018-12-18
Fix helper macros for 2018 edition.

theduke:
Fixes #298


LegNeato:
Should we have an integration test that uses the 2018 edition to make sure we don't regress again here?

theduke:
@LegNeato exactly what I thought.

theduke:
I refactored the macros now to use the $crate:: method instead of local_inner_macro.

LegNeato:
Do we want to release a patch to 0.11? If so we should probably not use $crate as to not bump the min version. I'm not a stickler for semver so would also be willing to just push a point release and be done with it so we don't have to republish all the other crates. 🤷‍♂️

theduke:
Yeah I'd say we just retroactively say that 0.11 requires 1.30+.
We officially only support the current release - 2 anyway, so that works out fine. 

I'd just really like to find a fix for the custom derive issue that's not so awkward.
I found one but that only works on 1.32+. :/

theduke:
Alright, it's done!

A explanation of the fix is in the commit message: https://github.com/graphql-rust/juniper/pull/298/commits/7b2b6e9f3df0c9978c970565dec8b8938b6e165a

theduke:
Added a VERY minimal 2018 edition test.

Issue #299, opened 2018-12-19
Rename ScalarValue custom derive to GraphQLScalarValue


theduke:
Don't know when/how this slipped through but the custom derive should not just be named `ScalarValue`.

It breaks the pattern from the other derives.

LegNeato:
Sorry, I missed the naming in code review.

PR #301, opened 2018-12-19
Rename ScalarValue derive

LegNeato:
Fixes https://github.com/graphql-rust/juniper/issues/299.


theduke:
:+1: 

As a related thought: I/we've been pretty lazy about ensuring continuity for juniper users with the breaking changes and no upgrade paths etc.

Maybe we should resolve to hold off on 0.12 for quite a while and batch up breaking changes.
Until then we can release patch versions that prepare for the breaking changes.

So here we could leave a `#[deprecated]` `ScalarValue` derive in addition to the new `GraphQLScalarValue` and only remove it directly before we publish 0.12.

PR #312, opened 2019-01-09
Fix TODO for return type

LegNeato:
We have bumped past the point of this TODO so we can make
the changes.


theduke:
Hopefully nightly is fixed soon.

PS: this is technically a breaking change so needs to wait for 0.12.

PR #317, opened 2019-01-25
add GraphQL Playground integration

### LINKED ISSUE #274: Add playground integration?
I saw in #230 playground integration was removed along the way. Is that just because it would need to be refactored into a central `playground_source` function similar to [`graphiql_source`](https://github.com/graphql-rust/juniper/blob/4be687f73eeb8c80044f542dfea49910e8347a71/juniper/src/http/graphiql.rs#L4-L66) or because of some other reason?

Playground has a lot more features than graphiql, like the ability to set/copy http headers (which is necessary for many APIs), so it would be nice to have... 

Although it looks like juniper's graphiql implementation allows sending the auth header? It still would be nice to have playground support out of the box.

jgillich:
This adds [GraphQL Playground](https://github.com/prisma/graphql-playground) integration, a much nicer and more powerful alternative to graphiql.

resolves #274

Notes:
* My nightly compiler panics, so I couldn't run (all) tests
* I did not add a top level alias because the one for graphiql [will be removed](https://github.com/graphql-rust/juniper/blob/29389e8f7231265dd4af398f17ad440cbefade55/juniper/src/lib.rs#L139)
* Contains no documentation or exaples, should I add some?
* I'd even propose to make this the default in the examples, since Playground is much nicer.

### DIFF
--- a/juniper_iron/src/lib.rs
+++ b/juniper_iron/src/lib.rs
@@ -221,6 +221,11 @@ pub struct GraphiQLHandler {
     graphql_url: String,
 }
 
+/// Handler that renders `GraphQL Playground` - a graphical query editor interface
+pub struct PlaygroundHandler {
+    graphql_url: String,
+}
+
 fn get_single_value<T>(mut values: Vec<T>) -> IronResult<T> {
     if values.len() == 1 {
         Ok(values.remove(0))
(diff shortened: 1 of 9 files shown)


juniper_iron/src/lib.rs
```diff
@@ -327,6 +332,18 @@ impl GraphiQLHandler {
     }
 }
 
+impl PlaygroundHandler {
+    /// Build a new GraphiQL handler targeting the specified URL.
```
LegNeato: Replace `GraphiQL` with `GraphQL Playground`.

jgillich:
All done.

jgillich:
Oh wait a second

jgillich:
I accidentally added it to the wrong changelog (please squish commits on merge). I guess this doesn't need a mention in each of the web framework crates?

LegNeato:
Nope, that's great...thanks!

Issue #307, opened 2018-12-20
Implement introspection schema.json generation


theduke:
Implement a method to easily create a `schema.json` introspection schema used by tools like `apollo` cli and https://github.com/graphql-rust/graphql-client.

For simplicity I'd just execute the introspection query against the schema and return the result.

theduke:
Example query:

```graphql
query IntrospectionQuery {

  __schema {

    queryType {

      name

    }

    mutationType {

      name
    

    }

    subscriptionType {

      name

    }

    types {

      ...FullType

    }

    directives {

      name

      description

      locations

      args {

        ...InputValue

      }

    }

  }

}

fragment FullType on __Type {

  kind

  name

  description

  fields(includeDeprecated: true) {

    name

    description

    args {

      ...InputValue

    }

    type {

      ...TypeRef

    }

    isDeprecated

    deprecationReason

  }

  inputFields {

    ...InputValue

  }

  interfaces {

    ...TypeRef

  }

  enumValues(includeDeprecated: true) {

    name

    description

    isDeprecated

    deprecationReason

  }

  possibleTypes {

    ...TypeRef

  }

}

fragment InputValue on __InputValue {

  name

  description

  type {

    ...TypeRef

  }

  defaultValue

}

fragment TypeRef on __Type {

  kind

  name

  ofType {

    kind

    name

    ofType {

      kind

      name

      ofType {

        kind

        name

        ofType {

          kind

          name

          ofType {

            kind

            name

            ofType {

              kind

              name

              ofType {

                kind

                name

              }

            }

          }

        }

      }

    }

  }

}
```

mrceperka:
Juniper is able to process this query, but `apollo` fails to generate `schema.json`.
I've received this error:
```
 Expected a value of type "__DirectiveLocation" but received: "INLINE_SPREAD"
```
I was looking at https://github.com/facebook/graphql/blob/master/spec/Section%204%20--%20Introspection.md and there `FRAGMENT_SPREAD`. Maybe there is a typo in juniper...

LegNeato:
I have implemented this. Just cleaning up the code and adding tests before putting up a PR.

LegNeato:
Hmmm, actually my PR may not fix this issue...it doesn't output in JSON. Which is different.

Issue #138, opened 2018-02-05
Use graphql-parser for parsing query language


tailhook:
Hi,

Since there is no activity in #32 I've decided to write a separate crate for [graphql-parser](https://crates.io/crates/graphql-parser).

It's not based on code here but built using excellent `combine` crate (making the grammar more readable in my opinion). So in a weekend project I could build the parser for the whole graphql grammar. Which means it should be on par with this crate for features (I've not checked every feature, but we also have subscriptions and triple-quoted strings already implemented).

What do you think about integrating with the `graphql-parser` crate instead of keeping custom parser?

Keats:
That looks much better than my half-baked Pest parser (although I still prefer pest to combine but that's not important)!

It would be really neat for the Rust ecosystem to standardise on a single parser so at least every crate wanting to do something with GraphQL doesn't have to rewrite it and it has many eyes on it.

theduke:
@tailhook I'll look over the code, but in principle, as long as the AST produced is decent and can easily be converted (without much cloning) I'd definitely be in favour.

This also concers  https://github.com/nrc/graphql and https://github.com/nrc.
We've already talked about sharing fundamental infrastructure like the parser.

We'd also need the parser to handle parsing the GraphQL schema definition language.

theduke:
@tailhook also, it's important to have great error reporting for parsing errors and to be able to map each ast node to the query location.

tailhook:
> This also concers https://github.com/nrc/graphql and @nrc.
> We've already talked about sharing fundamental infrastructure like the parser.

Sure, here is the issue: https://github.com/nrc/graphql/issues/36

> it's important to have great error reporting for parsing errors

I think it's good enough. Of course, we have positions of all the errors. And "unexpected A, expected B, C". I'm not sure what else to say. Perhaps, there aren't much test for error reporting yet.

Currently, we don't have decent API for the error besides: format me a string, but it should be easy to add just need to figure out how it might look like.

> and to be able to map each ast node to the query location.

We have positions for all important elements. The approach is different that one here, but we can discuss details.

> We'd also need the parser to handle parsing the GraphQL schema definition language.

Surely, on a todo list. Does seem like another weekend project in it's size. The good thing that tokenizer may be reused entirely.

theduke:
@tailhook all sounds great, I'll look over the code!

Would you be interested in moving the parser into the `graphql-rust` organization?

[...]

PR #324, opened 2019-03-02
Add support for GraphQL Schema Language

LegNeato:


LegNeato:
A couple of notes:

1. Naming is probably bad, please give suggestions if so.
2.  I didn't document the `graphql-parser-integration` feature. Not sure if we want that as public API, though I currently made it public in case other projects need to use it.
3. I had a trait `AsSchemaLanguage` and implemented it for every `MetaType`. I didn't like how it made the code look with all the conditional compilation, so I centralized it into an `AstTranslator` instead.

theduke:
I'm not so sure what the use case for this is.

The schema language is usually used the other way around AFAIK, as in bootstrapping a server from a schema file and adding resolver methods.

Downstream tools usually use a introspection query to get the schema.

Where do you see this being useful?

LegNeato:
You need a schema file to use https://github.com/graphql-rust/graphql-client among other use cases such as external tools. If the schema is defined in Juniper/rust, it is essentially "locked up" until runtime.

LegNeato:
But now that I think about it, perhaps it makes sense for this to be its own little crate/project? An argument can be made that it should be part of juniper as well as not. I guess the same arguments would apply to https://github.com/graphql-rust/juniper/issues/307 as well?

theduke:
First glance looks good, I'll review thoroughly later  today.

One question is how much work it would be to fully switch over to  `graphql-parser` instead of bringing it in as a feature.

Also maybe there is a more elegant  way to detect builting types  than `__` prefix checking.

LegNeato:
@theduke For switching entirely over to `graphql-parser`'s AST I think the concerns in https://github.com/graphql-rust/juniper/issues/138#issuecomment-368644460 are still valid?

theduke:
@LegNeato as a first step I have opened a PR that allows a borrowed AST on the parser: https://github.com/graphql-rust/graphql-parser/pull/19

I'll investigate wheter we'd really want to switch to that AST, and compare it with our own.
But, mapping the AST from the parser to ours would be a very lightweight operation that shouldn't make a difference performance wise.

theduke:
@LegNeato I'm working on switching to graphql-parser.

We could wait with this until that happens, or move ahead now if you'd like to get it in sooner.

juniper/src/schema/meta.rs
```diff
@@ -165,6 +165,14 @@ pub struct Field<'a, S> {
     pub deprecation_status: DeprecationStatus,
 }
 
+impl<'a, S> Field<'a, S> {
+    /// Returns true if the type is built-in to GraphQL.
+    pub fn is_builtin(&self) -> bool {
+        // "used exclusively by GraphQL’s introspection system"
```
theduke: This seems a bit hacky, `__` is not reserved so users could add their own objects/ fields/arguments/...

We could add a `builtin: bool` flag to the relevant schema types.

theduke:
Mostly looks good. 

Actually I'd be in favor of including it now, but without the feature, let's just build it in.

LegNeato:
Ok, I'll rip out the features, rebase, and switch to borrowing after your patch lands.

arlyon:
Juniper has changed quite a bit since August (especially with the recommendation to base new PRs on the async branch). That said, keen to test this PR out, I rebased this onto #493 and updated graphql-parser to depend on master and after sprinkling in some lifetimes this works like a charm.

Seems like for this to be merged we are waiting on:

- the `async-await` branch merge
- a new version of `graphql-parser`

I can open a PR with my (minimal) changes, just know that I am still quite new to rust.

LegNeato:
I updated master to use the async-await branch.

LegNeato:
Would love a PR with the updated changes!

juniper/src/schema/meta.rs
LegNeato: From https://github.com/graphql/graphql-spec/blob/master/spec/Section%203%20--%20Type%20System.md#schema:

"All types and directives defined within a schema must not have a name which begins with {"__"} (two underscores), as this is used exclusively by GraphQL's introspection system."

LegNeato:
@arlyon any chance you can put up the rebased changes?

LegNeato:
Friendly 👉@arlyon

arlyon:
Hey, missed these for some reason. I was going to open a PR, but I noticed a new version of graphql-parser is out and it seems silly not to update. Lifetimes are a bit annoying to iron out so it's taking a bit of time.

LegNeato:
@arlyon great, thank you! Yes, the new version should allow us to use &str instead of Strings and be a lot more memory efficient feel free to post WIPs if you get stuck!

LegNeato:
This was landed as https://github.com/graphql-rust/juniper/pull/676

PR #331, opened 2019-03-06
Disable juniper_hyper testing due to rustc bug

theduke:
This will make the tests pass for now when dealing with pull requests.

Once merged we should create an issue to track the re-enabling of all channels (as an addition to the `TODO` comment).


theduke:
ps: the changes have reached beta, so it's not working there anymore either

theduke:
Ok well, it's not even working on stable anymore!

PR #337, opened 2019-04-04
Add support for the canonical introspection query

LegNeato:
Fixes https://github.com/graphql-rust/juniper/issues/307.

I also fixed all the book tests to run and I am too lazy to put them in a different PR 🤷‍♂️

### DIFF
--- a/docs/book/content/SUMMARY.md
+++ b/docs/book/content/SUMMARY.md
@@ -2,20 +2,22 @@
 - [Quickstart](quickstart.md)
 
 - [Type System](types/index.md)
-    - [Defining objects](types/objects/defining_objects.md)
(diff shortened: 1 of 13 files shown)


theduke:
:+1: 

LGTM

theduke:
@LegNeato actually, why do we need the with description / no description split?

LegNeato:
@theduke I don't know, I just followed what graphqljs did. I would imagine it is probably for space savings?

LegNeato:
@theduke confirmed with the GraphQL.js team that it to save data/space when the result is read by a machine. If the target is a dev tool that wants descriptions then you include them, but if the target is a code generator that doesn't need them then you can save the download bytes.

PR #342, opened 2019-04-09
Fix broken docs links

murielsilveira:
This PR fixes many links on the documentation that were broken because they were not using the correct relative paths.


theduke:
Whoops, thanks for catching that!

theduke:
It's very weird that `mdbook` doesn't complain about non-existant links...

PR #344, opened 2019-04-17
Include README content in book

LegNeato:
This is brittle (uses line numbers) but I think better than
copying and pasting.

### DIFF
--- a/docs/book/content/README.md
+++ b/docs/book/content/README.md
@@ -1,74 +1,19 @@
-# Juniper
+{{#include ../../../README.md:0:3}}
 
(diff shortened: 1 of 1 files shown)


theduke:
Phew, we would need to remember this whenever the readme changes.

Could we have a simple build script that excludes stuff without relying on line numbers?

Also imo the badges wouldn't really hurt in the book, any specific reason why you removed them?

andy128k:
Could [cargo-sync-readme](https://github.com/phaazon/cargo-sync-readme) be used instead?

PR #348, opened 2019-05-02
Fix lookahead for optional variables

### LINKED ISSUE #347: Panic when using look ahead together nullable variables
**Describe the bug**
If you have a query with a nullable variable, and the resolver calls `executor.look_ahead()` then you get a panic if you don't set the variable at all. However if you set the variable to `null` it doesn't panic 🤔 

**To Reproduce**
Steps to reproduce the behavior: Download https://github.com/davidpdrsn/juniper-variable-bug and run `cargo test`.

**Expected behavior**
I imagine these two scenarios should behave identically. I would expect neither of them to panic or error.

**Additional context**
I discovered this because the code generated by [juniper-from-schema](https://github.com/davidpdrsn/juniper-from-schema) calls `.look_ahead()` for all fields.

theduke:
Closes #347


theduke:
Checks failing due to nightly ordering issue, will be fixed in a separate pr.

PR #349, opened 2019-05-02
Make introspection tests ordering independent.

theduke:
With the new HashMap implementation, the maps on stable and nightly
have different ordering.

The introspection tests were relying on strict ordering.
This commit adds a simple sorting of the results to prevent
test failures.

A related discussion should be if we want to switch to index map for the types and directives, or just manually sort them for introspection queries, but this is the quick fix for now.


theduke:
@LegNeato merging myself because of the small change.

PR #352, opened 2019-05-06
Specify path to graphql_object module

rogefm:
This example doesn't work because we are not specifying the relative path for the graphql_object module.


LegNeato:
Thanks for the patch!

PR #354, opened 2019-05-16
Book: add ::object field/arg customization docs

theduke:


theduke:
Merging, note the build failure is a spurrious error from `Cirrus CI` which we don't actually use.

PR #357, opened 2019-05-16
(ci) Re-enable juniper_hyper testing on nightly

theduke:


theduke:
Now, let's fast forward into the future where stable and nightly are updated. ^^

PR #367, opened 2019-05-25
fix(docs): correct warp example link in book

iamclaytonray:
This fix adds the correct link for the Warp Server example.


LegNeato:
Thank you!

caio:
Hi! Just thought I'd say that it doesn't look like this patch is live at https://graphql-rust.github.io/juniper/current/servers/warp.html

PR #368, opened 2019-05-26
Implement newtype derive for scalars.

### LINKED ISSUE #303: ScalarValue custom derive: newtype support

theduke:
This commit implements a newtype style custom derive
for scalars via `#[derive(GraphQLScalarValue)]`, which now
supports both deriving a base enum scalar type and newtypes.

For newtypes, the `#[graphql(transparent)]` attribute is
required.

This commit:
* implements the derive
* adds integration tests
* updates the book

Closes #303

### DIFF
--- a/docs/book/content/types/scalars.md
+++ b/docs/book/content/types/scalars.md
@@ -6,43 +6,14 @@ but this often requires coordination with the client library intended to consume
 the API you're building.
 
 Since any value going over the wire is eventually transformed into JSON, you're
-also limited in the data types you can use. Typically, you represent your custom
-scalars as strings.
+also limited in the data types you can use. 
 
-In Juniper, you use the `graphql_scalar!` macro to create a custom scalar. In
-this example, we're representing a user ID as a string wrapped in a custom type:
+There are two ways to define custom scalars. 
+* For simple scalars that just wrap a primitive type, you can use the newtype pattern with
+a custom derive. 
+* For more advanced use cases with custom validation, you can use
+the `graphql_scalar!` macro.
 
-```rust
-use juniper::Value;
-
-struct UserID(String);
-
-juniper::graphql_scalar!(UserID where Scalar = <S> {
-    description: "An opaque identifier, represented as a string"
-
-    resolve(&self) -> Value {
-        Value::scalar(self.0.clone())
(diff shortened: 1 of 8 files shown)


theduke:
@LegNeato ping

LegNeato:
I haven't looked at this super closely, but why would you ever not want `transparent` for a newtype?

Can we do it automatically for newtypes like serde does?

theduke:
Serde requires the `#[serde(transparent)]` attribute.

Technically we could just do newtype for tuple structs with a single filed, but since we are reusing the macro for two different purposes I prefer to be explicit. It also gives some more clarity as to what's happening for the reader of the code, and it's analogous to serde and `#[repr(transparent)]`.

LegNeato:
But what about this from https://serde.rs/impl-serialize.html:

> Data formats are encouraged to treat newtype structs as insignificant wrappers around the inner value, serializing just the inner value


And this link shows that the `json` serializer at least doesn't need `serde(transparent)`:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=d9d970167e8550e808b153252eb89eaf

theduke:
I always thought the [#transparent](https://serde.rs/container-attrs.html#transparent) attr was mandatory.

I think it's nicer for readability to require it.

LegNeato:
I think we should mirror serde's behavior whenever possible...it's very ergonomic and battle-tested.

LegNeato:
Any update on this? I need it for a project so happy to do the final tweaks if you don't have time.

theduke:
I removed the requirement for the `transparent` attribute.

I would'nt support anything other than single field tuple structs for now though. Better to be conservative and rethink once someone asks for it.

So should be good to merge.

mwilliammyers:
Just to nitpick - should we take out `#[graphql(transparent)]` from the examples in the docs?

PR #370, opened 2019-06-11
Add logging

### LINKED ISSUE #48: Leverage Slog And Or Log For Debugging Output
There is some debugging information that is getting outputted in the graphql_input_object macro namely on [line 70](https://github.com/mhallin/juniper/blob/master/src/macros/input_object.rs#L70). Either, logging framework like Slog or Log should be used for outputting debugging information or this line and others like it should be removed.

mwilliammyers:
As it stands right now, this is an MVP of logging, with just the bare minimum I thought useful (at least at for my work).

Some questions need to be answered:
- Should we treat logging from the perspective of Juniper users or from Juniper's perspective? 
  - e.g. should validation errors be `warn`/`error` or simply `debug` and reserve `warn` and `error` for recoverable errors and unrecoverable errors?
  - In general, how _much_ logging should we do?
- Right now, I just logged the [`execute`](https://github.com/graphql-rust/juniper/blob/5b9a0bd31b2eb6822f1181c8088715e01f821df7/juniper/src/lib.rs#L188) function. Maybe we should go 1+ levels deeper, and for example, log in the [`execute_validated_query`](https://github.com/graphql-rust/juniper/blob/acf4207ff1de862b799360d98a1bfd616b2872cc/juniper/src/executor/mod.rs#L591) function so that users can be more granular and [enable/disable](https://docs.rs/env_logger/0.6.1/env_logger/#enabling-logging) `juniper::executor` etc.
- We might want to add a warning to the docs about  logging sensitive information? We probably shouldn't log requests in case they contain passwords? 
  - I am thinking about creating a `ClearTextPassword` scalar type in my code and making the `Display`/`Debug` impl be something like `[password]`. Perhaps we add something to that effect to the book?

I didn't and probably won't (yet) do much logging of the actual parser because #138...

I am happy to keep adding commits (and squashing?) to this PR and/or do this in chunks and open smaller PRs as I add more logging.

Closes #48.

### DIFF
--- a/juniper/Cargo.toml
+++ b/juniper/Cargo.toml
@@ -37,6 +37,7 @@ fnv = "1.0.3"
 indexmap = { version = "1.0.0", features = ["serde-1"] }
 serde = { version = "1.0.8" }
 serde_derive = { version = "1.0.2" }
+log = { version = "0.4" }
 
 chrono = { version = "0.4.0", optional = true }
 serde_json = { version="1.0.2", optional = true }
--- a/juniper/src/lib.rs
+++ b/juniper/src/lib.rs
@@ -118,6 +118,7 @@ pub use juniper_codegen::{
 use juniper_codegen::{
     object_internal, GraphQLEnumInternal, GraphQLInputObjectInternal, GraphQLScalarValueInternal,
 };
+use log::{debug, warn};
 
 #[macro_use]
 mod value;
@@ -203,7 +204,9 @@ where
         let errors = validate_input_values(variables, &document, &root_node.schema);
 
         if !errors.is_empty() {
-            return Err(GraphQLError::ValidationError(errors));
+            let err = GraphQLError::ValidationError(errors);
+            warn!("{:#?}", &err);
+            return Err(err);
         }
     }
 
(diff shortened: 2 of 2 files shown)


LegNeato:
Thanks for the patch! I need to think about this for a little bit.

Empty2k12:
I think this should rather be using https://github.com/tokio-rs/tracing where the user can choose how something is logged. Additionally, traces can be used for much more than just logging, for example users could send them to Zipkin or some trace viewer and visually check how the application is performing.

mwilliammyers:
I will need to look at tracing more, but why not use the much more popular and official (maintained by the Rust lang team) [logging facade](https://crates.io/crates/log) and then users can decide which logging front end to use?

theduke:
Thanks for the PR @mwilliammyers , but I'm closing this for now.

We will want to use the [tracing](https://docs.rs/tracing) crate, especially with async support, but even without it.
It gives us the possibility for fine-grained resolver based tracing.

mwilliammyers:
@theduke I am happy to open a new PR using `tracing` (or update this one), but if we want fine-grained tracing events (e.g. `span` etc.) and async support; then why not use [`async_log`](https://github.com/rustasync/async-log)? It is built by the Rust async team and builds on the standard `log` crate, and will probably be the de facto async logging crate.

There is also [`tracing_log`](https://docs.rs/tracing-log/0.0.1-alpha.1/tracing_log/#convert-log-records-to-tracing-events) which users could use if they wanted to use `tracing`: 

```rust
use tracing_log::LogTracer;
use log;

LogTracer::init()?;

// will be available for Subscribers as a tracing Event
log::trace!("an example trace log");
```

I would really prefer to not force users to use a specific logging adapter and let them choose the crate.

Empty2k12:
You can fine-grain tracing! The idea is, that there are different implementations on subscribers which can indicate interest in a specific trace or event. These subscribers could - for example - only listen to events emitted by Tokio and Juniper, or just Tokio, or just events from a specific thread, or ... What the subscriber does with the events is also not set in stone. There's existing [logging subscribers](https://github.com/tokio-rs/tracing/tree/master/tracing-fmt) or [timing subscribers](https://github.com/jonhoo/tracing-timing), but only your imagination can set the limits here. I, for example, are tinkering with a subscriber that exports events in the [OpenZipkin](https://zipkin.io/) format, so traces can be visualized. One could also think about exporting to [Honeycomb](https://docs.honeycomb.io/working-with-your-data/tracing/).

theduke:
Open to all ideas. Let's discuss in #423.

PR #381, opened 2019-06-25
Update cargo-make

LegNeato:


theduke:
@LegNeato there is something very weird going on with Azure.

Often, the rustup installation reports that an existing installation is reused. That installation is in a non-user directory `/usr/local/bin` on Linux and `C:/Program Files/Rust` on Window.

The normal installation puts the files in the regular location at `$HOME/.cargo`.

No idea how this could happen.

theduke:
Had another very weird issue with a sudden "bash not found" error in `check_release_automation`. Fixed by rolling it into the previous script task.

Again, no idea why this happens....

PR #382, opened 2019-06-25
Add test_post_with_variables and update test_duplicate_keys method get

Jekshmek:
Added a missing post test method with variables and made a valid get method test_duplicate_keys


Jekshmek:
Not fix run

PR #384, opened 2019-06-25
Update testing juniper::http::mod

Jekshmek:
Added a missing post test method with variables and made a valid get method test_duplicate_keys and fix run testing.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,9 +9,9 @@ To get started, you can look for [issues with the "help wanted" label](https://g
 
 Before submitting a PR, you should follow these steps to prevent redundant churn or CI failures:
 
-[ ] Ensure proper formatting
-[ ] Run all tests
-[ ] Update the CHANGELOG
+- [ ] Ensure proper formatting
+- [ ] Run all tests
+- [ ] Update the CHANGELOG
 
 ### Ensure proper formatting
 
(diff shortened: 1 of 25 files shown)


LegNeato:
Thanks for the PR! Looks great, can you rebase to get CI working?

Jekshmek:
@LegNeato Tell me to understand, if you remove the duplicate key, it is still not the correct syntax!
```rust
// /?query={"query": "{hero{name}}"}
let response = integration.get("/?query=%7B%22query%22%3A%20%22%7Bhero%7Bname%7D%7D%22%7D");
assert_eq!(response.status_code, 400);
```

```rust
Is the URL correct for checking for duplicate key?
// /?query={hero{blah}}&query={hero{blah}}
let response = integration.get("/?query=%7Bhero%7Bblah%7D%7D&query=%7Bhero%7Bblah%7D%7D"); 
```
p.s. The `Hyper` integration does not check the key duplication by the current url, since the code block does not work at all responsible for this, as does `Iron`. They give the correct result for other reasons not related to the current test.

LegNeato:
Hmmm, not sure. It's been a while since I looked at this harness.

Jekshmek:
@LegNeato I think that 
```/?query = {"query": "{hero {name}}", "query": "{hero {name}}"}```
 this URL checks the correctness of the graphql query syntax but not the duplicate key, meaning not controlling truncation of data

PR #385, opened 2019-06-27
Add support for "dyn" trait object syntax

petoknm:


LegNeato:
Thank you for the patch! Can you add it to the CHANGELOG as well?

juniper_codegen/src/impl_object.rs
```diff
@@ -52,19 +52,23 @@ pub fn build_object(args: TokenStream, body: TokenStream, is_internal: bool) ->
...
-                    .unwrap()
-                    .ident
-                    .to_string(),
-                _ => {
-                    panic!("Could not determine a name for the object type: specify one with #[juniper::object(name = \"SomeName\")");
-                }
-            },
+            syn::Type::Reference(ref reference) => {
+                let path = match &*reference.elem {
+                    syn::Type::Path(ref type_path) => &type_path.path,
```
LegNeato: Do we not need to get the last segment anymore?

petoknm: This just extracts the path from the reference. Getting the last segment is done on [line 70](https://github.com/graphql-rust/juniper/pull/385/files/300d2fd352aaf0bb099e6de427fb7c857056eaf1#diff-0c158a3c35852b789d818d59989ccb55R70).

LegNeato:
Ah, missed that line. Thank you so much for the PR! :beers:

PR #386, opened 2019-06-27
Cleanup after #385

petoknm:
Just realized that I can pull the `path` variable up one level and deduplicate the `path.segments.iter().last().unwrap().ident.to_string()` part.


LegNeato:
Nice, thanks for looping back around!

PR #395, opened 2019-07-19
Mention juniper-from-schema in README

davidpdrsn:
This is kind of a shameless plug (I maintain juniper-from-schema) so feel free to close this if you don't want third-party libs mentioned in your readme.

However I think it might make sense to mention [juniper-from-schema](https://github.com/davidpdrsn/juniper-from-schema) in case people come to juniper expecting to find code generation features.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ see the [hyper][hyper_examples], [rocket][rocket_examples], [iron][iron_examples
 Juniper supports the full GraphQL query language according to the
 [specification][graphql_spec], including interfaces, unions, schema
 introspection, and validations.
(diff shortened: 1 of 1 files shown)


LegNeato:
I've been meaning to do this! Thank you for the PR and your project.

theduke:
Same!

I think we should go ahead and add a new section that lists related projects.

(the entire README could use an overhaul)

LegNeato:
I agree. I also want a marketing page for the whole `graphql-rust` org.

davidpdrsn:
I would be happy to help out 😊

Issue #391, opened 2019-07-12
Give `IntoFieldError::into_field_error` access to `Context`


davidpdrsn:
In the app I'm working on we want to track all errors in [Rollbar][]. The way we're currently doing that is by having a custom error type (`AppError`) and implementing `IntoFieldError` for that type. In `into_field_error` we then make an API call to Rollbar and return a `FieldError`.

This works well however we would like for the data we send to Rollbar to contain meta data such as the current user's id and the operation name(s). This is a bit problematic because `IntoFieldError::into_field_error` doesn't have anyway of getting to the `Context`, which is where we store these things.

We solve that by having a type that wraps a normal error and has the data from the `Context` we need. Something like:

```rust
pub struct ReportableError {
    error: AppError,
    current_user_id: Option<i32>,
    operation_names: Vec<String>,
}
```

We then have some setup for creating those errors from a `Result<_, AppError>` and the `Executor`. Something like:

```rust
let result: Result<_, ReportableError> = some_result_value
    .map_err_to_reportable(executor);
```

This does work but I'm thinking we could simplify things quite a bit if `IntoFieldError::into_field_error` got the `Context` or `Executor` as an argument. That way I imagine we wouldn't need a wrapper type and the conversion should happen automatically.

What do you think?

I'm guessing this would be a breaking change.

[Rollbar]: https://rollbar.com

davidpdrsn:
I wouldn't mind giving this a shot if you think its a good idea 😊

theduke:
We definitely want a way to do this, but this is closely correlated with things like general logging and tracing.
 
My preferred solution to this is moving towards a stateful `Executor`, which makes adding things like error/reporting/tracing hooks easier.

davidpdrsn:
Fair enough. I'll close the PR.

PR #397, opened 2019-07-21
Give `IntoFieldError::into_field_error` access to the context

davidpdrsn:
Fixes https://github.com/graphql-rust/juniper/issues/391

`IntoFieldError` is useful for sending errors to error tracking services
such as Rollbar or Bugsnag. When you do that it is often useful to
include data accessible via the context such as data about the current
user.

Getting access to the context inside `IntoFieldError` previously required
[wrapping your error][] which lead to a lot of boilerplate. This changes that
so `IntoFieldError` gets passed a reference to the context.

This is a breaking change but I think the small bit of breakage is worth it. I
doubt users have more than one implementation of `IntoFieldError`.

[wrapping your error]: https://github.com/graphql-rust/juniper/issues/391

### DIFF
--- a/juniper/CHANGELOG.md
+++ b/juniper/CHANGELOG.md
@@ -1,6 +1,6 @@
 # master
 
-- No changes yet
+- Give `IntoFieldError::into_field_error` access to the context [#397](https://github.com/graphql-rust/juniper/pull/397)
 
 - Allow `mut` arguments for resolver functions in `#[object]` macros [#402](https://github.com/graphql-rust/juniper/pull/402)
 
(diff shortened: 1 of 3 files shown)


theduke:
Hey, sorry for the late reply.

I agree that we want facilities for error reporting, but:

I'm not really onboard with this design. `IntoFieldError` is really just meant for basic error conversion as in: `Into<OurError>`. I'm  uncomfortable with encouraging this kind of mapping logic there.
I'll comment a bit more on the issue.

davidpdrsn:
Closed because of https://github.com/graphql-rust/juniper/issues/391#issuecomment-514840485

Issue #398, opened 2019-07-22
"Panic when using look ahead from fragment with nested type" has come back in 0.13.0


davidpdrsn:
It seems https://github.com/graphql-rust/juniper/issues/335 has come back in 0.13.0

**Describe the bug**
Given the schema

```graphql
schema {
  query: Query
}

type Query {
  users: [User!]!
}

type User {
  country: Country!
}

type Country {
  id: Int!
}
```

and the query

```graphql
query Query {
    users {
        ...userFields
    }
}

fragment userFields on User {
    country {
        id
    }
}
```

Juniper panics if you call `executor.look_ahead();` inside the `User::country` resolver.

Message is

```
thread 'main' panicked at 'lookahead to find the field', src/libcore/option.rs:1036:5
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

**To Reproduce**
I have made a reproduction script here https://github.com/davidpdrsn/juniper-look-ahead-with-fragments-error. Download that and run `cargo run` to see the panic.

theduke:
Looking into it.

PR #404, opened 2019-07-24
Don't assume lookaheads are always fields

LegNeato:
Fixes https://github.com/graphql-rust/juniper/issues/398.

### DIFF
--- a/integration_tests/juniper_tests/src/issue_398.rs
+++ b/integration_tests/juniper_tests/src/issue_398.rs
@@ -43,7 +43,7 @@ impl Country {
 type Schema = juniper::RootNode<'static, Query, EmptyMutation<()>>;
 
 #[test]
-fn it_works() {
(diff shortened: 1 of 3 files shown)


integration_tests/juniper_tests/src/issue_398.rs
```diff
@@ -0,0 +1,67 @@
...
+impl Country {
+    fn id(&self) -> i32 {
+        self.id
+    }
+}
+
+type Schema = juniper::RootNode<'static, Query, EmptyMutation<()>>;
+
+#[test]
+fn it_works() {
```
theduke: That's great, but a different name would still be good. :P

LegNeato: 😅

davidpdrsn:
Thanks for fixing this! Any chance you would publish a patch release? The bug is currently preventing me from updating my work project to 0.13.

PR #412, opened 2019-08-15
Fix deprecation notice when using macros

LegNeato:


theduke:
Needs an additional trait bound (and potentially a `.as_str()` )

PR #414, opened 2019-08-18
Remove GraphQL prefix from derives + add deprecated fallbacks

theduke:
This commit renames all derive macros to a version without the
GraphQL prefix.

The prefix is no longer needed with the 2018 edition,
since they can be specified with the `juniper::` prefix or
imported and conflcits can be avoided.

The old names are still kept around as a deprecated version.

### DIFF
--- a/juniper/CHANGELOG.md
+++ b/juniper/CHANGELOG.md
@@ -1,6 +1,7 @@
 # master
 
-- No changes yet
+- Rename all derive macros by removing the `GraphQL` prefix.
+  The old names are supported by deprecated fallbacks.
 
 # [[0.13.1] 2019-07-29](https://github.com/graphql-rust/juniper/releases/tag/juniper-0.13.1)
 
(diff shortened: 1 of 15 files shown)


theduke:
Todo: 
- [ ] Update the book.

Also not sure if we want to merge this yet or wait for some other changes like async/await.

LegNeato:
I'm not sure we want to do this TBH. It really reduces clarity and grepability and could impact readability in large codebases. 

Words like `Enum`, `Interface`, etc are too generic.

Essentially it forces prefixing with `juniper::` to stay clear, and while mostly that's how it will be used you can also `use juniper::*` and it would be less clear. With the GraphQL prefix it stays clear regardless of how it is used.

theduke:
True, I'm also not sure.
We should probably have some issue where we talk about a general API overhaul, there is quite a bit of accumulated cruft.
Also, if we don't want `Object` etc, we also don't want `juniper::object`, but `juniper::graphql_object` etc.

PR #417, opened 2019-08-21
Update juniper_rocket for async.

obmarg:
This updates juniper_rocket to support async Juniper & Rocket.  Async rocket is on a completely seperate branch just now, so had to pull that in.  I couldn't hide the futures-preview dependency behind a feature flag as Rocket hasn't put async behind a feature flag as far as I can see.


theduke:
Thanks.

theduke:
Also thanks for the fix... I'll stop working on this branch directly and go through PR s myself now to avoid that.

theduke:
Also thanks for the fix... I'll stop working on this branch directly and go through PR s myself now to avoid that.

PR #418, opened 2019-08-23
Add introspection async benchmark

LegNeato:


theduke:
As mentioned in the issue, this is not a meaningful benchmark atm since the schema types skip all the async logic. (See the `noasync` annotation on the schema types) because we know those types do not contain any nested async resolvers.

We can still merge them to act as something like a regression test, but best with a comment explaining this.

Issue #415, opened 2019-08-20
Implement std::error::Error for all public error types


theduke:
* GraphQLError
* ParseError
* LexerError
* RuleError
* FieldError

davidpdrsn:
I looked into this yesterday and it isn't totally straight forward. `std::error::Error` requires a `Display` implementation which `Spanning<ParseError<'a>>` and `RuleError` doesn't have. They should be possible to add but I wasn't sure what the most helpful implementations might be.

I might just make some initial implementation so we have something to discuss.

theduke:
All errors should get a `Error` impl, it's a big oversight that they didn't have it yet.

The only limitation here is that we can't include surrounding input query context in the generated error message but that's fine I think.

PR #419, opened 2019-08-23
Implement `std::error::Error` for all error types

davidpdrsn:
Fixes https://github.com/graphql-rust/juniper/issues/415

- [x] GraphQLError
- [x] ParseError
- [x] LexerError
- [x] RuleError
- [x] FieldError

---

I'm having some issues with `FieldError`. I end up hitting a conflicting implementation error when I add

```rust
impl<S> std::error::Error for FieldError<S> {}

impl<S> fmt::Display for FieldError<S> {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        unimplemented!()
    }
}
```

I get this error:

```rust
error[E0119]: conflicting implementations of trait `std::convert::From<executor::FieldError<_>>` for type `executor::FieldError<_>`:
   --> juniper/src/executor/mod.rs:142:1
    |
142 | / impl<T: Display, S> From<T> for FieldError<S>
143 | | where
144 | |     S: crate::value::ScalarValue,
145 | | {
...   |
151 | |     }
152 | | }
    | |_^
    |
    = note: conflicting implementation in crate `core`:
            - impl<T> std::convert::From<T> for T;
```

I'm not quite sure what is causing that.


juniper/src/lib.rs
```diff
@@ -184,6 +186,28 @@ pub enum GraphQLError<'a> {
...
+                let msgs = errors
+                    .iter()
+                    .map(|error| format!("{}", error))
+                    .collect::<Vec<_>>()
+                    .join("\n");
+
+                write!(f, "{}", msgs)
+            }
+            GraphQLError::NoOperationProvided => write!(f, "No operation provided"),
+            GraphQLError::MultipleOperationsProvided => write!(f, "No operation provided"),
```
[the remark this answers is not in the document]
davidpdrsn: e8b627a49e266 👍

theduke:
> I'm not quite sure what is causing that.

Since FieldError has a `From<T: Display>` impl, if it implements Display itself the compiler can't resolve that cyclical relationship.

Not sure what's the best way around that, I'll have a look at the code.

theduke:
I never was a big fan of the `From T: Display` impl to be honest.

We could think about removing it, and only having a `From T: Error` impl.
We can still have a manual constructor like `FieldError::from_str`.

davidpdrsn:
@theduke I like that. Since thats a breaking change do we have to go through a round of deprecation or is it fine?

theduke:
Nah let's batch it together with the other changes related to async/await and removing the old macros.

What's the status here, did you implement the FieldError?

davidpdrsn:
I had totally forgotten about this. I'll give it another look soon.

davidpdrsn:
Sorry for the delay but this should be ready for next review.

I had to make a few changes to get `impl std::error::Error for FieldError` working. They're outlined in the [commit message](https://github.com/graphql-rust/juniper/pull/419/commits/38751fac17d6f8f5f4f5da3d58257aabcdd9f20d).

theduke:
This seems to be missing a `From<E: std::error::Error>` impl right now?

We need that more than a the `From<String/&str>` / variants.

davidpdrsn:
I did try adding `From<E: std::error::Error>` but unfortunately it gives the same error:

```
error[E0119]: conflicting implementations of trait `std::convert::From<executor::FieldError<_>>` for type `executor::FieldError<_>`:
   --> juniper/src/executor/mod.rs:182:1
    |
182 | / impl<S, E> From<E> for FieldError<S>
183 | | where
184 | |     S: crate::value::ScalarValue,
185 | |     E: std::error::Error,
...   |
189 | |     }
190 | | }
    | |_^
    |
    = note: conflicting implementation in crate `core`:
            - impl<T> std::convert::From<T> for T;
```

It seems that you cannot both implement `Display` and `From<T> where T: Error`.

```rust
use std::fmt::{Formatter, self, Display};

#[derive(Debug)]
struct MyError(String);

impl std::error::Error for MyError {}

impl Display for MyError {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        unimplemented!()
    }
}

impl<T: std::error::Error> From<T> for MyError {
    fn from(s: T) -> Self {
        unimplemented!()
    }
}
```

That fails to compile with the same error.

That is why I added

```rust
pub fn from_error<E>(error: E) -> FieldError<S>
where
    E: std::error::Error,
{
    FieldError::from(error.to_string())
}
```

Was the only working solution I could think of.

davidpdrsn:
Does it make sense to not implement std::error::Error for FieldError because of this? I’m thinking it’s pretty inconvenient 🤔

theduke:
Ah yeah I forgot about the coherence rules. You cant both implement `Error` and `From<Error>` without specialization...

Yeah let's just drop the impl on FieldError. It doesn't really matter anyway since it is not a type users work with directly.

davidpdrsn:
Alright cool. I’ll revert it tomorrow.

Does it make sense to keep the Display impl for Value?

davidpdrsn:
I have brought back `impl<T: Display, S> From<T> for FieldError<S>` now.

Guess this should be good now 😊

theduke:
I have some additional changes. Will push tomorrow.

davidpdrsn:
@theduke Anything I can help out with?

jkoudys:
@davidpdrsn Is there anything more to go in on this PR? juniper is the last holdout in my app before I can simply `Box<dyn Error>` then use `?` like mad everywhere. The open change request looks like a pretty trivial message change.

juniper/src/lib.rs
```diff
@@ -188,6 +189,29 @@ pub enum GraphQLError<'a> {
     IsSubscription,
 }
 
+impl<'a> fmt::Display for GraphQLError<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            GraphQLError::ParseError(error) => write!(f, "{}", error),
+            GraphQLError::ValidationError(errors) => {
+                let msgs = errors
```
MikailBag: I'd rewrite this code to:
```rust
for err in &errors {
    writeln!(f, "{}", err)?;
}
Ok(())
```
It is shorter and avoids allocations.

davidpdrsn: Good call. b7eb36a858217

davidpdrsn:
@jkoudys No it should be good to go. I wonder which error you're having to convert into `Box<dyn Error>`? I haven't run into such issues with the juniper version that is out now.

Also note that as per the discussions above you wont be able to convert`FieldError` into `Box<dyn Error>`. That shouldn't be an issue in practice however.

jkoudys:
> @jkoudys No it should be good to go. I wonder which error you're having to convert into `Box<dyn Error>`? I haven't run into such issues with the juniper version that is out now.

Trying to into a `Box<dyn Error>` using the `?`, but erroring on `^ the trait `std::error::Error` is not implemented for `juniper::GraphQLError<'_>`` because GraphQLError doesn't impl Error. I'm on the async-await branch, so that may need to be rebased.

davidpdrsn:
@theduke @LegNeato This has been done for quite some time as you can see someone is asking for it 😊 Do you have any last minute things I should correct before merging? I might also try to get it merged into the async/await branch afterwards.

andy128k:
I think proper implementation of std::error::Error should include implementation of `source` method too.

jkoudys:
@andy128k you mean like on the GraphQLError itself, to give a simple mapping back to the original error type? e.g.

```rust
impl<'a> std::error::Error for GraphQLError<'a> {
    fn source(&self) -> Option<&(dyn Error + 'a)> {
        Some(&self)
    }
}
```
Can't think what else we'd need, unless you mean `source`s on e.g. ParserError could return a LexerError or something.
At any rate, does this really need to go in before accepting the PR? It's been done but sitting in limbo for such a long time. Juniper is the one holdout we have that's keeping us from having an easy time managing only Errors that impl std:error::Error. 

It's been 4 months....

davidpdrsn:
This PR has been blocked by all async-flux. That seems to have settled now so I do hope to get it merged soon.

davidpdrsn:
@LegNeato master still doesn't build due to some async stuff it seems. Should I merge this into the async branch instead?

LegNeato:
Everything on master except the book builds / passes?

davidpdrsn:
If I clone a fresh copy of the repo and run `cargo test` on nightly I get a bunch of errors like these:

```
error[E0432]: unresolved import `futures`
  --> integration_tests/juniper_tests/src/codegen/derive_enum.rs:10:10
   |
10 | #[derive(juniper::GraphQLEnum, Debug, PartialEq)]
   |          ^^^^^^^^^^^^^^^^^^^^ use of undeclared type or module `futures`

error[E0432]: unresolved import `futures`
  --> integration_tests/juniper_tests/src/codegen/derive_enum.rs:19:10
   |
19 | #[derive(juniper::GraphQLEnum)]
   |          ^^^^^^^^^^^^^^^^^^^^ use of undeclared type or module `futures`

error[E0432]: unresolved import `futures`
  --> integration_tests/juniper_tests/src/codegen/derive_enum.rs:29:10
   |
29 | #[derive(juniper::GraphQLEnum, Debug, PartialEq)]
   |          ^^^^^^^^^^^^^^^^^^^^ use of undeclared type or module `futures`

error[E0432]: unresolved import `futures`
  --> integration_tests/juniper_tests/src/codegen/derive_enum.rs:37:10
   |
37 | #[derive(juniper::GraphQLEnum, Debug, PartialEq)]
   |          ^^^^^^^^^^^^^^^^^^^^ use of undeclared type or module `futures`

error[E0432]: unresolved import `futures`
  --> integration_tests/juniper_tests/src/codegen/derive_object.rs:13:10
   |
13 | #[derive(GraphQLObject, Debug, PartialEq)]
   |          ^^^^^^^^^^^^^ use of undeclared type or module `futures`

error[E0432]: unresolved import `futures`
  --> integration_tests/juniper_tests/src/codegen/derive_object.rs:29:10
   |
29 | #[derive(GraphQLObject, Debug, PartialEq)]
   |          ^^^^^^^^^^^^^ use of undeclared type or module `futures`

error[E0432]: unresolved import `futures`
  --> integration_tests/juniper_tests/src/codegen/derive_object.rs:38:10
```

LegNeato:
@davidpdrsn yeah weirdly just doing `cargo test` doesn't work for integration tests. Haven't looked into it yet. In any case, you can see the non-Github CI is all passing (Github CI failures are https://github.com/graphql-rust/juniper/issues/531.

davidpdrsn:
Thanks for merging!

PR #422, opened 2019-08-30
Juniper Chinese documentation

zzy:
Juniper Chinese documentation.

All markdown files are completed, and run the tests on CI


theduke:
Hey, this is awesome!

The only problem I see is that we won't be able to keep the book updated ourselves, but will have to rely on contributors to do so. 

Chinese is probably the only translation we would get for a while (or ever) though. 

Options:
* Just have it in this repo, but not run the tests on CI, and add a warning about it being potentially outdated
* Have translations in a semi-official status in a `graphql-rust/juniper-book-intl` repo, but linked from here

I'm leaning towards the second option. (@LegNeato )

zzy:
> 
> 
> Hey, this is awesome!
> 
> The only problem I see is that we won't be able to keep the book updated ourselves, but will have to rely on contributors to do so.
> 
> Chinese is probably the only translation we would get for a while though (probably ever).
> 
> Options:
> 
>     * Just have it in this repo, but not run the tests on CI, and add a warning about it being potentially outdated
> 
>     * Have translations in a semi-official status in a `graphql-rust/juniper-book-intl` repo, but linked from here
> 
> 
> I'm leaning towards the second option. (@LegNeato )

Hi, @theduke , thank you.  

Recently, our team developed graphql server using actix-web + juniper + diesel. in order to get help quickly, I translated the book. I want to share it for other Chinese  pay attention to Juniper.

Have a nice day.

PR #424, opened 2019-09-04
juniper_codegen: Drop regex & lazy_static dependencies

archseer:
https://www.reddit.com/r/rust/comments/cz7u0j/psa_regex_13_permits_disabling_unicodeperformance/

The regex dependency is used in a single place, and it only validates ascii characters:

https://github.com/graphql-rust/juniper/blob/a3caf126b03533ea9b75b64056152500d38ce148/juniper_codegen/src/util.rs#L283

As such, we can drop the unicode support for faster compile times.

### DIFF
--- a/juniper_codegen/Cargo.toml
+++ b/juniper_codegen/Cargo.toml
@@ -18,8 +18,6 @@ proc-macro = true
 proc-macro2 = "1.0.1"
 syn = { version = "1.0.3", features = ["full", "extra-traits", "parsing"] }
 quote = "1.0.2"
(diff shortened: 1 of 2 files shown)


theduke:
Hey, thanks.

It would actually be really trivial to do the validation with a simple helper method.
Then we could completely drop the dependency.

theduke:
@archseer you feel like writing up the validation as a simple helper function and remove regex?

archseer:
Sure! I'll see if I can come up with something simple tomorrow.

Issue #54, opened 2017-06-17
Support subscriptions


LegNeato:
Looks like [subscriptions made it into the spec](https://facebook.github.io/graphql/#sec-Subscription).

While most users will implement subscription logic outside `juniper` (like how Facebook interfaces with a MQTT server handling the actual delivery and connections) it might be useful to add some integrations as optional features.

mhallin:
Wow, I hadn't seen that happen. It certainly complicates things :)

I'll have to look and see how different implementations solve this, but it seems that there are (at least) three parts to this feature:

* Storing/managing active subscriptions.
* An API to push messages into Juniper.
* An API to get transformed messages from Juniper out to clients.

We could probably use `Stream` from the futures crate to represent the two latter points, while the first one is just a trait that you have to implement for the backing storage you're interested in.

For convenience, it would be nice to have some reasonable implementations for these points behind feature flags just like the Iron and Rocket integrations. What it'll be specifically depends on the state of respective crates - I haven't looked at any of the AMQP/MQTT or websockets crates that are available.

LegNeato:
That sounds about right to me! The AMQP and MQTT crates look pretty early days sadly...

japrogramer:
In graphene they just went with a consumer type interface where on next the user decides how the message is transported.

tailhook:
I've just prototyped subscriptions interface for my app. The basic idea is to [rewrite subscription to a query](https://github.com/tailhook/cantal/compare/5d10a42...a2534dc#diff-367228009ff62cc47518ecd0e815a7bcR190) using [grapqhql-parser](https://crates.io/crates/graphql-parser) and execute the query each time underlying data is updated. This works fine except serializing request and parsing it again is surely quite costly.

Another finding is that [subscriptions-transport-ws protocol](https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md) fits serde model [quite fine with some annotations](https://github.com/tailhook/cantal/compare/5d10a42...a2534dc#diff-9930a6616519833751ecf2db5026d6bbR22) (at least parts of the proto I needed so far).

So in the first iteration, I would say that API for subscriptions may look like this:
```
let (data, sub) = execute_with_subscription(...);
```
Where the function works similarly to ``execute``, and ``data`` is result of a query. Except it's unclear what ``sub`` here means.

In the first iteration, it might even be ``true`` if request contains subscriptions. Meaning that you must manually inspect AST to find out events you need to subscribe to.

[...]

PR #430, opened 2019-09-25
Add subscription support to query parser

nWacky:
Recognize subscriptions when parsing queries and return `Err(IsSubscription)` from `execute_validated_query()`  (called inside `execute()`) if query is of subscription type.

(will be useful if there will be a separate method [`subscribe()` for handling subscriptions](https://github.com/graphql-rust/juniper/issues/54#issuecomment-534506503) and `execute()` will only handle queries)

_This PR changes introspection queries (added type `SUBSCRIPTION`)_

### DIFF
--- a/juniper/CHANGELOG.md
+++ b/juniper/CHANGELOG.md
@@ -1,6 +1,6 @@
 # master
 
-- No changes yet
+- Add ability to parse 'subscription'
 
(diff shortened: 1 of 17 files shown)


LegNeato:
This looks reasonable, thanks!

LegNeato:
Just published a release with this change 🎉 . Thanks again! 👍

PR #431, opened 2019-09-29
Improve visitability of lookahead types.

jimsynz:
We're planning on using Juniper to power our API in the future, however we want to be able to dynamically compile our queries so that we can support a similar interface to our [existing ODATA solution](https://github.com/balena-io/pinejs/).  In order to support this we need to be able to visit the entire lookahead tree for an inbound query.

I've extended `LookAheadMethods` to add a few helper methods, the most important of which is the `child_names` method, which means that we're able to use it to iterate the children of each node.

I've also added `name` to `LookAheadArgument` for the same reason.

### DIFF
--- a/juniper/CHANGELOG.md
+++ b/juniper/CHANGELOG.md
@@ -1,10 +1,10 @@
 # master
 
-- No changes yet
+- Improve lookahead visitability.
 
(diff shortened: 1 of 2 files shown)


LegNeato:
Thanks for the PR! 🍻

LegNeato:
Just published a release with this change 🎉 . Thanks again! 👍

jimsynz:
Wow. That was fast! Thanks!

PR #433, opened 2019-09-30
Add subscription support

nWacky:
**Before merge:** Resolve `TODO#433` (import crates in subscriptions example from GitHub instead of local)

**Related issue: #54** 

## How to implement custom GraphQL subscription

*Note: by `SubscriptionCoordinator` and `SubscriptionConnection` I mean types implementing these traits*

Current execution logic is inspired with [this](https://github.com/graphql-rust/juniper/issues/54#issuecomment-534506503) and [this](https://github.com/graphql-rust/juniper/pull/433#issuecomment-553570544) idea. In a nutshell:

When a request is received, it gets processed by a `SubscriptionCoordinator`. This struct decides if incoming request should be resolved into a new stream or if one of already existing streams (that coordinator knows about) is good enough. `juniper_subscriptions::Coordinator` resolves each request into a new stream. 

`SubscriptionCoordinator` should call `juniper::http::resolve_into_stream` each time it needs to resolve a `GraphQLRequest` to `Value<Stream<Item = Value<S>>`. If subscripiton-related macros are used, `Value::Object<Stream<Item = Value<_>>` is returned. It keeps each field and corresponding `Stream<Item = Value<_>>`. `Object`s can be iterated over and needed GraphQL objects can be assembled from returned `Stream`s. When iterating over an object returned from subscription macros, each object field should be mapped to `Value::Scalar<StreamT>`, where `StreamT` is the type user has specified in field resolvers. 

`#[juniper::graphql_subscription]` macro implements `GraphQLType` + `GraphQLSubscriptionType`. `GraphQLType` is used to store which fields this subscription has, which types they have, etc. Its resolvers ( `resolve`, `resolve_field`, `resolve_into_type`) are not used. `GraphQLSubscriptionType`'s `resolve_into_stream` (called by Executor), `resolve_field_into_stream`, `resolve_type_into_stream`  (called by `resolve_into_stream`) are used instead. 

Once a `Value<Stream>` is returned from `resolve_into_stream`, it is returned to the user (user most probably called `juniper::http::resolve_into_stream` to get to resolvers and resolve a request into stream).

Once a stream is returned, it could be put inside a `SubscriptionConnection` (if called by `SubscriptionCoordinator`). `SubscriptionConnection` implements `Stream<Item = GraphQLResponse>`. `GraphQLResponse` can be deserialized and sent to end user => `SubscriptionConnection` can be polled by server implementation. `SubscriptionConnection` could also have functionality for stopping, reconnecting, etc (will be implemented in the follow-up PRs).

`SubscriptionCoordinator` and `SubscriptionConnection` are types that manage and control already obtained `Stream`s. `GraphQLSubscriptionType` manages resolvers on GraphQL objects.

<details><summary><b>Data flow in `warp_subscriptions` example</b></summary>
<p>

- HTTP server is initialized with `RootNode` and `SubscriptionCoordinator`  

1. server gets a request 
2. request is deserialized into `GraphQlRequest`
3. `graphql_subscriptions_async(ws_connection, coordinator)` - example's helper for subscription execution
    1. `coordinator.subscribe(request, context)` - resolve `GraphQlRequest` into `Stream` and return `SubscriptionConnection` containing that stream
        - `juniper::http::resolve_into_stream(...)` - get operation name and variables from request  
        - `crate::resolve_into_stream(...)` - parse document, make sure no errors are found 
        - `executor::resolve_validated_subscription(...)` - create executor and resolve subscription query into `Value<Stream<Item = Value<S>>>`
        - `executor.resolve_into_stream(...)` - check if `self.subscribe` has errors and push them to executor
        - `self.subscribe(...)` - return `value.resolve_into_stream(...)`.
        - `value.resolve_into_stream(...)` - user logic how to resolve this value into stream
            - by default, each value `V` returned from stream is treated as a if a query was executed asynchronously and 
              `V` was returned. Ongoing execution logic creates new `Value::Object` with fields requested by user 
              (for example, if `user {id name}` was requested, each `User { id name email }` returned from stream will be returned as `User { id name }` GraphQL object)
    2. Returned `SubscriptionConnection` can be used as a `Stream` over `GraphQLResponse`s, which can be deserialized and sent to the end user.
        `juniper_subscriptions::Connection` waits until all object's fields are resolved and then returns the whole object. 
    3. Every time a `GraphQlResponse` is yielded from a stream, send it to end user
</p>
</details>

<details><summary><b>Starting playground server</b></summary>
<p>

- Go to `examples/warp_subscriptions` and `cargo run`. New warp server should start on [localhost:8080](http://localhost:8080/) 
-  Subscription queries should be able to be executed at this point:
```graphql
subscription {
    users {
        id
        kind
        name
        friends {
            id
            kind
        }
    }
}
```
- New users are returned every 5 seconds. Fragments and multiple subscriptions are also supported.

<details><summary><b>Note:</b> playground doesn't support anonymous subscriptions referencing fragments (prisma-labs/graphql-playground/issues/1067), so the following request can be sent to `localhost:8000/subscriptions` to test it</summary>
<pre><code>
{"id":"1","type":"start","payload":{"variables":{},"extensions":{},"operationName":null,"query":"subscription { ...myFragment } fragment myFragment on Subscription { users { id name friends { id name } }"}}
</code></pre>
</details>

</p>
</details>


theduke:
Hey, awesome that you are working on this!

I'll browse the code.

tyranron:
@theduke this is still quite a WIP and we will come through multiple iterations until it will ready for serious review.

However, your comments and considerations would be vital for us at any stage.

nWacky:
@theduke,

Great, thanks for code review. It is quite in some state at the moment: it's possible to create a subscription, map it to some user code with `SubscriptionHandlerAsync::resolve_field_async` (for async subscriptions only) and return an iterator/stream.

Issues I know about that I will try to resolve today/tomorrow:
- [ ] Subscription handler returning all values from the first stream, then all values from the second and so on

juniper/src/executor/mod.rs
```diff
@@ -857,6 +1054,10 @@ where
                 .schema
                 .mutation_type()
                 .expect("No mutation type found"),
+            OperationType::Subscription => root_node
+                .schema
+                .subscription_type()
+                .expect("No mutation type found"),
```
LegNeato: Nit: s/mutation/subscription/

nWacky:
This branch is still WIP, but slowly getting there. 
Here's what's new:
- With using `Value<ValuesStream<S>>` for asynchronous subscriptions and `Object<ValuesIterator<S>>` for synchronous subscriptions (I will eventually figure out which type it's better to return later) user is now able to get field names and streams/iterators
- Updated futures version to `0.3.0-alpha.19`

LegNeato:
Ok, I landed @tyranron's PR and rebased the async branch onto master.

nWacky:
Still WIP.

This week I added functionality for filtering out response fields (for synchronous subscriptions only).

For now it works this way: after subscription fields are resolved and some iterator over user's type is returned, each returned object is treated as a separate query. (for example, got `Iterator<Human>` where `Human` is user's struct. After getting each object `executor.replaced_context(ctx).resolve_with_ctx(&(), &r)` is called, which creates `Value::Object` with fields which were specified in the request).

The issue I had was with `Executor` not living long enough because it stores a lot of references to variables, which are defined in `execute_validated_subcription` (and I am mapping `Executor` to returned iterator to resolve each variable when it is returned):
```rust
...
request is deserialized --->
request.subscribe(root_node, context) --->
    ...
    executor::execute_validated_subcription(...) 
     
        // Parse definitions and set default variables
        let errors = RwLock::new(Vec::new()); // create a variable to store errors in
        // Create `Executor` (&errors and other references are passed there)
        executor.resolve_into_iterator(...) // next function for execution process
        Ok((value, errors)) //returns execution result and errors
      --->
    ...
    value.resolve_into_stream(..., &executor) ---> 
     
        let iter: Iterator<Obj> = (|| {
            // User creates `Iterator<Obj>`
        })();
        
        // At this point it's necessary to add user's object resolver
        iter.map(|x| executor.resolve_with_ctx(x)) 
        
     --->
HTTP server got iterator / stream --->
rather return it as a bunch of objects or one by one;
```

So that `Executor` lives no longer than `execute_validated_subcription()`. 

When we discussed it we decided to create new `OwnedExecutor`, which is going to store all variables. However, `Executor`'s `fragments` is a reference to a hash map pointing to a reference to local `Fragment<S>` vector. 

So (at least for now) I just moved all these variables to a function where `subscribe` is called

```rust
let mut executor_variables = juniper::OwnedExecutor::new(); // this is where Executor variables are stored (and it can return Executor)
let mut fragments = vec![]; // this is here to escape `OwnedExecutor` self referencing 
let mut executor = juniper::OptionalExecutor::new(); // this is `Option<Executor>` to keep because executor is still declared in `execute_validated_subcription`
let (res, err) = request.subscribe(
    root_node,
    context,
    &mut executor_variables,
    &mut fragments,
    &mut executor
).0.unwrap();
```

theduke:
Sorry for the delay, I'll finally be able to carve out some time this weekend.

nWacky:
@theduke

It's still in progress, so you didn't miss anything important 🙃
But any comments or suggestions are welcome

Type1J:
We're pretty happy about this PR. Keep up the good work!

nWacky:
Still WIP, but this push can be considered somewhat like early preview.

- Added `#[juniper::subscription]` macro
- Resolved some todos
- Updated schema in `juniper_benchmarks` (tests still do not compile)

What's left:

- Fragment spread and inline fragment support
- Update tests

juniper/src/schema/model.rs
```diff
@@ -186,6 +219,11 @@ impl<'a, S> SchemaType<'a, S> {
             } else {
                 None
             },
+            subscription_type_name: if &subscription_type_name != "_EmptyMutation" {
```
LegNeato: This should be `_EmptySubscription`

juniper/src/types/async_await.rs
```diff
@@ -43,6 +48,57 @@ where
...
+
+    #[allow(unused_variables)]
+    fn stream_resolve_into_type(
+        &self,
+        info: &Self::TypeInfo,
+        type_name: &str,
+        selection_set: Option<&[Selection<S>]>,
+        executor: &Executor<Self::Context, S>,
+    ) -> Result<Value<ValuesStream<S>>, FieldError<S>> {
+        // todo: cannot resolve by default because future is returned
```
LegNeato: I believe we can use https://github.com/dtolnay/async-trait here.

juniper/src/schema/model.rs
nWacky: I think we need one.
There are both [`EmptyMutation`](https://github.com/instrumentisto/juniper/blob/async-await-subscriptions/juniper/src/types/scalars.rs#L298) and `EmptySubscription` (can't see it on GitHub for some reason, will re-push later) in `src/types/scalars.rs`  

As I understood, these are used for schemas only with queries

LegNeato:
So much great work has gone into this, keep it up! 🍻

nWacky:
What's new in recently pushed commits:
- Added subscriptions example for warp
- Changed `juniper_warp`'s `warp` version to one from GitHub (with tokio 2)
- Fixed some bugs in introspection query
- Updated juniper's `playground` endpoint with latest version and separate subscriptions url

<details><summary>Subscriptions example walkthrough</summary>
<p>


- Go to `examples/warp_subscriptions` and run `cargo run`
- Visit `localhost:8080`
- Paste test subscription from below and press ▶
```graphql
subscription  {
  users {
    id
    name
    friends {
      id
      name
    }
  }
}
```
- Every 5 seconds new user should be received. First 3 users will have different sets `friends`, each for their own id
</p>
</details>

tyranron:
@theduke @LegNeato the initial raw implementation is landed. Now we're starting to make review cycles for polishing, refactorings, etc. Once we're satisfied with it, we will ping you to review it as maintainers.

nWacky:
@theduke, @LegNeato,

As @tyranron mentioned earlier the initial raw implementation is landed. There is a problem with default fragments resolving though. 

For now, fragments can be executed in subscriptions only if user's `iter_resolve_into_type`/`stream_resolve_into_type`  implementation exists, but they could be executed by default implementation if fragment's type is the same as current selection set.

For example, a fragment inside this subscription could be executed by default implementation
```graphql
subscription { # this fragment is of MySubscription type
  ... on MySubscription { # this fragment is of MySubscription type as well =>
                          # it would be possible to resolve it with default implementation  
    human(id: "32") {
      id
    }
  }
}
```

But in [`iter_resolve_into_type`](https://github.com/instrumentisto/juniper/blob/1f1837b5c269c1839f5680db1b0e8b16aad9ebe4/juniper/src/types/base.rs#L387) it's not possible to return an iterator because iterator should live longer than executor:
```rust
fn iter_resolve_into_type<'s, 'i, 'tn, 'ss, 'e, 'res>(
    &'s self,
    info: &'i Self::TypeInfo,
    type_name: &'tn str,
    selection_set: Option<&'ss [Selection<S>]>,
    executor: Rc<Executor<'e, Self::Context, S>>, // <--- executor is valid for 'e
) -> Result<Value<ValuesIterator<'res, S>>, FieldError<S>>  // <--- iterator should be valid for 'res 
                                                            // because executor will be used later on 
                                                            // to filter out object fields that user didn't ask for
where    
    's: 'res,
    'i: 'res, 
    'tn: 'res, 
    'ss: 'res, 
    'e: 'res
{ 
    if Self::name(info).unwrap() == type_name {
        Ok(self.resolve_into_iterator(info, selection_set, executor)) // <--- and iterator cannot be passed by value 
                                                                      // because it is passed by reference to resolve_into_iterator   
                                                                      // (and passing iterator by value to resolve_into_iterator
                                                                      // causes the same error in resolve_into_iterator)
    } else {
        panic!("iter_resolve_into_type must be implemented by unions and interfaces");
    }
}
```

For now, default `iter_resolve_into_type` panics

tyranron:
@nWacky 

> Use `Arc` in `EmptySubscription` not to have `unsafe impl Send`

That's not the topic of this PR, so remove it from the list.

nWacky:
Previous commits resolve [these todos](https://github.com/graphql-rust/juniper/pull/433#issuecomment-549860964) and rebase current branch onto `async-await`

examples/warp_subscriptions/src/main.rs
```diff
@@ -0,0 +1,213 @@
...
+                },
+            ];
+        } else {
+            return vec![];
+        }
+    }
+}
+
+#[derive(juniper::GraphQLObject)]
+#[graphql(description = "A humanoid creature in the Star Wars universe")]
```
LegNeato: FYI, you can use normal rust doc comments here and they will be used for GraphQL

juniper_codegen/src/util.rs
```diff
@@ -1021,10 +1017,470 @@ impl GraphQLTypeDefiniton {
...
+        let idents_of_path = path
+            .segments
+            .iter()
+            .into_iter()
+            .fold(String::new(), |mut acc, v| {
+                acc.push_str(&v.ident.to_string());
+                acc.push('|');
+                acc
+            });
+        vec!["Result|", "std|result|Result|", "core|result|Result|"]
```
LegNeato: Should this include `FieldResult` too?

juniper_warp/src/lib.rs
```diff
@@ -289,68 +317,243 @@ where
...
+                    got_close_signal.store(true, Ordering::Relaxed);
+                }
+                _ => panic!("unknown type"),
+            }
+
+            futures03::future::ready(())
+        })
+        .for_each(|_| {
+            // empty `for_each` here to keep executing this stream
+            async {}
```
LegNeato: Why is this needed? This seems pretty "magical", I wonder if there are helpers in the `futures` crate.

juniper_warp/src/lib.rs
```diff
@@ -37,17 +37,29 @@ Check the LICENSE file for details.
 */
 
 #![deny(missing_docs)]
-#![deny(warnings)]
+//#![deny(warnings)]
```
LegNeato: add back

juniper_warp/src/lib.rs
```diff
@@ -200,12 +217,12 @@ where
 /// let graphql_filter = make_graphql_filter(schema, context_extractor);
 ///
 /// let graphql_endpoint = warp::path("graphql")
-///     .and(warp::post2())
+///     .and(warp::post())
```
LegNeato: Why this change?

juniper_warp/src/lib.rs
```diff
@@ -231,9 +256,10 @@ where
                 .and_then(|result| ::futures::future::done(Ok(build_response(result))))
                 .map_err(|e: tokio_threadpool::BlockingError| warp::reject::custom(e)),
             )
+            .boxed()
         };
 
-    let post_filter = warp::post2()
+    let post_filter = warp::post()
```
LegNeato: ditto.

juniper_warp/src/lib.rs
```diff
@@ -396,39 +600,50 @@ pub fn graphiql_filter(
...
+
     #[test]
     fn graphiql_response_does_not_panic() {
         graphiql_response("/abcd");
     }
 
     #[test]
     fn graphiql_endpoint_matches() {
-        let filter = warp::get2()
+        let filter = warp::get()
```
LegNeato: Why these changes?

juniper_warp/src/lib.rs
nWacky: I decided to use wapr's master branch because it supports futures 0.3 and they recently [replaced all `get`, `post`, `put` and `delete` functions with `get2`, `post2`, `put2` and `delete2`](https://github.com/seanmonstar/warp/commit/b9637567f96c64f1bfe734842718e28a7c633e23)

juniper/src/value/object.rs
```diff
@@ -90,6 +90,27 @@ impl<S> Object<S> {
...
+    /// of an Object.
+    pub fn try_from_iter<I, K>(iter: I) -> Option<Self>
+    where
+        I: IntoIterator<Item = (K, Option<Value<S>>)>,
+        K: Into<String>,
+    {
+        Some(Self {
+            key_value_list: iter
+                .into_iter()
+                .map(|(k, v)| v.map(|v| (k.into(), v)))
```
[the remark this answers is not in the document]
nWacky: Yes. It is used to [collect object of iterators to iterator over objects](https://github.com/instrumentisto/juniper/blob/async-await-subscriptions/juniper/src/http/mod.rs#L322) for default object into iterator implementation (for example, Object(["first_field": Iterator(1, 2, 3), "second_field": Iterator(5, 6)]) to Iterator(Object("first_field": 1, "second_field": 5), Object("first_field": 2, "second_field": 6)) )

I added [`Iterator::take_while`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.take_while) to stop iteration over fields after first `None` is found

juniper/src/value/object.rs
```diff
@@ -62,7 +62,7 @@ impl<S> Object<S> {
     }
 
     /// Get the current number of fields
-    pub fn field_count(&self) -> usize {
+    pub fn fields_count(&self) -> usize {
```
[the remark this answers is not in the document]
nWacky: Reverted changes

juniper_codegen/src/impl_object.rs
```diff
@@ -32,26 +48,27 @@ pub fn build_object(args: TokenStream, body: TokenStream, is_internal: bool) ->
...
+            if !(name == obj_name || name == format!("juniper.{}", obj_name)) {
+                panic!(format!(
+                    "The impl block must implement the '{}' trait",
+                    name
+                ));
             }
         }
         None => {
-            // panic!("The impl block must implement the 'GraphQLObject' trait");
+            //            panic!(format!("The impl block must implement the '{}' trait", name))
```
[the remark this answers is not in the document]
nWacky: It is commented in `async-await` branch as well. I think it is here to use `#[juniper::object]` for both `GraphQLType` and `GraphQLTypeAsync`:

```rust
#[juniper::object(Context = Context)]
impl GraphQLType for Query { 
//   ^--- with this panic user has to specify that he\she is implementing `GraphQLObject` trait  
    fn some_field(...) { ... }
}
```

```rust
#[juniper::object(Context = Context)]
impl Query {
//   ^--- by not specifying type implicitly user can implement both traits with single impl block  
    fn some_field(...) { ... }

    async fn async_field(...) { ... }
}
```

juniper_codegen/src/util.rs
nWacky: Yes, didn't think about it

nWacky:
I won't rebase onto `async-await` for now because warp [does not support](https://github.com/seanmonstar/warp/blob/master/Cargo.toml#L23) stable futures 0.3 yet and juniper does => `warp_subscriptions` example does not compile

LegNeato:
Thanks for the quick turnaround! 🍻 

Soooooo close! 🤞

theduke:
Hey, after a quick initial pass, this is a good start.  I see there is still a lot to do, so I won't go crazy with comments, we can live small things like naming to the end or when you feel ready. 

Just a few notes: 

### Sync Subscriptions

Non-async subscriptions  don't really work with this design. They would require some in-between threading/management layer and I don't really see much value in supporting it anyway. I think we can just drop them and do async only.

### Coordination Layer

With regular queries we could get away with not having some management layer in-between (although this makes writing the integration crates more difficult and we want to introduce one there too).  

But for subscriptions we really need one, otherwise the integration crates will become really messy and cumbersome to maintain. 
Subscriptions are also quite a bit  more stability sensitive than regular queries, they provide a great vector for DOS attacks and can bring down a server easily if not handled right. We really want to try and mitigate that out of the box. 

That layer should be transport agnostic, I think we can do that reasonably well with traits.

Rough ideas, names are just pulled out of the air. Also, we don't need most  of the functionality in a MVP, I've split everything into what I think we need initially and what can come later.

`SubscriptionCoordinator`:

* global coordinator
* contains the schema
* keeps track of subscription connections
* handles subscription start, maintains a global subscription id

*post MVP*:
    - max subscription limits / concurrency limits
    - subscription de-duplication (this can also be done "manually" in the subscription resolvers themselves with a `dataloader` / `mpmc` like approach, but for efficiency we will want something higher up eventually)
    - reconnection on connection loss / buffering / re-synchronisation


`SubscriptionConnection`

* handles a single connection
* processes subscribe, unsubscribe
* close/shutdown: unregister from coordinator

post MVP:
* connection-local + global de-duplication, talk to coordinator
* concurrency limits
* machinery with coordinator to allow reconnection

theduke:
Also, as a note: I think I found a nice way to unify the `GraphQLType` to prevent the need for a seperate `TypeAsync`, so this will probably affect the branch a bit.

theduke:
@nWacky just to clarify, all this doesn't need to be done before we merge this branch into `async-await`, we should do that soon.

But I would like to wait until I have the big type system refactoring out of the way because it will be awkward otherwise.

tyranron:
@theduke thanks for the info! To clarify some things:

> Non-async subscriptions don't really work with this design. They would require some in-between threading/management layer and I don't really see much value in supporting it anyway. I think we can just drop them and do async only.

> I think I found a nice way to unify the GraphQLType to prevent the need for a seperate TypeAsync, so this will probably affect the branch a bit.

I'm very-very agree and happy with that. Even more, having `async` landed in stable we don't need sync resolving at all (I mean having one async `request.execute()` is more than enough). Who really needs sync execution may do it with in-place single thread executor, or we can even provide such capabilities out-of-the-box, like other libraries do (for example, `postgres` which now is a wrapper over `tokio-postgres`).

> Coordination Layer

Hmm... while I don't see _so much_ sense in the described, however, I _clearly see_ how it will simplify the integration and implementing integration crates. One concern that troubles me is to make this layer flexible enough for custom tweaks.  
For example, in our codebase where we use Juniper we had to bake our own integration crate, because existing ones don't provide an ability to propagate custom status codes and headers up from resolvers code (of course, we implement this via errors and context, but HTTP request handlers has no notion about what is contained in that context).

> But I would like to wait until I have the big type system refactoring out of the way because it will be awkward otherwise. (hopefully by tomorrow evening)

We, at the moment, have some troubles with subscription resolving and executor to tackle with and to polish (there is some self-referential stuff occurs). And, also, `#[juniper::subscription]` macro ergonomics are far from ideal. This will take some time.  
So, we are totally OK to wait for the big refactoring, and than update this branch in accordance to it.

teh-cmc:
> For example, in our codebase where we use Juniper we had to bake our own integration crate, because existing ones don't provide an ability to propagate custom status codes and headers up from resolvers code.

Anecdotal evidence: we're facing the same issue.
(Btw thanks for the all the work everyone, Juniper is such a well-designed and nice to use lib; love it.)

theduke:
>  One concern that troubles me is to make this layer flexible enough for custom tweaks.

I have a good idea how this can look so it remains flexible. 

I'll hopefully have some time once the general async migrations are done.

nWacky:
What's new in this push:

- `Executor` is now used to execute both subscriptions and queries/mutations. In streams it can be cloned to `OwnedExecutor`
- `Executor` also has two lifetimes now: `pub struct Executor<'r, 'a, CtxT, S = DefaultScalarValue>`. `'r` shows how long references live for and `'a` shows how long inner data lives for. This was made so `OwnedExecutor` can live longer than `Executor`'s reference
- Macros support `Stream`, `Result<Stream, FieldError>`, `Stream<Result<Item, FieldError>>`, `Result<Stream<Result<Item, FieldError>>, FieldError>` and their type aliases
- Almost rebased onto `async-await`: removed `ScalarRefValue`, but didn't update futures because `warp` is used in subscriptions playground

enfipy:
Any updates here so far?

tyranron:
@enfipy unfortunately, no. We've reached the point where we see no reason to invest more time into this until [the major refactoring of `async-await` branch](https://github.com/graphql-rust/juniper/pull/433#issuecomment-553658415) is done (to omit doing the job twice). So, we're blocked at @theduke's work at the moment. Once it's landed we will push hard on finishing this PR, because we're among the ones who are in a huge need of it in out project.

mbccrack:
Any updates here ?

esmevane:
My understanding is that work in progress on #461 needs to land (or stabilize) before work on this branch can resume, @mbccrack

tyranron:
ping @theduke 
Is there anything we can do to help pushing async Juniper forward?  
Regarding [the refactoring](https://github.com/graphql-rust/juniper/pull/433#issuecomment-553658415), maybe we can try to catch it up and finish this blocker once you explain the desired design and the-work-to-be-finished further? We have time and resources to invest into this, and, of course, we need it very much for usage in our codebases and as a solid ground for further Juniper improvements (extended specs, frameworks integrations, etc).

LegNeato:
I landed the current async-await work on `master` so this should be rebased there. Sorry for taking so long!

nWacky:
Rebased onto master

enfipy:
Any news?

nWacky:
@enfipy,

I have rebased it onto master. I'd like @tyranron to take another look at it, there are some moments to discuss (I've marked those as `TODO#433`). This PR can be submitted for a review to maintainers after those resolved, I think.

In the meantime, any feedback would be helpful 🙂

DrYSG:
@nWacky you said **any** feedback would be helpful. Well firstly, I am so happy that that you are working this. I am a bit new to subscriptions, so I don't know if this is helpful but I checked out the main branch today, and when I started cargo I found the server was on port 8080 not 8000 (minor)

```
D:\CRIME\juniper\examples\warp_async>cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 2.28s
     Running `target\debug\warp_async.exe`
[2020-02-21T15:02:22Z INFO  warp_async] Listening on 127.0.0.1:8080
```
Seems to use the GraphiQL interface, not the playground as advertised (minor).

I am using insomnia (not postman) for the queries, which works fine::

```
{"query":"query {\n  users {\n    id\n    kind\n    name\n    friends {\n      id\n      kind\n    }\n  }\n}\n","variables":null}
```

But when I tried the subscribe as you suggested:

```
{"id":"1","type":"start","payload":{"variables":{},"extensions":{},"operationName":null,"query":"subscription { ...myFragment } fragment myFragment on Subscription { users { id name friends { id name } }"}}
```

I got this as the response:

```
Request body deserialize error: data did not match any variant of untagged enum GraphQLBatchRequest

```

everything is being sent as a POST to:

http://localhost:8080/graphql

stoically:
@DrYSG did you try the [PR branch](https://github.com/instrumentisto/juniper/tree/async-await-subscriptions)?

DrYSG:
Sorry, when @nWacky  said he rebased to "master" (four comments above) I thought that was where it was. I am now on the async-await-subscriptions (not ws) branch of: instrumentisto/juniper.

A few observations: (I call them observations, since it could well be that I am still doing things that this branch was not designed for).

**Compile waring:**
```
   Compiling hyper v0.13.2
warning: missing documentation for macro
   --> juniper\src\macros\scalar.rs:347:1
    |
347 | macro_rules! graphql_scalar {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
note: lint level defined here
   --> juniper\src\lib.rs:92:9
    |
92  | #![warn(missing_docs)]
```
**panic**  from the playground the query:
```
query {
  users {
    id
    kind
    name
  }
}
```
caused the server to:

```
     Running `D:\CRIME\juniper\target\debug\warp_subscriptions.exe`
[2020-02-24T15:18:38Z INFO  warp_subscriptions] Listening on 127.0.0.1:8080
thread 'tokio-runtime-worker' panicked at 'Internal error: missing argument id - validation must have failed', src\libcore\option.rs:1188:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
thread 'tokio-runtime-worker' panicked at 'Internal error: missing argument id - validation must have failed', src\libcore\option.rs:1188:5
```

As for the subscriptions, from the above it appears I should not use the playground, but to what address do I use insomnia to send the below query:

a. http://localhost:8000/subscriptions (as the top directions)
b. http://localhost:8080/graphql (which seems to respond, but with an error) (bad request:)
```
Request body deserialize error: data did not match any variant of untagged enum GraphQLBatchRequest
```
subscribe POST request
```
{"id":"1","type":"start","payload":{"variables":{},"extensions":{},"operationName":null,"query":"subscription { ...myFragment } fragment myFragment on Subscription { users { id name friends { id name } }"}}
``

**Comment by stoically** [comment]
@nWacky subscriptions flowing in as described, great work!

@DrYSG playground is running on `http://localhost:8080/playground`, and you'd have to give the query an id: `query { users(id: 1) ..`

**Comment by DrYSG** [comment]
@stoically , Ok, sorry for the mixup. So the query is working fine for me in Insomia: (note there I send direct to localhost:8080/graphql)

```
> POST http://localhost:8080/graphql HTTP/1.1
> Host: localhost:8080
> Authorization: Basic eXNnNDIwNjp5ZWNoZXprYWw=
> User-Agent: insomnia/7.1.1
> Proxy-Connection: Keep-Alive
> Content-Type: application/json
> Accept: */*
> Content-Length: 105

| {"operationName":null,"variables":{},"query":"{\n  users(id: 1) {\n    id\n    kind\n    name\n  }\n}\n"}

* upload completely sent off: 105 out of 105 bytes

< HTTP/1.1 200 OK
< Content-Length: 59
< Content-Type: application/json
< Date: Tue, 25 Feb 2020 15:23:04 GMT

```

But the subscription is a different problem:

Via playground:

```
subscription {
    users {
        id
        kind
        name
        friends {
            id
            kind
        }
    }
}
```

reply:

```
{
  "error": "Could not connect to websocket endpoint ws://localhost:8080/subscriptions. Please check if the endpoint url is correct."
}
```
and via insomnia:

```
* Preparing request to http://localhost:8080/graphql
* Using libcurl/7.57.0-DEV OpenSSL/1.0.2o zlib/1.2.11 libssh2/1.7.0_DEV
* Current time is 2020-02-25T15:29:58.010Z
* Disable timeout
* Enable automatic URL encoding
* Enable SSL validation
* Enable cookie sending with jar of 3 cookies
* Enable network proxy for http:
* Found bundle for host localhost: 0x1be3b2dfce0 [can pipeline]
* Re-using existing connection! (#0) with proxy localhost
* Connected to localhost (127.0.0.1) port 3128 (#0)

> POST http://localhost:8080/graphql HTTP/1.1
> Host: localhost:8080
> User-Agent: insomnia/7.1.1
> Proxy-Connection: Keep-Alive
> Content-Type: application/json
> Accept: */*
> Content-Length: 206

| {"id":"1","type":"start","payload":{"variables":{},"extensions":{},"operationName":null,"query":"subscription { ...myFragment } fragment myFragment on Subscription { users { id name friends { id name } }"}}

* upload completely sent off: 206 out of 206 bytes

< HTTP/1.1 400 Bad Request
< Content-Length: 99
< Content-Type: text/plain; charset=utf-8
< Date: Tue, 25 Feb 2020 15:29:58 GMT
```

Or as suggested in the directions at the top of this thread:

```
* Preparing request to http://localhost:8080/subscriptions
* Using libcurl/7.57.0-DEV OpenSSL/1.0.2o zlib/1.2.11 libssh2/1.7.0_DEV
* Current time is 2020-02-25T15:32:18.716Z
* Disable timeout
* Enable automatic URL encoding
* Enable SSL validation
* Enable cookie sending with jar of 3 cookies
* Enable network proxy for http:
* Found bundle for host localhost: 0x1be3b2dfce0 [can pipeline]
* Re-using existing connection! (#0) with proxy localhost
* Connected to localhost (127.0.0.1) port 3128 (#0)

> POST http://localhost:8080/subscriptions HTTP/1.1
> Host: localhost:8080
> User-Agent: insomnia/7.1.1
> Proxy-Connection: Keep-Alive
> Content-Type: application/json
> Accept: */*
> Content-Length: 206

| {"id":"1","type":"start","payload":{"variables":{},"extensions":{},"operationName":null,"query":"subscription { ...myFragment } fragment myFragment on Subscription { users { id name friends { id name } }"}}

* upload completely sent off: 206 out of 206 bytes

< HTTP/1.1 405 Method Not Allowed
< Content-Length: 23
< Content-Type: text/plain; charset=utf-8
< Date: Tue, 25 Feb 2020 15:32:18 GMT
```

**Comment by nWacky** [comment]
@DrYSG,

>   "error": "Could not connect to websocket endpoint ws://localhost:8080/subscriptions. Please check if the endpoint url is correct."

It seems like you're using `127.0.0.1:8080/playground` with `localhost:8080/graphql` endpoint url. Try using playground exposed at [localhost](http://localhost:8080/playground).

> and via insomnia:
> | {"id":"1","type":"start","payload":{"variables":{},"extensions":{},"operationName":null,"query":"subscription { ...myFragment } fragment myFragment on Subscription { users { id name friends { id name } }"}}
> Upload completely sent off: 206 out of 206 bytes
HTTP/1.1 400 Bad Request

There's an unexpected end of input at line 1, column 107 in the test query. Try sending this:
```
{"id":"1","type":"start","payload":{"variables":{},"extensions":{},"operationName":null,"query":"subscription { ...myFragment } fragment myFragment on Subscription { users { id name friends { id name } } }"}}
```

> HTTP/1.1 405 Method Not Allowed

This is demo server only implementation detail: it does not respond anyhow to `OPTIONS` => CORS requests are not supported. Try sending requests from the playground.

**Comment by DrYSG** [comment]
This is what I tried in the playground (server is printing out: message is error: Err(AlreadyClosed) in a loop.

A picture of the playground might capture what I am doing better (from http://localhost/playground)

image

**Comment by DrYSG** [comment]
I fixed the request via insomnia, but I still get the Method not allowed. I don't have a handy WS client. (Do you know of one?) because you are probably right, it might well be a CORS issue:

```
* Preparing request to http://localhost:8080/subscriptions
* Using libcurl/7.57.0-DEV OpenSSL/1.0.2o zlib/1.2.11 libssh2/1.7.0_DEV
* Current time is 2020-02-26T21:00:28.225Z
* Disable timeout
* Enable automatic URL encoding
* Enable SSL validation
* Enable cookie sending with jar of 3 cookies
* Enable network proxy for http:
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 3128 (#0)

> POST http://localhost:8080/subscriptions HTTP/1.1
> Host: localhost:8080
> User-Agent: insomnia/7.1.1
> Proxy-Connection: Keep-Alive
> Content-Type: application/json
> Accept: */*
> Content-Length: 208

| {"id":"1","type":"start","payload":{"variables":{},"extensions":{},"operationName":null,"query":"subscription { ...myFragment } fragment myFragment on Subscription { users { id name friends { id name } } }"}}

* upload completely sent off: 208 out of 208 bytes

< HTTP/1.1 405 Method Not Allowed
< Content-Length: 23
< Content-Type: text/plain; charset=utf-8
< Date: Wed, 26 Feb 2020 21:00:28 GMT


* Received 23 B chunk
* Connection #0 to host localhost left intact
```

**Comment by DrYSG** [comment]
I did just now try this WebSocket Chrome Extension: https://chrome.google.com/webstore/detail/websocket-test-client/fgponpodhbmadfljofbimhhlengambbn 

and tell me if it is working right?

image

**Comment by geropl** [comment]
Hi :wave: !

I'm new to `juniper` but would like to help out with this effort (if possible, I'd call my Rust "intermediate").

@nWacky Anything specifically that one could test/add testcases for/do otherwise to help out?

Thanks!

**Comment by nWacky** [comment]
@geropl,

Hi 👋️.

I've pushed changes I've been working on recently. 

The main change is [these features implementation](https://github.com/graphql-rust/juniper/pull/433#issuecomment-553570544). Any feedback on this would be helpful.

For now I'm planning to add some more tests and clean up the code a little more. I'll post details on what changed in the description later today.

I've also added `juniper_subscriptions` crate. It helps to resolve `Stream<Item = Result<Value<S>, ExecutionError<S>>>` to `Stream<Item = GraphQLResponse<'a, S>>`. `GraphQLResponse` can be deserialized and sent to the user. Any feedback on that would be helpful as well. 

I don't know about PRs to `juniper_subscriptions`, though. It might be heavily refactored or its functionality might be moved somewhere else. 

I'm planning to send this PR to @tyranron to review changes first, then send it to @LegNeato or @theduke for a review. After reviews are completed and this PR is merged it would be more safe to add more functionality to `juniper_subscriptions`, I think. I'm planning to send this PR for a review in the following 2-3 days.

As far as I know, cargo clippy doesn't pass successfully both here and on master. You might want to work on that in the meanwhile

**Comment by DrYSG** [comment]
what does " I've also added juniper_subscriptions crate" mean?

**Comment by LegNeato** [comment]
@DrYSG it is not in crates.io yet...it is just in this PR (which has yet to land but it is so close!)

**Comment by nWacky** [comment]
Cleaned up code, updated docs, formatted and tested it. Updated description.

Everything should be ready for a pre-review now

File: juniper_subscriptions/src/lib.rs [hunk 14]
```diff
@@ -0,0 +1,446 @@
...
+
+            let stream = futures::stream::poll_fn(
+                move |mut ctx| -> Poll<Option<GraphQLResponse<'static, S>>> {
+                    let mut obj_iterator = object.iter_mut();
+
+                    // Due to having to modify `ready_vec` contents (by-move pattern)
+                    // and only being able to iterate over `object`'s mutable references (by-ref pattern)
+                    // `ready_vec` and `object` cannot be iterated simultaneously.
+                    // TODO: iterate over i and (ref field_name, ref val) once
+                    //       [this RFC](https://github.com/rust-lang/rust/issues/68354)
```
LegNeato: Please file an issue for this.

juniper_warp/Cargo.toml
```diff
@@ -8,21 +8,19 @@ documentation = "https://docs.rs/juniper_warp"
...
 [dependencies]
-warp = "0.1.8"
+warp = "0.2"
 juniper = { version = "0.14.2", path = "../juniper", default-features = false  }
 serde_json = "1.0.24"
 serde_derive = "1.0.75"
-failure = "0.1.2"
+failure = "0.1.7"
 # TODO: rebase juniper_warp to futures 3 once warp supports it
 futures = "0.1.29"
```
LegNeato: Because warp 2.0 uses futures, all the futures-0.1 stuff should be removed. We should be able to remove the old futures and rename futures03

juniper_warp/Cargo.toml
```diff
@@ -8,21 +8,19 @@ documentation = "https://docs.rs/juniper_warp"
...
 serde = "1.0.75"
-tokio-threadpool = "0.1.7"
+tokio = { version = "0.2", features = ["blocking", "fs", "stream", "sync", "time"] }
+tokio-threadpool = "0.1.16"
 
-futures03 = { version = "=0.3.1", optional = true, package = "futures", features = ["compat"] }
+futures03 = { version = "0.3.1", package = "futures", features = ["compat"] }
```
LegNeato: We don't need compat anymore. Also rename this to futures after.

juniper_warp/src/lib.rs
```diff
@@ -40,15 +40,13 @@ Check the LICENSE file for details.
...
-use serde::Deserialize;
-use std::sync::Arc;
-use warp::{filters::BoxedFilter, Filter};
-
-#[cfg(feature = "async")]
-use futures03::future::{FutureExt, TryFutureExt};
+use std::{pin::Pin, sync::Arc};
 
+use futures::future::poll_fn;
+use futures03::future::{Future, FutureExt as _};
```
LegNeato: rename

juniper_warp/src/lib.rs
```diff
@@ -40,15 +40,13 @@ Check the LICENSE file for details.
...
-use futures::{future::poll_fn, Future};
-use serde::Deserialize;
-use std::sync::Arc;
-use warp::{filters::BoxedFilter, Filter};
-
-#[cfg(feature = "async")]
-use futures03::future::{FutureExt, TryFutureExt};
+use std::{pin::Pin, sync::Arc};
 
+use futures::future::poll_fn;
```
LegNeato: This looks the same in newer futures.

juniper_warp/src/lib.rs
```diff
@@ -199,39 +200,45 @@ where
...
 ) -> BoxedFilter<(warp::http::Response<Vec<u8>>,)>
 where
     S: ScalarValue + Send + Sync + 'static,
-    Context: Send + 'static,
+    Context: Send + Sync + 'static,
     Query: juniper::GraphQLType<S, Context = Context, TypeInfo = ()> + Send + Sync + 'static,
     Mutation: juniper::GraphQLType<S, Context = Context, TypeInfo = ()> + Send + Sync + 'static,
+    Subscription: juniper::GraphQLType<S, Context = Context, TypeInfo = ()> + Send + Sync + 'static,
 {
+    use futures::future::Future as _;
```
LegNeato: remove

juniper_warp/src/lib.rs
```diff
@@ -199,39 +200,45 @@ where
...
+
     let schema = Arc::new(schema);
     let post_schema = schema.clone();
 
     let handle_post_request =
         move |context: Context, request: GraphQLBatchRequest<S>| -> Response {
             let schema = post_schema.clone();
-            Box::new(
+
+            futures03::compat::Compat01As03::new(
```
LegNeato: remove compat

juniper_warp/src/lib.rs
```diff
@@ -240,7 +247,7 @@ where
                                    mut request: std::collections::HashMap<String, String>|
           -> Response {
         let schema = schema.clone();
-        Box::new(
+        futures03::compat::Compat01As03::new(
```
LegNeato: remove compat

juniper_warp/src/lib.rs
```diff
@@ -256,27 +263,29 @@ where
                         variables,
                     );
 
-                    let response = graphql_request.execute(&schema, &context);
+                    let response = graphql_request.execute_sync(&schema, &context);
```
LegNeato: I don't know about warp filters, should this be sync?

juniper_warp/src/lib.rs
```diff
@@ -199,39 +200,45 @@ where
...
 ///
 /// let graphql_endpoint = warp::path("graphql")
-///     .and(warp::post2())
+///     .and(warp::post())
 ///     .and(graphql_filter);
-/// # }
 /// ```
-pub fn make_graphql_filter<Query, Mutation, Context, S>(
-    schema: juniper::RootNode<'static, Query, Mutation, S>,
+pub fn make_graphql_filter<Query, Mutation, Subscription, Context, S>(
```
LegNeato: This should have a `_sync` suffix and the other ones with `_async` need it deleted.

Cargo.toml
```diff
@@ -9,11 +9,15 @@ members = [
   "juniper_hyper",
   "juniper_iron",
   "juniper_rocket",
+  "juniper_subscriptions",
   "juniper_warp",
 ]
 exclude = [
   "docs/book/tests",
+  # TODO enable warp
+  "juniper_warp",
```
LegNeato: This should no longer be excluded.

examples/warp_subscriptions/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+[dependencies]
+env_logger = "0.6.2"
+futures = { version = "=0.3.1" }
+log = "0.4.8"
+serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0"
+tokio = { version = "0.2", features = ["rt-core", "macros"] }
+warp = "0.2.1"
+
+# TODO#433: get crates from GitHub
```
LegNeato: File an issue once this lands please.

examples/warp_subscriptions/src/warp_subscriptions.rs
```diff
@@ -0,0 +1,181 @@
+//! `juniper_warp` subscriptions handler implementation.
```
LegNeato: This might be better in `warp` behind  an `unstable_` or `non-standard_` flag?

juniper_benchmarks/src/lib.rs
```diff
@@ -95,20 +96,25 @@ pub struct Mutation;
...
 
-pub async fn execute_async(query: &str, vars: Variables) -> QueryResult {
+pub async fn execute(query: &str, vars: Variables) -> QueryResult {
     let root = new_schema();
     let ctx = Context::new();
-    juniper::execute_async(query, None, &root, &vars, &ctx)
+    juniper::execute(query, None, &root, &vars, &ctx)
         .await
         .map_err(|e| format!("{:?}", e))
 }
```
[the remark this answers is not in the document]
LegNeato: Ah, I see you said that will come later.

juniper_subscriptions/src/lib.rs
```diff
@@ -0,0 +1,446 @@
+//! This crate supplies [`SubscriptionCoordinator`] and
```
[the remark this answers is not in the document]
nWacky: This is an interesting question. 

I was thinking about providing a coordinator and a subscription type, which will cover most of the use cases, but allow users to implement their own coordinators and connections if needed. 

`Value<Stream<Item = Value>>` is returned from Juniper's resolvers, but it would be better to iterate over `GraphQLResponse`s in a server library (+ have some functionality on top of this stream). Each Object/List is assembled from items, which are returned from streams. One might want to wait until all values from `Object`/`List` are yielded and then return the complete thing, while other might want to return values as soon as they arrive. Some people might not mind returning incomplete objects, as long as they have some required fields filled up. This is the main reason for moving `SubscriptionConnection` into a separate trait.

I didn't implement a lot in `SubscriptionCoordinator` yet, but I think it should be in a separate trait as well. Some users might want to add a custom health checker to a connection, others might want to decline another connection if CPU usage is greater than 80%.

These cases will not be covered by `juniper_subscriptions`, but I was planning to cover [those cases](https://github.com/graphql-rust/juniper/pull/433#issuecomment-553570544) for now

juniper_subscriptions/src/lib.rs
```diff
@@ -0,0 +1,446 @@
...
+use juniper::{
+    http::{GraphQLRequest, GraphQLResponse},
+    BoxFuture, ExecutionError, GraphQLError, GraphQLSubscriptionType, GraphQLTypeAsync, Object,
+    ScalarValue, SubscriptionConnection, SubscriptionCoordinator, Value, ValuesStream,
+};
+
+/// Simple [`SubscriptionCoordinator`] implementation:
+/// - contains the schema
+/// - handles subscription start
+pub struct Coordinator<'a, QueryT, MutationT, SubscriptionT, CtxT, S>
```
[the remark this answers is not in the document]
nWacky: I like the `SimpleCoordinator` name, but I'm planning to add [some of those](https://github.com/graphql-rust/juniper/pull/433#issuecomment-553570544) things in the upcoming PRs.

juniper_subscriptions/src/lib.rs
```diff
@@ -0,0 +1,446 @@
...
+    MutationT: GraphQLTypeAsync<S, Context = CtxT> + Send + Sync,
+    MutationT::TypeInfo: Send + Sync,
+    SubscriptionT: GraphQLSubscriptionType<S, Context = CtxT> + Send + Sync,
+    SubscriptionT::TypeInfo: Send + Sync,
+    CtxT: Send + Sync,
+{
+    root_node: juniper::RootNode<'a, QueryT, MutationT, SubscriptionT, S>,
+}
+
+impl<'a, QueryT, MutationT, SubscriptionT, CtxT, S>
```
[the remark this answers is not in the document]
nWacky: This is a local `Coordinator` implementation. It requires `<QueryT, MutationT, SubscriptionT, CtxT, S>` to be able to hold `RootNode`.

Users can avoid implementing `SubscriptionCoordinator` if they'll depend on `juniper_subscriptions::Coordinator`

juniper_subscriptions/src/lib.rs
```diff
@@ -0,0 +1,446 @@
...
+    S: ScalarValue + Send + Sync + 'a,
+{
+    type Item = GraphQLResponse<'a, S>;
+
+    fn poll_next(
+        self: Pin<&mut Self>,
+        cx: &mut futures::task::Context<'_>,
+    ) -> Poll<Option<Self::Item>> {
+        // this is safe as stream is only mutated here and is not moved anywhere
+        let Connection { stream } = unsafe { self.get_unchecked_mut() };
```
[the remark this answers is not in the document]
nWacky: I couldn't implement this without `unsafe` as `poll_next` needs `Pin<&mut Self>` and I couldn't get a mutable reference for the underlying field, so I did it similar to [`Select` implementation](https://docs.rs/futures-util/0.3.4/src/futures_util/stream/select.rs.html#94-97). 

Any suggestions on how to do this without `unsafe` would be helpful

juniper_warp/src/lib.rs
nWacky: [tokio_threadpool](https://crates.io/crates/tokio-threadpool) does not support futures 0.3 yet. 

I would suggest leaving it as is in sync resolvers for now

examples/warp_subscriptions/src/warp_subscriptions.rs
nWacky: I'll add it under `subscriptions` feature flag to `juniper_warp` for now

juniper_warp/src/lib.rs
LegNeato: From the docs:

"Note: This crate is deprecated in tokio 0.2.x and has been moved and refactored into various places in the tokio::runtime module of the tokio crate. Note that there is no longer a ThreadPool type, you are instead encouraged to make use of the thread pool used by a Runtime configured to use the threaded scheduler."

Doesn't look like it will be ported. I really don't want futures 01 in here anymore.

juniper_subscriptions/src/lib.rs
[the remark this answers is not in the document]
LegNeato: Yes, but if we expect people to be implementing these things it might be better to create a proc macro in the future. Similar to how juniper's macros abstract out the juniper boilerplate even though you can still implement the boilerplate without the macros directly.

LegNeato:
To fix the release stuff you probably need to add `juniper_subscriptions` to https://github.com/instrumentisto/juniper/blob/67ba411e4436d1bd327bab5b7765497244511477/juniper/release.toml#L7

LegNeato:
I'm going to land this and we can iterate on master, including `TODO#433`.

🍻  😍   🍾 🎉 👍 💯 🥇 ❤️ **THANK YOU SO MUCH FOR THIS PR** 🍻  😍   🍾 🎉 👍 💯 🥇 ❤️ 

It was a ton of work and will be moving the ecosystem forward.

PR #438, opened 2019-10-10
Remove feature(async_closure)

tyranron:
## Synopsis

This PR removes the necessity of `#![feature(async_closure)]` usage making `juniper` and `juniper_codegen` usable on current `beta` Rust (and so on `stable` when 1.39 Rust will land).

The `#![feature(async_closure)]` was used only in one place in `juniper_codegen` for being able to annotate the return type of resolver code gathered by `#[juniper::object]` macro. 

## Solution

This PR moves resolved type annotation from the closure signature to the left side of generated `let` binding.  
So, instead of:
```rust
let f = async move || -> #_type {
    #resolver_code
}()
```
It generates now:
```rust
let f: #_type = async move {
    #resolver_code
}
```

The generated code for sync resolvers is affected too in the same manner, by replacing immediately invoked closure with a regular block.

## Checked

- [x] `integration_tests/juniper_tests` passes
- [x] `integration_tests/async_await` passes

## Additionally

- Fixes some inconsistencies appeared in a `juniper` crate after [latest merging with `master`](https://github.com/graphql-rust/juniper/commit/2884ef313bf73cf307d31924df4b6f32b6905168).

### DIFF
--- a/examples/warp_async/Cargo.toml
+++ b/examples/warp_async/Cargo.toml
@@ -10,7 +10,7 @@ edition = "2018"
 log = "0.4.8"
 env_logger = "0.6.2"
 warp = "0.1.19"
-futures-preview = { version = "0.3.0-alpha.18", features = ["nightly", "async-await", "compat"] }
+futures-preview = { version = "0.3.0-alpha.19", features = ["async-await", "compat"] }
 reqwest = "0.9.19"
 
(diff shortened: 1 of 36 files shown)


LegNeato:
Whoops, sorry about the bad merge!

theduke:
Most good ideas are obvious, in *hindsight*. ;)

PR #449, opened 2019-11-01
Fix name resolving error in `#[juniper::object]`

nWacky:
### Overview
Due to changes in name resolving logic in `#[juniper::object]` macro name attribute is not resolved correctly. This PR reverts name resolving logic to the one present in `master` branch.

### More detailed description
When running certain queries (for example, introspection query) some types are not found by juniper, though they should exist. Consider the following example:
```graphql
query {
  __type(name: "User") {
    ...FullType
  }
}

fragment FullType on __Type {
  kind
  name
  description
}
```

Server returns `"Unknown type \"__Type\"` error. High-level resolving logic:
```
got request ->
started validation ->
    query has correct syntax ->
    type "__Type" not found ->
-> return error
```

Type `__Type` is declared in `TypeType` impl block in `src/schema/schema.rs`:
```rust
#[crate::object_internal(
    name = "__Type"   // <-- type name for validator declared here
    Context = SchemaType<'a, S>,
    Scalar = S,
    // FIXME: make this redundant.
    noasync,
)]
impl<'a, S> TypeType<'a, S>
where
    S: crate::ScalarValue + 'a,
{ ... }
```
But in macro code looks at "name" field only if it can't find type name itself:
```rust
 let name = if let Some(ident) = util::name_of_type(&*_impl.self_ty) {
        ident
    } else {
        panic!("Could not determine a name for the object type: specify one with #[juniper::object(name = \"SomeName\")");
    };
```

### DIFF
--- a/juniper_codegen/src/impl_object.rs
+++ b/juniper_codegen/src/impl_object.rs
@@ -41,29 +41,18 @@ pub fn build_object(args: TokenStream, body: TokenStream, is_internal: bool) -> 
         }
     }
 
-    let name = match impl_attrs.name.as_ref() {
-        Some(type_name) => type_name.clone(),
-        None => {
-            let error_msg = "Could not determine a name for the object type: specify one with #[juniper::object(name = \"SomeName\")";
 
-            let path = match &*_impl.self_ty {
-                syn::Type::Path(ref type_path) => &type_path.path,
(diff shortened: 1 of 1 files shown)


theduke:
Whoops, thanks.

Note that the `name_of_type` function does the right thing, so please use that function after checking the attribute.

theduke:
Thanks!

PR #450, opened 2019-11-01
Handle raw identifiers

davidpdrsn:
If you have a schema like

```rust
#[juniper::object]
impl Query {
    fn r#type(r#fn: MyInputType) -> Vec<String> {
        unimplemented!()
    }
}

#[derive(GraphQLInputObject, Debug, PartialEq)]
struct MyInputType {
    r#trait: String,
}
```

The names in the GraphQL schema would still include `r#`. This patch fixes that for field names on fields of types, fields of input types, and arguments. Any other places where this should be fixed?

I discovered this while working on https://github.com/davidpdrsn/juniper-from-schema/pull/93.

### DIFF
--- a/juniper_codegen/src/util.rs
+++ b/juniper_codegen/src/util.rs
@@ -593,6 +593,11 @@ pub struct GraphQLTypeDefinitionField {
     pub resolver_code: proc_macro2::TokenStream,
 }
 
+pub fn unraw(s: &str) -> String {
+    use syn::ext::IdentExt;
+    quote::format_ident!("{}", s).unraw().to_string()
+}
+
 /// Definition of a graphql type based on information extracted
 /// by various macros.
 /// The definition can be rendered to Rust code.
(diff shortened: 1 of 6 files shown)


juniper_codegen/src/util.rs
```diff
@@ -593,6 +593,11 @@ pub struct GraphQLTypeDefinitionField {
     pub resolver_code: proc_macro2::TokenStream,
 }
 
+pub fn unraw(s: &str) -> String {
```
theduke: I don't think we really need this helper.

`arg.name.unraw().to_string()` would be the same.

davidpdrsn: [`unraw`](https://docs.rs/syn/1.0.7/syn/ext/trait.IdentExt.html) only works on `Ident` and the names here are `String`s. So it would become `format_ident!("{}", field.name).unraw().to_string();` which is a bit wordy. I don't mind changing it if you still think the helper is unnecessary.

davidpdrsn:
Docs updated in b93fa61e9718a

juniper_codegen/src/util.rs
theduke: Ah, I missed that it was taking a `&str`. I think we will want some general refactoring of the macro code to keep the `Ident`s around to provide better error messages.

But we can do that in future work.

theduke:
Squashed and merged, thanks!

PR #441, opened 2019-10-21
Make proc macro support parameter attributes

davidpdrsn:
Fixes #421.

I decided to add https://crates.io/crates/proc-macro-error as mentioned [here](https://github.com/graphql-rust/juniper/issues/435). It made returning a nice error significantly easier.

## TODO

- [x] Fix use of old style in other tests/examples. `cargo test` currently produces quite a few deprecation warnings.
- [x] Make `rename = "..."` work. @theduke not quite sure how to handle this one. Got a hint?
- [ ] Look into allowing docs on arguments with normal attribute https://github.com/graphql-rust/juniper/issues/525

PR #452, opened 2019-11-04
Resolve some todos in async-await branch

nWacky:
This PR resolves some todos in `async-await` branch.

Changelog:
- Resolve RFC 2565 related todos
- Remove `__juniper_extract_generic` macro
- Add `resolve_into_type_async` to `GraphQLTypeAsync` for asynchronous fragments resolving
- Update `GraphQLTypeAsync` with `async-trait`, update macros
- Resolve "better error message with field/type name" todo, update other panics with query name (query name retrieved by calling `self::name(info)`)
- Rebase onto master 
- Format


examples/warp_async/Cargo.toml
```diff
@@ -13,7 +13,10 @@ warp = "0.1.19"
 futures-preview = { version = "0.3.0-alpha.19", features = ["async-await", "compat"] }
 reqwest = "0.9.19"
 
-juniper_codegen = { git = "https://github.com/graphql-rust/juniper", branch = "async-await", features = ["async"] }
-juniper = { git = "https://github.com/graphql-rust/juniper", branch = "async-await", features = ["async"] }
-juniper_warp = { git = "https://github.com/graphql-rust/juniper", branch = "async-await", features = ["async"] }
+#juniper_codegen = { git = "https://github.com/graphql-rust/juniper", branch = "async-await", features = ["async"] }
+#juniper = { git = "https://github.com/graphql-rust/juniper", branch = "async-await", features = ["async"] }
+#juniper_warp = { git = "https://github.com/graphql-rust/juniper", branch = "async-await", features = ["async"] }
```
LegNeato: Nit: Remove these

juniper/src/macros/tests/args.rs
```diff
@@ -73,13 +73,12 @@ impl Root {
...
     }
 
-    // TODO: enable once [RFC 2565](https://github.com/rust-lang/rust/issues/60406) is implemented
-    // fn attr_arg_descr(#[doc = "The arg"] arg: i32) -> i32 { 0 }
-    // fn attr_arg_descr_collapse(
-    //     #[doc = "The arg"]
-    //     #[doc = "and more details"]
-    //     arg: i32,
-    // ) -> i32 { 0 }
+    fn attr_arg_descr(arg: i32) -> i32 {
```
LegNeato: This looks wrong? Also, we have https://github.com/graphql-rust/juniper/pull/441 outstanding

juniper/src/macros/tests/union.rs
```diff
@@ -36,6 +44,16 @@ enum WithGenerics<T> {
...
+enum ResolversFirst {
+    Concrete(Concrete),
+}
+
+enum CommasWithTrailing {
+    Concrete(Concrete),
+}
+enum ResolversWithTrailingComma {
+    Concrete(Concrete),
+}
```
LegNeato: Why were these added back?

juniper/src/macros/tests/args.rs
nWacky: Got it. 
I put it back to todos and updated a link to #441

juniper/src/macros/tests/union.rs
nWacky: Oops, my mistake

juniper/src/macros/tests/field.rs
```diff
@@ -95,13 +95,16 @@ impl Root {
         Ok(0)
     }
 
+    /*
+     * FIXME: make this work again
```
LegNeato: This looks like you need to rebase or a bad merge?

juniper/src/macros/tests/union.rs
LegNeato: I don't see this fixed? Is this a bad merge?

nWacky:
Yes, git messed everything up while merging.
Now everything should be good

theduke:
I'm sorry I didn't see this before, but I'm not actually on board with using `async-trait`. 

It's a nice library, but it just adds a bit of syntax sugar. Which isn't really needed when 99% of the code will be auto-generated by the macros.

theduke:
Could you handle reverting, @nWacky ?

nWacky:
@theduke 

Yes, I'll remove async-trait today. Juniper's macros don't depend on async-trait, I wrote all lifetimes by hand

theduke:
Thanks. I think we will want to merge the async branch in the next few days to avoid extra churn.

I also have some bigger refactorings upcoming that could affect you, we might want to chat about that somewhere.

nWacky:
Yeah, that would be great

LegNeato:
@theduke I converted a project at work to async_trait and it made the implementations much more readable as everything didn't have to have `Box` and `Pin` everywhere. It also let me clean up some `futures::future::Either` use as well. But I guess I don't feel strongly enough to keep it.

theduke:
I have a type system change which gets rid of `GraphlQLTypeAsync` that makes this impossible. 

Should be ready today.

PR #457, opened 2019-11-12
Use futures 0.3.1 in async-await

nWacky:
Rebase `async-await` onto futures 0.3.1:

- Use futures 0.3.1 in all packages except `juniper_warp` (warp supports futures 0.3 only on git)
- Replace all `futures::future::BoxFuture` with `juniper::BoxFuture` in macros


theduke:
Thanks again.

I already had done this myself, but the merge should be trivial so I'm merging this.

PR #442, opened 2019-10-23
Accept Content-Type: application/graphql in web server integrations

sd2k:
When finished, this should fix #426.

Hyper, Iron and Rocket integrations are working fine with tests. I can't for the life of me figure out how to get the warp integration working - any guidance there would be much appreciated!

PR #459, opened 2019-11-13
Add integration for actix_web

ericmarkmartin:
I tried to follow the contributing guidelines best I could but couldn't find anything about what to write for the `Makefile.toml` and `release.toml` files, so I was hoping that could be discussed here. Also, I figured I should wait until merging this before publishing the crate to crates.io, but if there's a different procedure for that as well, please let me know.

Also, with respect to #442, I have already implemented the functionality necessary to accept application/graphql content-typed requests.


docs/book/content/servers/index.md
```diff
@@ -1,17 +1,10 @@
-# Adding A Server
-
-To allow using Juniper with the HTTP server of your choice,
```
theduke: Why did you remove this? It remains accurate.

juniper_actix_web/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+[dev-dependencies]
+actix-rt = "0.2.0"
+pretty_env_logger = "0.2"
+awc = "0.2.8"
+url = "2.1.0"
+
+[dev-dependencies.juniper]
+version = "0.14.1"
+features = ["expose-test-schema", "serde_json"]
+path = "../juniper"
```
theduke: Missing a newline. ;)

juniper_actix_web/src/lib.rs
```diff
@@ -0,0 +1,620 @@
...
+        Err(de::Error::invalid_length(0, &"a positive integer"))
+    } else {
+        Ok(v)
+    }
+}
+
+#[derive(Debug, serde_derive::Deserialize, PartialEq)]
+#[serde(untagged)]
+#[serde(bound = "InputValue<S>: Deserialize<'de>")]
+enum GraphQLBatchRequest<S = DefaultScalarValue>
```
theduke: We really need to factor out the batch request into the main crate, right now we are just duplicating it everywhere. 

(You would be very welcome to do this if interested, but just a suggestion. We can do this as a follow up. )

juniper_actix_web/src/lib.rs
```diff
@@ -0,0 +1,620 @@
...
+
+        Ok(HttpResponse::Ok()
+            .status(status)
+            .content_type("application/json")
+            .body(body))
+    }
+}
+
+#[cfg(test)]
+mod fromrequest_tests {
```
theduke: Since you wrote quite a lot of tests ( :+1: ), let's factor them out to a separate file.

theduke:
> couldn't find anything about what to write for the `Makefile.toml`

We don't need a `Makefile.toml` if no customizations are required. It works automatically.

> I figured I should wait until merging this before publishing the crate to crates.io

If this is a official integration, the crate will have to be owned and published by us.

docs/book/content/servers/index.md
ericmarkmartin: I apologize, I didn't mean to do that

juniper_actix_web/src/lib.rs
ericmarkmartin: Good to know—should I leave this in here for now then?

docs/book/content/servers/index.md
ericmarkmartin: Sorry, just a mistake

juniper_actix_web/src/lib.rs
ericmarkmartin: Good to know—once I wrap this up I'd definitely be interested in taking a look

docs/book/content/servers/index.md
theduke: No need to apologize, that's what reviews are for.

theduke:
@ericmarkmartin thanks for working on this btw.

If you need help/advice anywhere just say so.

ericmarkmartin:
@theduke is this PR blocked on #461? Also, is it necessary that I test the graphiql and playground functions?

ericmarkmartin:
@theduke is there anything that needs to be fixed for this pr to be merged into the `async-await` branch?

adwhit:
This looks really useful, but am concerned it will become immediately obsolete as appears to be based on `actix-web` v1.0? v2.0 has since been released which is wholly based off `std::future::Future`/`async/await`. Haven't looked too closely at this PR but the migrations I have done have tended to significantly simplify the code.

ericmarkmartin:
@adwhit—I’m currently working on actix 2.0, but I’m not sure whether I should create a separate integration or not, as some integrations have taken this route while other have simply opted to replace theirs. Hence why this PR is still up.

LegNeato:
Thanks again for the PR! I believe we landed actix 2.0 support. Do you want to still land 1.0 support? Is there anything missing from the current landed implementation that this has and should be ported there?

ericmarkmartin:
I don't see any reason to land 1.0 given the presence of 2.0, but I will definitely see if there's anything missing from the 2.0 integration (or in its tests) that can be taken from here. Closing.

Issue #455, opened 2019-11-10
Incorrect `ValidationError` for non-executed operations in multiple-operations documents


cairomassimo:
**Describe the bug**
Calling `juniper::execute` fails with a `ValidationError` when a multiple-operation document is given and the provided variables do not validate for *any* of the operations defined in the document, even if they are *not* executed.

Use Case: multiple-operations documents ared used to generate client code using `graphql-client generate` (they are especially convenient for many small-size operations). In this case, the whole original document is sent every time an operation is performed.

**To Reproduce**

Running the following code fails with `ValidationError([RuleError { locations: [SourcePosition { index: 22, line: 0, col: 22 }], message: "Variable \"$x\" of required type \"String!\" was not provided." }])', src/libcore/result.rs:1084:5`:
```rust
extern crate juniper;

use juniper::RootNode;
use std::collections::HashMap;

struct Query;
struct Mutation;

#[juniper::object]
impl Query {
    fn y(x: Option<String>) -> i32 { 0 }
}

#[juniper::object]
impl Mutation {}

type Schema = RootNode<'static, Query, Mutation>;

fn main() {
    juniper::execute(
        "query A { y } query B($x: String!) { y(x: $x) }",
        Some("A"),
        &Schema::new(Query, Mutation),
        &HashMap::new(),
        &(),
    ).unwrap();
}
```

**Expected behavior**
It should run the operation `A` successfully, as it does not define the variable `$x`.

Per:
https://graphql.github.io/graphql-spec/draft/#sec-Executing-Requests
only the variables defined by the executed operation are validated.
> Let coercedVariableValues be the result of CoerceVariableValues(schema, operation, variableValues).

**Additional context**
Version 0.14.1

davidpdrsn:
I'm still new to the codebase but it seems this isn't possible to do currently.

`execute` returns `Result<(Value<S>, Vec<ExecutionError<S>>), GraphQLError<'a>>`, meaning we either return some data or an error, but not both. The [function that validate the input variables](https://github.com/graphql-rust/juniper/blob/master/juniper/src/validation/input_value.rs#L61) returns `GraphQLError` and and since `execute` returns a `Result` it cannot return both `(Value, Vec<ExecutionError>)` and `GraphQLError` which I guess is needed for this case.

cairomassimo:
The issue is with the signature of `validate_input_values`: it should also accept the selected operation.

Say:

```rust
pub fn validate_input_values<S>(
[...]

Issue #427, opened 2019-09-21
Literal Scalar Parsing should not panic


dreamcodez:
**Describe the bug**
Literal Scalar Parsing should not panic. Sending badly shaped data is a thing that Denial-Of-Service attackers like to do -- and being able to panic the thread this easy is a smell in my opinion.

**To Reproduce**
1. Define a schema with a scalar String, Int, or Float type
2. Query the schema with query literals which violate the contract in step (1)
3. Notice that the thread panics
4. Code reference here: https://github.com/graphql-rust/juniper/blob/master/juniper/src/parser/value.rs ... look for `parse_scalar_literal_by_infered_type` and then look for the panics.

**Expected behavior**
The framework should handle this class of error gracefully without a panic. It should return in this case a proper `400 Bad Request` HTTP Response with the standard graphql json error reporting array format: `{ "errors": [ "... must be an Int" ] }`

**Additional context**
Violating the api shape is one of the easiest things a would-be denial-of-service attacker can try to attempt to crash your service and/or waste cpu cycles/compute resources/network resources.

LegNeato:
We should probably add this case to `ParseError` and bubble it up.

andy128k:
@dreamcodez  I am trying to reproduce panic [here](https://github.com/andy128k/juniper/pull/1/files) but I cannot. Could you provide more details?

dreamcodez:
sure. sorry i have been busy shlepping my stuff from place to place.

I am going to create a small repo and post it here. give me an hour or so

dreamcodez:
I added more context to 434, and I have also posted a bounty for this issue (427)

$50 out of my own pocket if someone can just get this one thing fixed, its the only thing which is causing me to not want to use juniper right now :)

https://www.bountysource.com/issues/80699749-literal-scalar-parsing-should-not-panic


Thanks !

dreamcodez:
(and unit tests of course)

theduke:
@andy128k @dreamcodez I tried to reproduce this, but haven't been able to.

All seems to work as expected, without a panic.

Can you please provide a simple reproduction with a sample schema definition and query?

andy128k:
@dreamcodez @theduke Looks like I've reproduced it. `String` was a bad example (any schema has it because it is used in introspection).

But passing Int or Float literal causes panic. So, this is not because of violation of contract but because of absence of scalar type in schema.

theduke:
Can you add a test reproducing the panic?

[...]

PR #462, opened 2019-11-15
Validate variables of the executed operation only

cairomassimo:
Fix for #455 

This PR changes a larger number of files than expected. Here is the rationale.

* We want to validate input variables only for the operation selected for execution. The code that extracts the selected operation was already present in `execute_validated_query`. However, we need the result in `validate_input_values` which is called before `execute_validated_query`.
* So, I extracted the code to a `get_operation` method (the naming mimics the GraphQL spec, see https://graphql.github.io/graphql-spec/June2018/#GetOperation() ), meant to be called directly from `execute`. The resulting Operation can be then passed directly to `validate_input_values` instead of the whole document.
* However, the `get_operation` code originally runs after the validation of the document, but `validate_input_values` is run before this validation. Calling `get_operation` before the validation of the document is dangerous, as it could impact the error reporting for invalid documents (say, multiple operations defined with the same name). I thought that it made much more sense to validate the document first, and _then_ the input variables, so I made this change.
* After this change, I got two failing unit tests. However, I realized that _type_ of the input variables was validated twice (once as part of the document validation, and again during input values validation), and these failing tests were actually depending on which of the redundant validations is applied first.
* After checking that the validation of the _type_ of input variables as part of the document validation is already unit-tested, I decided to remove the tests `does_not_allow_invalid_types_to_be_used_as_values` and `does_not_allow_unknown_types_to_be_used_as_values`.

### DIFF
--- a/juniper/src/executor/mod.rs
+++ b/juniper/src/executor/mod.rs
@@ -4,8 +4,8 @@ use fnv::FnvHashMap;
 
 use crate::{
     ast::{
-        Definition, Document, Fragment, FromInputValue, InputValue, OperationType, Selection,
-        ToInputValue, Type,
+        Definition, Document, Fragment, FromInputValue, InputValue, Operation, OperationType,
+        Selection, ToInputValue, Type,
     },
     parser::SourcePosition,
     value::Value,
@@ -31,6 +31,7 @@ pub use self::look_ahead::{
     Applies, ChildSelection, ConcreteLookAheadSelection, LookAheadArgument, LookAheadMethods,
     LookAheadSelection, LookAheadValue,
 };
+use crate::parser::Spanning;
 
 /// A type registry used to build schemas
 ///
@@ -609,9 +610,9 @@ impl<S> ExecutionError<S> {
     }
 }
 
-pub fn execute_validated_query<'a, QueryT, MutationT, CtxT, S>(
-    document: Document<S>,
-    operation_name: Option<&str>,
+pub fn execute_validated_query<'a, 'b, QueryT, MutationT, CtxT, S>(
+    document: &'b Document<S>,
+    operation: &'b Spanning<Operation<S>>,
     root_node: &RootNode<QueryT, MutationT, S>,
     variables: &Variables<S>,
     context: &CtxT,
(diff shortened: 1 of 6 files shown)


juniper/src/validation/input_value.rs
```diff
@@ -70,13 +66,10 @@ fn validate_var_defs<S>(
...
                 }
             }
-            _ => errors.push(RuleError::new(
-                &format!(
-                    r#"Variable "${}" expected value of type "{}" which cannot be used as an input type."#,
-                    name.item, def.var_type.item,
-                ),
-                &[ name.start ],
-            )),
+            _ => panic!(
```
LegNeato: I believe making this a panic undoes the fix for https://github.com/graphql-rust/juniper/issues/427 ?

juniper/src/executor/mod.rs
```diff
@@ -720,6 +699,41 @@ where
     Ok((value, errors))
 }
 
+pub fn get_operation<'a, 'b, S>(
```
LegNeato: Thanks for pulling out the duplicated code to a function. Can we get a unit test for it?

juniper/src/validation/input_value.rs
cairomassimo: Ok, this was intended as an `unreachable!` instead of `panic!` (I will change it), since now the code is run after the validation of the document so the case should never occur. I will make sure that this is the case, and either find a unit test that checks it, or write a new one.
Thanks

juniper/src/executor/mod.rs
cairomassimo: Ok, I will try to make some tests specific for this function.
However, this is not new functionality so as a whole it should be covered by existing test cases.

LegNeato: Totally. Just before the invariants were tangentially tested due to the implementation around this code. Now that this is a function with discrete inputs and outputs we should have some test that ensure the input and output invariants hold on their own.

cairomassimo: I had a look at the test suite, I found that all the relevant cases for `get_operation` are covered in test module `named_operations` at `juniper/src/executor_tests/executor.rs:1061`.

Here are the test cases:

- `uses_inline_operation_if_no_name_provided`
- `uses_only_named_operation`
- `uses_named_operation_if_name_provided`
- `error_if_multiple_operations_provided_but_no_name`
- `error_if_unknown_operation_name_provided`

I was thinking of having unit test cases for `get_operation` very similar to these above.
However, in my experience, when a thorough testing of a specific feature is possible at the integration-test level without extra effort, I prefer having only integration tests and not unit tests (integration tests are more robust against refactorings, and also test more stuff with less code, etc.).
Since this seems the case here, IMHO adding unit tests here would not add any value, and instead make the code less maintainable.

Let me know your thought about this, and also if you see a better option.

Sorry for the long talk and no code :sweat_smile: 
If you think it is better to have the tests for whatever reason, I can definitely write them! :smiley:

juniper/src/validation/input_value.rs
cairomassimo: These test cases make sure that variable types are valid.

- `unknown_type_names_are_invalid` at `juniper/src/validation/rules/known_type_names.rs:90`
- `output_types_are_invalid` at `juniper/src/validation/rules/variables_are_input_types.rs:67`

It could make sense to check that `validate_var_defs` and the rest of execution is never invoked when document validation fails (in a very general sense), but beside this I don't see any meaningful way to test that the `unreachable!` is never reached...

LegNeato:
Thanks for the PR and the explanation!

PR #464, opened 2019-11-21
Loosen constraints on uuid version.

jimsynz:
The `uuid` maintainers have started releasing in the `0.8` version train.  I've relaxed the version requirements on `juniper`'s dependencies to allow juniper users to specify a different version in their `Cargo.toml` and still have the integration work.


theduke:
Hey, please keep the constraint as `>= 0.7, < 0.9`, since we can't be sure there won't be breaking changes after 0.8 that affect our code.

jimsynz:
@theduke updated.

PR #466, opened 2019-11-23
docs(quickstart): Fix juniper version so the examples work

klausi:
Got the error

```
error[E0433]: failed to resolve: could not find `object` in `juniper`
  --> server/src/main.rs:20:12
   |
20 | #[juniper::object(
   |            ^^^^^^ could not find `object` in `juniper`
```

due to the wrong version mentioned in the quickstart docs. This should fix it!


theduke:
Thanks

p-alik:
https://graphql-rust.github.io/juniper/current/quickstart.html still shows `juniper = "0.11"`

PR #468, opened 2019-11-28
WIP: Add authorization to juniper.

AurelienFT:
WIP

### DIFF
--- a/integration_tests/juniper_tests/src/codegen/derive_object_with_raw_idents.rs
+++ b/integration_tests/juniper_tests/src/codegen/derive_object_with_raw_idents.rs
@@ -1,0 +1,105 @@
+#[cfg(test)]
+use fnv::FnvHashMap;
+#[cfg(test)]
+use juniper::Object;
+
+#[cfg(test)]
+use juniper::{
+    self, execute, graphql_value, DefaultScalarValue, EmptyMutation, GraphQLInputObject,
+    GraphQLType, RootNode, Value, Variables,
+};
+
+pub struct Query;
+
+#[juniper::object]
+impl Query {
+    fn r#type(r#fn: MyInputType) -> Vec<String> {
+        unimplemented!()
+    }
+}
+
+#[derive(GraphQLInputObject, Debug, PartialEq)]
+struct MyInputType {
+    r#trait: String,
+}
+
+#[test]
+fn supports_raw_idents_in_types_and_args() {
+    let doc = r#"
+    {
+        __type(name: "Query") {
+            fields {
+                name
+                args {
+                    name
+                }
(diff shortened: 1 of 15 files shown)


LegNeato:
I am not sure we would take a PR to do this. For example, this assumes a certain structure of a permission system which may not work for all apps.

What is stopping you from doing this on top of juniper? Why not wrap juniper `objects` and such so that your authorization code is called by the resolver?

EituKo:
`What is stopping you from doing this on top of juniper? Why not wrap juniper objects and such so that your authorization code is called by the resolver?`
This is what we were planning to do, but I don't quite understand how we could implement this without a PR. Can you provide us some pseudo code to help us comprehend?

andy128k:
@EituKo Resolvers can return an error (`FieldResult<...>`).

```rust
#[juniper::object(context = MyContext)]
impl Query {
    fn protected_field(mycontext: &MyContext) -> FieldResult<YourTypeHere> {
         if context.authorize() {
              Ok(/* your code here */)
         } else {
              Err("Access denied".to_string())
         }
    }
}
```

LegNeato:
As @andy128k points out, can you can call whatever you want in the resolvers. You would manually something similar to above for each field you need auth. If that is repetitive you can create your own macros or derives with your own annotations.

For example, you could create a macro `myauth_graphql_object:`:

```rust
#[myauth_graphql_object(context = MyContext)]
impl Query {
    #[myauth(protected)]
    fn protected_field(mycontext: &MyContext) -> YourTypeHere {
     // your protected code here
    }
    fn unprotected_field(mycontext: &MyContext) -> YourTypeHere {
     // your unprotected code here
    }
}
```
That would spit out:
```rust
#[juniper::object(context = MyContext)]
impl Query {
    fn protected_field(mycontext: &MyContext) -> FieldResult<YourTypeHere> {
         if context.authorize() {
              Ok(/* your protected code here */)
         } else {
              Err("Access denied".to_string())
         }
    }

    fn unprotected_field(mycontext: &MyContext) -> YourTypeHere {
         /* your unprotected code here */)
    }
}
```

(probably best to make it protected by default in a real implementation though to prevent security bugs).

Note the generated `FieldResult` and if statement due to the `myauth` annotation.

Here `myauth_graphql_object` is a [procedural macro](https://doc.rust-lang.org/reference/procedural-macros.html). I'm not sure if proc macros are multiple passes so if you can directly generate the call to`juniper::graphql_object`. If not your auth macro can depend on `juniper_codegen` and just output the juniper boilerplate directly. 

You could add whatever you want in there and generate whatever code makes sense for you.

Let me know if this doesn't solve the issue!

PR #478, opened 2019-12-10
Fix rocket default features

SamDavies:
This removes the dependencies on `ring` inside the `rocket` crate.

Allowing me to use the `jsonwebtoken` crate which requires a newer version of `ring`.


LegNeato:
I landed this as https://github.com/graphql-rust/juniper/pull/530, thank you so much for the pull request!  ❤️

PR #479, opened 2019-12-10
Implement GraphQLTypeAsync for Arc

obmarg:
I'm building a GraphQL API using Juniper that proxies another GraphQL
API.  It does a large fetch upfront from the underlying GraphQL API,
transforms it into a different format and then implements some
resolvers that do some further filtering.

One of these resolvers ends up looking like:

```
async fn items(&self, ...) -> Vec<Item> {
    self.items.iter().filter(...).collect()
}
```

This causes us problems as we're returning owned Item's and Item is a
large nested structure that would be expensive to clone.

Our current work around was to put Item into an Arc, as Arc is
comparatively cheap to clone.  So our method becomes:

```
async fn items(&self, ...) -> Vec<Arc<Item>> {
    self.items.iter().filter(...).map(Arc::clone).collect()
}
```

However to support this we needed Arc to implement GraphQLTypeAsync.
This commit adds that support to juniper, by forwarding to the
GraphQLTypeAsync implementation for the contained type.

It's possible that we could have acheived something similar by adding
some lifetimes to our resolver and returning a reference, but using an
Arc was easier for us in this case.  I'm not sure if there's any reason
why this would be a bad addition to Juniper overall?

### DIFF
--- a/juniper/src/types/pointers.rs
+++ b/juniper/src/types/pointers.rs
@@ -216,6 +216,25 @@ where
     }
 }
 
+#[cfg(feature = "async")]
+impl<'e, S, T> crate::GraphQLTypeAsync<S> for std::sync::Arc<T>
+where
+    S: ScalarValue + Send + Sync,
+    T: crate::GraphQLTypeAsync<S>,
+    <T as crate::types::base::GraphQLType<S>>::TypeInfo: Sync + Send,
+    <T as crate::types::base::GraphQLType<S>>::Context: Sync + Send,
+{
+    fn resolve_async<'a>(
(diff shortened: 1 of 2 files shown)


LegNeato:
Thank you for the PR! Let me think about this. I think we would want the corresponding sync version too, no?

obmarg:
@LegNeato I went to implement this for the sync types as well, but turns out it's already implemented here: https://github.com/graphql-rust/juniper/blob/master/juniper/src/types/pointers.rs#L149-L198

So I guess this PR is just bringing support into the Async feature that already existed for sync.

Issue #482, opened 2019-12-12
`#[graphql(skip)]` does not works on complex fields


IcanDivideBy0:
**Describe the bug**
When using `#[graphql(skip)]` on a "complex" field (a struct fn) the field is still avaliable through the GraphQL endpoint.

**To Reproduce**
```rust
pub struct Foo {
    pub foo: i32,
    pub bar: i32,
}

#[juniper::object]
impl Foo {
    pub fn foo(&self) -> i32 {
        self.foo
    }

    #[graphql(skip)]
    pub fn bar(&self) -> i32 { // This field can still be accessed through the API
        self.bar
    }
}
```

**Expected behavior**
`#[graphql(skip)]` should works on both "simple" and "complex" fields.

**Additional context**
N/A

davidpdrsn:
Would you expect that `bar` would become a method on `Foo`? If so that isn't quite how `juniper::object` works. [Attribute macros](https://doc.rust-lang.org/reference/procedural-macros.html#attribute-macros) are able to "erase" the input code and instead insert something completely different. That is what `juniper::object` does. It generates an implementation of [GraphQLType](https://docs.rs/juniper/0.14.1/juniper/trait.GraphQLType.html) and doesn't keep the `impl` block that you wrote. So `foo` and `bar` don't become methods on `Foo`.

Long story short, but if you want to be able to call `Foo::bar` you need to add another `impl Foo` that isn't annotated with `#[juniper::object]` and define the method there.

LegNeato:
We would love a PR explaining this in the book!

IcanDivideBy0:
Ok, thanks for the answer. It feels a bit weird to write twice the `impl` block, but it works well this way.

PR #483, opened 2019-12-13
Add note on complex fields impl block in doc

IcanDivideBy0:
Adding a small note in doc to avoid confusion about `impl` block for complex fields.
See #482

### DIFF
--- a/docs/book/content/types/objects/complex_fields.md
+++ b/docs/book/content/types/objects/complex_fields.md
@@ -25,6 +25,14 @@ impl Person {
     }
 }
 
+// Note that this syntax generates an implementation of the GraphQLType trait,
+// the base impl of your struct can still be written like usual:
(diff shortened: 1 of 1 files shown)


docs/book/content/types/objects/complex_fields.md
```diff
@@ -25,6 +25,14 @@ impl Person {
     }
 }
 
+// Note that this syntax generates an impl block for the GraphQLType trait,
+// the base impl of your struct can still be writen like usual:
```
davidpdrsn: Sounds good but I would tweak the wording a bit:

```rust
// Note that this syntax generates an implementation of the GraphQLType trait,
// the base impl of your struct can still be written like usual:
```

LegNeato:
Thanks for the PR!

PR #493, opened 2020-01-12
Master to async await

iancormac84:
Attempt at rebase of async-await branch on to master. One test fails.


LegNeato:
Wow, thank you! I haven't had a ton of time due to work and this should help unblock a ton of folks.

PR #505, opened 2020-01-26
[WIP] Update juniper_hyper to hyper 0.13 and add async resolution

wingertge:
The integration test times out at this point, but I've tested it in my own project and it works there so I'm open for any tips on how to fix the test and get this ready for merging.

### DIFF
--- a/juniper_hyper/src/lib.rs
+++ b/juniper_hyper/src/lib.rs
@@ -405,7 +405,7 @@ mod tests {
         EmptyMutation, RootNode,
     };
     use reqwest::{self, Response as ReqwestResponse};
-    use std::{net::SocketAddr, sync::Arc, time::Duration};
+    use std::{net::SocketAddr, sync::Arc, time::Duration, thread};
 
     struct TestHyperIntegration;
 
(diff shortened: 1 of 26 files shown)


LegNeato:
I re-ran the tests, looks like a legit failure:

```
error[E0432]: unresolved import `juniper::GraphQLTypeAsync`
  --> juniper_hyper/src/lib.rs:10:150
   |
10 | use juniper::{http::GraphQLRequest as JuniperGraphQLRequest, serde::Deserialize, DefaultScalarValue, GraphQLType, InputValue, RootNode, ScalarValue, GraphQLTypeAsync};
   |                                                                                                                                                      ^^^^^^^^^^^^^^^^
   |                                                                                                                                                      |
   |                                                                                                                                                      no `GraphQLTypeAsync` in the root
   |                                                                                                                                                      help: a similar name exists in the module: `GraphQLType`
```

Likely that import needs to be gated on the `async` feature?

juniper_hyper/src/lib.rs
```diff
@@ -357,44 +430,58 @@ mod tests {
 
     #[test]
     fn test_hyper_integration() {
```
LegNeato: I believe because you use the `macros` feature you can use `#[tokio::test]` on an async fn.

wingertge:
Aha, it seems the test only fails on nightly, which I was using. Unfortunately I get an error saying pear requires nightly if I try to test on stable, even though the nightly version it requires is 8 minor versions behind stable. So I can't test it on anything but nightly myself, which makes it hard to diagnose.

However, through print debugging I found out the server starts responding to requests as soon as the integration times out, so apparently even with tokio::spawn, the integration test blocks the main thread.

juniper_hyper/src/lib.rs
```diff
@@ -215,45 +276,74 @@ where
...
+                (is_ok, body)
+            }
+            GraphQLRequest::Batch(requests) => {
+                let futures = requests
+                    .iter()
+                    .map(|request| request.execute_async(&root_node, &context))
+                    .collect::<Vec<_>>();
+                let results = futures::future::join_all(futures).await;
+
+                let is_ok = results.iter().all(|res| res.is_ok());
```
LegNeato: Rather than checking for `all`, you can just look for one failure with `any`. Likely a  tiny bit faster in the case of an error.

wingertge: That was from the original code, but I fixed it anyways now.

LegNeato:
I think it is fine as-is and iterate in follow-ups. We are running both async and non-async on CI. Thanks for the PR! 🍻

PR #517, opened 2020-01-30
Add bson crate's ObjectId to juniper foreign scalar type integrations.

iancormac84:
ObjectId is used in the [mongodb](https://crates.io/crates/mongodb) crate.


iancormac84:
Done! Let me know if anything else is needed.

Issue #444, opened 2019-10-28
Document N+1 mitigation techniques in the book


MikailBag:
**Is your feature request related to a problem? Please describe.**
Now, resolver signature looks like:
`fn field(&self, /*args*/)`.
This can easily lead to N + 1 DB query problem. (Or even N**K if query has K levels of depth). To avoid this problem, special libraries should be used, that somehow accumulate queries and merge them into one.

**Describe the solution you'd like**
(Idea itself is very simple. It is possible that is was already discussed and rejected, but I didn't find such an issue). I propose to allow batched resolvers. Signature now looks smth like:
`fn field(self: &[Self], ctx:&Ctx, args: ...)`
Now, from N**K requests we go to K (each level will be processed with _one_ resolver execution and _one_ db request).
**Describe alternatives you've considered**
Alternative: do nothing.
However, I believe batched resolvers  provide good enough performance with high flexibility.

tyranron:
@MikailBag for N + 1 query problem the usual solution is either [dataloader](https://github.com/cksac/dataloader-rs) or [lookahead](https://github.com/graphql-rust/juniper/issues/387).

As for proposed solution I'm unsure how well it will fit. Usually, resolvers may contain allowed-resoling-depth checks, authorization-rights check, authentication checks, etc, which all are query-scoped. Resolving multiple queries at the same time, but now with the same context, seems to be less flexible.

davidpdrsn:
If you're interested in the lookahead approach you could consider [juniper-eager-loading](https://crates.io/crates/juniper-eager-loading) which handles all the boilerplate. We're using it in production at @tonsser and its been working well so far. Full disclosure: I'm the maintainer of the library.

theduke:
As mentioned by others, the possible approaches here are

* datalaoder
* using lookahead
* uniper-eager-loading

We should really add to the book, since it's a common stumbling block.

MikailBag:
Thanks for the responses!
I will study your proposals (so far juniper-eager-loading looks the most promising).

Regarding book, I would like to see quite complex example, e.g.:
- Response data are aggregated from DB and in-memory source (e.g. calculated).
- Application doesn't directly use diesel, it uses some trait, `MyAppDbProvider`.
- And without too much glue code :)

davidpdrsn:
@MikailBag I maintain a complete example app. You can find it here https://github.com/davidpdrsn/graphql-app-example. It does use Diesel directly however.

[...]

PR #518, opened 2020-01-31
WIP: Add dataloader explaination to book

jayy-lmao:
This PR is intended to address #444 
I had a similar problem to the issue author finding a decent explanation/example of using data loaders with juniper, and so want to give an example to make it easier for others.

Happy to make any changes from nitpicks to overhauls, cull some or explain more :grin: 

@MIkailBag feedback would be appreciated if you can spare the time!

### DIFF
--- a/docs/book/content/advanced/dataloaders.md
+++ b/docs/book/content/advanced/dataloaders.md
@@ -1,0 +1,132 @@
+# Avoiding the N+1 Problem With Dataloaders
+
+A common issue with graphql servers is how the resolvers query their datasource.
+his issue results in a large number of unneccessary database queries or http requests.
+Say you were wanting to list a bunch of cults people were in
+
+```graphql
+query {
+  persons {
+    id
+    name
+    cult {
+      id
+      name
+    }
+  }
+}
+```
+
+What would be executed by a SQL database would be:
+
+```sql
+SELECT id, name, cult_id FROM persons;
+SELECT id, name FROM cults WHERE id = 1;
+SELECT id, name FROM cults WHERE id = 1;
+SELECT id, name FROM cults WHERE id = 1;
+SELECT id, name FROM cults WHERE id = 1;
+SELECT id, name FROM cults WHERE id = 2;
+SELECT id, name FROM cults WHERE id = 2;
+SELECT id, name FROM cults WHERE id = 2;
+# ...
+```
+
+Once the list of users has been returned, a separate query is run to find the cult of each user.
+You can see how this could quickly become a problem.
+
+A common solution to this is to introduce a **dataloader**. A dataloader batches resolver execution (and _can_ even cache it).
+This can be done with Juniper using the crate [cksac/dataloader-rs](https://github.com/cksac/dataloader-rs). 
+
+!FILENAME Cargo.toml
+
+```toml
+[dependencies]
+actix-identity = "0.2"
+actix-rt = "1.0"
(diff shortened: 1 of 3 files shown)


docs/book/content/advanced/dataloaders.md
```diff
@@ -0,0 +1,132 @@
...
+    ctx.cult_loader.cult_by_id.load(id).await.unwrap()
+  }
+}
+
+```
+
+As seen immediately above, when a new `PersonBatcher` is instantiated,
+it can be called with `.load(id)` to return a `Future` which resolves to a value with `await`.
+
+A **Dataloader** should be created in context, either at the server start, or on a per-request basis.
```
LegNeato: It must be per-request or per auth context. If you don't do it per-request / auth context you get security bugs as each user can load cached data from another user. Not sure how best to get this across.

docs/book/content/advanced/dataloaders.md
```diff
@@ -0,0 +1,132 @@
...
+SELECT id, name FROM cults WHERE id = 2;
+SELECT id, name FROM cults WHERE id = 2;
+SELECT id, name FROM cults WHERE id = 2;
+# ...
+```
+
+Once the list of users has been returned, a separate query is run to find the cult of each user.
+You can see how this could quickly become a problem.
+
+A common solution to this is to introduce a **dataloader**. A dataloader batches resolver execution (and _can_ even cache it).
```
LegNeato: Nit: _may_ decide to cache it though doing so it not required.

docs/book/content/advanced/dataloaders.md
```diff
@@ -0,0 +1,132 @@
...
+  }
+}
+
+```
+
+As seen immediately above, when a new `PersonBatcher` is instantiated,
+it can be called with `.load(id)` to return a `Future` which resolves to a value with `await`.
+
+A **Dataloader** should be created in context, either at the server start, or on a per-request basis.
+If it is created within a resolver then it will be created for each time it needs to resolve a field,
```
LegNeato: This is good info but feels a bit clumsy.

jayy-lmao: I'll reword it :)

docs/book/content/advanced/dataloaders.md
jayy-lmao: Ah! I hadn't considered this. 
It makes plenty of sens but I may have to learn more about auth contexts as I am unfamiliar. (Unless it's just referring to a context which behaves as auth middleware and checks auth before returning some kind of state?)

LegNeato: Yeah, there is nothing built in. Basically dataloader can only share data by user. Even that gets dicey as permissions may have changed and you may be able to load cached data you can no longer see. It's really clean / safe to do it per request as you know 1) the user doing the request and 2) they have permission to see that data now. So dataloader is used for N+1 / request fanout but the cache isn't long-lived, as that would introduce possible temporal issues with permissions. With request-based loading you are reasonably sure the permissions don't change out from under the cache.

jayy-lmao: It sounds like it may be better to just suggest per-request in this section then, and omit other strategies. Would you be comfortable with me doing that?

LegNeato:
Thank you so much for this! This has been a big pain point for folks and documenting will help many people 🎉🍻

PR #540, opened 2020-02-20
feat(juniper_codegen): add conversions for enum variants to strings

garbados:
Hi!

This PR expands the `GraphQLEnum` derive macro to provide conversions from enum variants into strings. Currently, the macro only provides conversions for `InputValue`, but in my usage it became necessary to map enum variants into `Value::scalar::<String>` types so that they can be used in GraphQL responses, specifically as the second parameter of `FieldError::new`. Mapping enum variants to strings directly, rather than a Juniper type, is an artifact of my not knowing much.

I'm new to Rust and even newer to Juniper, so this PR is intended for review and discussion, and I am happy to incorporate any feedback.

### DIFF
--- a/juniper_codegen/src/derive_enum.rs
+++ b/juniper_codegen/src/derive_enum.rs
@@ -147,6 +147,7 @@ pub fn impl_enum(ast: &syn::DeriveInput, is_internal: bool) -> TokenStream {
     let mut resolves = TokenStream::new();
     let mut from_inputs = TokenStream::new();
     let mut to_inputs = TokenStream::new();
+    let mut to_strings = TokenStream::new();
 
     for variant in variants {
         match variant.fields {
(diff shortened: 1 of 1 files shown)


LegNeato:
Hello! Thank you for the PR and welcome! 🍻 

First, I think there are some legit errors on CI related to this PR:

https://dev.azure.com/graphql-rust/GraphQL%20Rust/_build/results?buildId=684&view=logs&j=150c9611-1cfc-59c0-d4e8-92df330f7380&t=10bef84d-ad0f-569a-c6cb-c54604a7beb6&l=458

(sorry, our CI is a bit messy right now...I am working to clean it up).

Second, I _think_ this sounds like a great change but need to think about it a bit more. I haven't really touched enums too much 😄

garbados:
> First, I think there are some legit errors on CI related to this PR:

I see those errors but none are related to the code I added, so I don't know how to make sense of them . Can you provide any guidance?

Issue #534, opened 2020-02-18
Need to add a `use futures;` statement to object derive w/ async feature.


thedodd:
**Describe the bug**
On the master branch, using the `async` feature, use of the derive macro `#[juniper::graphql_object(...)]` will cause a compilation error stating that `futures` is not in scope. Looks like this just needs a `use futures;` somewhere in the macro.

**To Reproduce**
Just pop a `#[juniper::graphql_object(...)]` on any valid struct with the `async` feature enabled, and don't `use futures;` anywhere in the module.

**Expected behavior**
Pretty common issue in macros. Just need to add a use statement.

samuela:
I think this might be why `cargo test` is failing on 547cf3e36fa0532476295186b86954fab3a258f6 too:
```
error[E0432]: unresolved import `futures`
  --> integration_tests/juniper_tests/src/issue_398.rs:39:1
   |
39 | #[juniper::graphql_object]
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^ use of undeclared type or module `futures`
   |
   = note: this error originates in an attribute macro (in Nightly builds, run with -Z macro-backtrace for more info)
```

PR #541, opened 2020-02-21
Add futures dependency

LegNeato:
Fixes https://github.com/graphql-rust/juniper/issues/534.

I am not sure if the best way is to have folks bring their own futures for the
derives / macros or to pull it in as a dep explicitly like I have done here.

Because people are filing issues, we will do it this way and see what other issues arise.


davidpdrsn:
I think this approach is fine 👍

Issue #532, opened 2020-02-14
execute_validated_query_async is using std::sync::RwLock  


jxs:
[`execute_validated_query_async`](https://github.com/graphql-rust/juniper/blob/master/juniper/src/executor/mod.rs#L745) is using `std::sync::RwLock` which blocks the thread while waiting to read/write, this will probably block the whole runtime if running under tokio's [`basic_scheduler`](https://docs.rs/tokio/0.2.11/tokio/runtime/struct.Builder.html#method.basic_scheduler).  

Both [`tokio`](https://docs.rs/tokio/0.2.11/tokio/sync/struct.RwLock.html) and [`async-std`](https://docs.rs/async-std/1.5.0/async_std/sync/struct.RwLock.html) offer `async` versions of the RwLock, there's also the `futures` crate [`Mutex`](https://docs.rs/futures/0.3.1/futures/lock/struct.Mutex.html)

tyranron:
@jxs @LegNeato looking at the code in the [current master](https://github.com/graphql-rust/juniper/tree/aedb2d1eb453f605380ff527f88d7467253ba341), the described above is totally non-issue. `RwLock`'s locking is used in a [single place only in sync function](https://github.com/graphql-rust/juniper/blob/aedb2d1eb453f605380ff527f88d7467253ba341/juniper/src/executor/mod.rs#L608) and there is literally no way for the lock guard to live accross any `.await` point to block the underlying executor. This is totally correct usage of `RwLock` there (I'd rather prefer even regular `Mutex`, as there is no `read()` blocks at all), and using any async `Mutex`/`RwLock` will just create extra overhead with no benefit at all, as we do not require to block across `.await` points.

PR #542, opened 2020-02-21
Do not use RwLock for async execution

LegNeato:
Fixes https://github.com/graphql-rust/juniper/issues/532.

This is pretty janky because `async` isn't an either-or. You can still
execute sync with the `async` feature enabled. I also don't want to require
`futures` if the `async` flag is off and didn't want to add more feature flags for
`tokio` and `async-std`.

I could have implemented an `AsyncExecutor` when the `async` feature is on
but that would have required a lot of plumbing. I will likely come back around
and do that eventually.

If other folks have better ideas, let me know.

### DIFF
--- a/juniper/src/executor/mod.rs
+++ b/juniper/src/executor/mod.rs
@@ -569,7 +569,8 @@ where
     /// Add an error to the execution engine at the current executor location
     #[cfg(feature = "async")]
     pub async fn push_error_async(&self, error: FieldError<S>) {
-        self.push_error_at_async(error, self.location().clone()).await;
+        self.push_error_at_async(error, self.location().clone())
+            .await;
     }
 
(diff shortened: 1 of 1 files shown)


tyranron:
@LegNeato I'm in favor of droping sync execution at all, leaving the async only, without any feature flags (similarly to what `reqwest` does). If one requires sync execution, he may do sync ops in resolvers and synchronously `.wait()` for the GraphQL request execution. There is no need to have 2 versions of execution, regarding the situation async ecosystem is growing in Rust.

davidpdrsn:
While dropping sync support would be annoying for the users who have to update I actually agree it might be the right move long term.

LegNeato:
I'm inclined to agree. It will also simplify the code immensely.

Issue #549, opened 2020-03-01
Macro `#[juniper::graphql_union]` does not impl `GraphQLTypeAsync`


UkonnRa:
**Describe the bug**

In the [master branch](https://github.com/graphql-rust/juniper/blob/a0b854f31c586fcf786f7ff1d5d563c5ce2d8b79/juniper_codegen/src/impl_union.rs#L155), bacause macro `juniper::graphql_union` does not impl `GraphQLTypeAsync`, the async function is not usable

LegNeato:
Yep, unions and interfaces are not supported yet on master.

jadedevin13:
This doesn't work in custom scalars as well.

arlyon:
All the tests pass on my machine with a small change to a lifetime:

```rust
diff --git a/juniper/src/macros/tests/union.rs b/juniper/src/macros/tests/union.rs
index 6dfbfe75..089ea3b2 100644
--- a/juniper/src/macros/tests/union.rs
+++ b/juniper/src/macros/tests/union.rs
@@ -57,7 +57,7 @@ impl CustomName {
 }
 
 #[crate::graphql_union_internal]
-impl<'a> WithLifetime<'a> {
+impl<'b> WithLifetime<'b> {
     fn resolve(&self) {
         match self {
             Concrete => match *self {
```

PR #552, opened 2020-03-06
Attempt first naive version of async union

xla:
This addreses #549 as we ran into the lack of support for unions when trying to migrate to `async/await`. **This is a severe case of i-have-no-clue what I'm doing.** As mostly inspired by the enum implementation copy-pastad code at the moment, I look very much for feedback on how to make the async support proper and according to the standards of the crate. So don't go easy here, thanks.

### DIFF
--- a/juniper_codegen/src/impl_union.rs
+++ b/juniper_codegen/src/impl_union.rs
@@ -2,6 +2,7 @@ use proc_macro::TokenStream;
 
 use proc_macro_error::MacroError;
 use quote::quote;
+use syn::parse_quote;
 use syn::spanned::Spanned;
 
 use crate::util;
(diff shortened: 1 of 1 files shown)


LegNeato:
It looks likes the tests are failing

arlyon:
All the tests pass on my machine with a small change to a lifetime:

```rust
diff --git a/juniper/src/macros/tests/union.rs b/juniper/src/macros/tests/union.rs
index 6dfbfe75..089ea3b2 100644
--- a/juniper/src/macros/tests/union.rs
+++ b/juniper/src/macros/tests/union.rs
@@ -57,7 +57,7 @@ impl CustomName {
 }
 
 #[crate::graphql_union_internal]
-impl<'a> WithLifetime<'a> {
+impl<'b> WithLifetime<'b> {
     fn resolve(&self) {
         match self {
             Concrete => match *self {
```

LegNeato:
I'll have to look more later, but one should be able to name lifetimes whatever. So likely the macro generated lifetime needs to have a name like '__a to avoid conflicts.

LegNeato:
Also note some union tests are turned off:

https://github.com/graphql-rust/juniper/blob/6e4bc127b05dc29f7832cbf468c9fb2bc18f78ec/juniper/src/executor_tests/mod.rs#L7

They need to be turned back on and pass.

PR #569, opened 2020-03-17
Executor.schema(&self) is kind of useless without declaring SchemaTyp…

jadedevin13:
…e as public


jadedevin13:
This is also useful if you want to search the concrete_type_by_name when using lookaheads.

LegNeato:
Thanks for the PR!

LegNeato:
Awesome, thanks again! 🍻

jadedevin13:
Thanks. Awesome crate. More power!

PR #572, opened 2020-03-18
doc: remove outdated examples from third-party integrations list

ubnt-intrepid:
...since I've stopped the development both of `tsukuyomi` and `finchers` a few months ago.


LegNeato:
Thank you! 🍻

PR #583, opened 2020-03-23
fix: websocket connection in subscriptions example

spacemeowx2:
Added `Sec-WebSocket-Protocol` header to avoid clients refusing connections.


LegNeato:
Thanks again!

PR #592, opened 2020-03-29
Add `gql!` procedural macro

### LINKED ISSUE #553: Automatic getters for `#[graphql_object]`
I'm always frustrated when I need to replace:
```rust
struct Foo {
    name: String
}
```
by
```rust
struct Foo {
    name: String
}
#[juniper::graphql_object]
impl Foo {
    fn computed(&self) -> String {
        ...
    }
}
```

because I lose the nice default field accessor for `name` and I have to manually add
```rust
#[juniper::graphql_object]
impl Foo {
    ...
    fn name(&self) -> &String {
        &self.name
    }
}
```
I'd like for `#[derive(juniper::GraphQLObject)]` and  `#[juniper::graphql_object]` to work together so I don't have to add the dummy accessors for every field in the `impl` block.
I would imagine it is complicated to achieve this because both macros currently generate the same `impl` block underneath.

I couldn't come up with any alternative to this :crying_cat_face:

**Additional context**
This is extra painful when wrapping REST interfaces which have many simple fields (like strings) and a few complex fields (references to other resources).

P.S.: `juniper` is such a great crate! many thanks to all the contributors! :100:

Victor-Savu:
:warning: This is just a draft PR to support the discussion around a proposal. Suggestions for improving the code are welcome, but please keep in mind that it is pretty much the first version that worked so it will need a few iterations.   

To avoid the boilerplate necessary for exposing struct members as fields whenever a complex field is needed as well, the `gql!` function-like procedural macro offers a unified syntax for simple and complex fields:

```rust
use juniper::graphql_value;

struct Textcon {
    accounts: HashMap<i32, Account>,
}

impl juniper::Context for Textcon {}

juniper::gql! {

#[derive(Debug)]
#[graphql(Context = Textcon)]
Account(
    usd_cents: i32,  // usd_cents *is not* exposed as a field
) {
    // balance *is* exposed as a complex field
    fn balance(&self) -> String {
        format!("${}.{:2}", self.dollars(), self.cents()
    }
}


#[derive(Debug)]
#[graphql(Context = Textcon)]
Customer(
    account_ids: Vec<i32>,   // account_ids *is not* exposed as a field
) {
    name: String   // name *is* exposed as a simple field

    // accounts *is* exposed as a complex field
    fn accounts(&self, context: &Textcon) -> juniper::FieldResult<Vec<&Account>, _> {
        let mut acc = vec![];
        for account_id in &self.account_ids {
            acc.push(
                context.accounts.get(account_id).ok_or(
                    juniper::FieldError::new(
                        format!("Invalid account id: {}", account_id),
                        graphql_value!({"": ""}))
                )?
            )
        }
        Ok(acc)
    }
}

#[graphql(context=Textcon)]
MutationRoot{}  // has no fields

#[graphql(context=Textcon)]
QueryRoot{
    // customers *is* exposed as a complex field
    async fn customers() -> Vec<Customer> {
        vec![Customer{
            name: "X".to_owned(),
            account_ids: vec![1, 4, 0],
        },
        Customer{
            name: "Y".to_owned(),
            account_ids: vec![2, 3, 7],
        },
        Customer{
            name: "Z".to_owned(),
            account_ids: vec![5, 6],
        }]
    }
}

}

impl Account {
    fn cents(&self) -> i32 {
        self.usd_cents % 100
    }

    fn dollars(&self) -> i32 {
        self.usd_cents / 100
    }
}

type Schema = RootNode<'static, QueryRoot, MutationRoot, EmptySubscription<Textcon>>;
```

There are a few particularities that stand out:
1. Simple and complex fields are defined between curly braces `{ ... }`
1. Members that are not exposed as fields are optionally defined between parentheses `( ... )`. I considered using individual specifiers (similar to how `pub` works) to tell fields from non-fields, but went with this approach as a matter of preference. Note that this applies only to members. There can always be an `impl` block to add methods, just like the `impl Account { .. }` in the example above.
1. There is no `struct` or `impl` specification. This is just because I didn't want to bother with what token to use there. Any suggestions :)?

Resolves #553

### DIFF
--- a/juniper/src/lib.rs
+++ b/juniper/src/lib.rs
@@ -115,7 +115,7 @@ extern crate bson;
 // This allows users to just depend on juniper and get the derive
 // functionality automatically.
 pub use juniper_codegen::{
-    graphql_object, graphql_subscription, graphql_union, GraphQLEnum, GraphQLInputObject,
+    gql, graphql_object, graphql_subscription, graphql_union, GraphQLEnum, GraphQLInputObject,
     GraphQLObject, GraphQLScalarValue,
 };
 // Internal macros are not exported,
--- a/juniper_codegen/src/gql.rs
+++ b/juniper_codegen/src/gql.rs
@@ -1,0 +1,447 @@
+extern crate proc_macro;
+extern crate proc_macro2;
+
+use crate::util;
+// use quote::quote;
+use quote::quote;
+use std::iter::{FromIterator, IntoIterator};
+use syn::{
+    parse::{Parse, ParseStream, Result},
+    Token,
+};
+
+#[derive(Debug, Clone)]
+struct NamedField {
+    attrs: Vec<syn::Attribute>,
+    vis: syn::Visibility,
+    ident: syn::Ident,
+    colon_token: syn::Token![:],
+    ty: syn::Type,
+}
+
+impl NamedField {
+    fn into_field(self) -> syn::Field {
+        syn::Field {
+            attrs: self
+                .attrs
+                .into_iter()
+                .filter(|attr| !util::path_eq_single(&attr.path, "graphql"))
+                .collect(),
+            vis: self.vis,
+            ident: Some(self.ident),
+            colon_token: Some(self.colon_token),
+            ty: self.ty,
+        }
+    }
+}
+
+impl Parse for NamedField {
+    fn parse(input: ParseStream) -> Result<Self> {
+        Ok(Self {
+            attrs: input.call(syn::Attribute::parse_outer)?,
+            vis: input.parse()?,
+            ident: input.parse()?,
+            colon_token: input.parse()?,
+            ty: input.parse()?,
+        })
+    }
+}
+
+#[derive(Debug)]
+struct GqlObject {
+    pub attrs: Vec<syn::Attribute>,
+    pub vis: syn::Visibility,
+    pub ident: syn::Ident,
+    pub data_fields: Option<syn::punctuated::Punctuated<NamedField, Token![,]>>,
+    pub fields: Vec<NamedField>,
+    pub methods: Vec<syn::ImplItemMethod>,
+}
+
+impl GqlObject {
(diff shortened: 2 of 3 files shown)


LegNeato:
I've been thinking about this a lot. Not sure I like wrapping everything in a macro, but also not sure how two proc macro / derive invocations would pass data back and forth without doing so.

Victor-Savu:
Thank you for the feedback! I also have the same worries. One of the great things about juniper is that it looks so much like rust. The only point where the illusion breaks a bit is that it looks like we have two impl blocks for the same type (one decorated with the macro).
I had a strong feeling that the design I went with in this PR is way too DSL-like to match the feeling that juniper users are accustomed to.

Do you think it would be useful if I created a separate crate (independent of juniper and in my own account) that offers this alternative syntax?  I am generally weary of software alternatives because they tend to divide the users and discourage collaboration. On the other hand, such alternatives offer the possibility to discover preferences that we might otherwise not be able to predict. Since you have a better feeling about the community around juniper and I just want to make shre that there is no risk of alienating / splitting the community, your opinion is in my view the most informed. I want people to adopt graphql (when it makes sense for their projects) and I think juniper should be the library they choose when they do that.

(... meanwhile, back on 🌏) : Of course, this concern is only assuming such a crate is worth using and that it would get some traction, both of which are huge assumptions. So there is a 99.999% chance that it does not matter anyway 😅 I just want to make sure that I don't do anything harmful, regardless of how unlikely it is to pan out.

(... aand were back on 🪐) : If I do end up with a new crate, the implementation in this PR depends on some private modules from juniper_generator but I wouldn't mind maintaining an alternative implementation.

Thank you for your feedback and I would really appreciate your input.

Victor-Savu:
Regarding the second part of yojr reply, please see my comment [here](https://github.com/graphql-rust/juniper/issues/553#issuecomment-612423320)

Victor-Savu:
I am closing this PR to avoid any further overhead. Thanks again for the feedback and for making juniper such great library!

PR #598, opened 2020-04-03
Section added `Testing clients/IDEs`

Nishchit14:


LegNeato:
Thank you very much for your PR! We already mention the two embedded IDEs elsewhere. I don't think we want juniper to be a general GraphQL resource.

Nishchit14:
No worries, Thanks for reviewing it. @LegNeato

PR #599, opened 2020-04-03
Add debug derive to GraphQLResponse

Dispersia:
It's useful to print a debug message after `execute` of its result


LegNeato:
Thanks! :beers:

PR #603, opened 2020-04-04
Implementation for juniper_actix

engylemure:
This PR implements the integration with actix-web 2.0 and has a initial study on the implementation of the GraphQL subscriptions int othe actix as well, i made some small addition in the juniper_subscriptions as well a submodule for storing the message types sent between the client and server, added a trait as well for handling some lifecycle events in the subscription such as on_connection and some other ones, since right now the implementation depends on a unsafe operation for sending messages this is just a alpha implementation and would like to receive some reviews on it.
I made some tests and examples as well for handling all things that were implemented.

## PR Checklist

Before submitting a PR, you should follow these steps to prevent redundant churn or CI failures:

- [X] Ensure proper formatting
- [X] Run all tests
- [X] Update the CHANGELOG

## Error in CI

for some reason my CI is not working and it is being caused by the juniper_rocket crate requiring the nightly build, the tests in the nightly are running just fine in the stable they are just breaking...


juniper/src/http/mod.rs
```diff
@@ -217,6 +217,125 @@ where
     }
 }
 
+/// The Expected structure of the decoded JSON document for either POST or GET requests
```
LegNeato, line 220: There is nothing JSON-specific about this, please remove.

juniper/src/http/mod.rs
```diff
@@ -217,6 +217,125 @@ where
...
+/// considering Batch request.
+///
+/// You can use Serde to deserialize the incoming JSON data directly into this
```
LegNeato, line 223: Ditto

juniper/src/http/mod.rs
LegNeato, line 220: Can this be pulled out into a separate change?

juniper_subscriptions/src/lib.rs
```diff
@@ -135,6 +135,94 @@ where
     }
 }
 
+/// Subscriptions Protocol Message Types
+/// to know more access https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md
+pub mod message_types {
```
LegNeato, line 140: Ideally this is an enum.

juniper_subscriptions/src/lib.rs
```diff
@@ -135,6 +135,94 @@ where
...
+    /// GQL_CONNECTION_INIT message.
+    fn on_connect(
+        &self,
+        // for now im considering connection_params as a &str
+        // but that could change since this API is not defined yet
+        // we could use a serde Value with only the payload value,
+        // in the moment im considering that this value is
+        // the first text message sent by the Client with the
+        // type GQL_CONNECTION_INIT
+        connection_params: &str,
```
LegNeato, line 189: I'd prefer to not have it there if it is not defined. If there are specific params, we should do an enum with an `::Other(String)`.

juniper_subscriptions/src/lib.rs
```diff
@@ -135,6 +135,94 @@ where
...
+    /// message to the client.
+    fn on_operation_complete(&self, context: &Context);
+    /// This function is called just before the connection is terminated.
+    fn on_disconnect(&self, context: &Context);
+}
+
+/// Empty SubscriptionLifeCycleHandler
+pub struct EmptySubscriptionLifecycleHandler {}
+
+impl EmptySubscriptionLifecycleHandler {
```
LegNeato, line 205: This should be deleted and just derive Default.

juniper_subscriptions/src/lib.rs
```diff
@@ -135,6 +135,94 @@ where
...
+        &self,
+        // for now im considering connection_params as a &str
+        // but that could change since this API is not defined yet
+        // we could use a serde Value with only the payload value,
+        // in the moment im considering that this value is
+        // the first text message sent by the Client with the
+        // type GQL_CONNECTION_INIT
+        connection_params: &str,
+        context: &mut Context,
+    ) -> Result<(), String>;
```
LegNeato, line 191: This should be an error enum rather than a string.

juniper_subscriptions/src/lib.rs
```diff
@@ -135,6 +135,94 @@ where
...
+    /// Server sends this message to indicate that a GraphQL operation is done, and no more data will arrive for the specific operation.
+    pub const GQL_COMPLETE: &str = "complete";
+    /// Client -> Server
+    /// Client sends this message in order to stop a running GraphQL operation execution (for example: unsubscribe)
+    pub const GQL_STOP: &str = "stop";
+}
+
+/// Trait based on the SubscriptionServer
+/// https://www.apollographql.com/docs/graphql-subscriptions/lifecycle-events/
+pub trait SubscriptionLifecycleHandler<Context>
```
LegNeato, line 175: I feel like this should be a state machine? Also `on_` functions don't feel idiomatic...not sure what the best way to handle it tough.

juniper/src/http/mod.rs
engylemure: https://github.com/graphql-rust/juniper/pull/608

engylemure: https://github.com/graphql-rust/juniper/pull/608

juniper/src/http/mod.rs
engylemure: https://github.com/graphql-rust/juniper/pull/608

juniper_subscriptions/src/lib.rs
engylemure: I made some changes but since I doesn't have much experience into how this state machine should be implemented a example should be of great help to me.

juniper_subscriptions/src/lib.rs
engylemure: What kind of values this enum should have?

juniper_subscriptions/src/lib.rs
```diff
@@ -135,6 +136,108 @@ where
     }
 }
 
+/// Enum of Subscription Protocol Message Types
```
LegNeato, line 139: Hmmmm, it looks like this is a non-standard extension right? The only standard I see is  http://spec.graphql.org/draft/#sec-Subscription

LegNeato:
Would it be too much trouble to pull actix-web stuff out of this and into another PR so we can look at the subscription changes on their own?

juniper_subscriptions/src/lib.rs
engylemure: Yeah, it make senses it is specific to the implementation of subscriptions over WS.

engylemure:
> Would it be too much trouble to pull actix-web stuff out of this and into another PR so we can look at the subscription changes on their own?

No, i made another PR with only these changes https://github.com/graphql-rust/juniper/pull/617 and merged this PR into it.

LegNeato:
It looks like the `ws` stuff is still in this PR. Can we just include the actix integration so I can merge it and then iterate on the websocket stuff in https://github.com/graphql-rust/juniper/pull/617?

engylemure:
> It looks like the `ws` stuff is still in this PR. Can we just include the actix integration so I can merge it and then iterate on the websocket stuff in #617?

Sure

LegNeato:
Thank you again, this looks great. The community has really wanted this!

Also, appreciate you making so many changes to the PR 🍻.

PR #606, opened 2020-04-07
Removed 'compat' feature from futures...

jairinhohw:
...on juniper_subscriptions and juniper_warp, which where not being used, permitting futures versions greater than 0.3.1 to be used.


LegNeato:
Thank you! 🥂

PR #610, opened 2020-04-09
Make `graphql_scalar` a proc macro

andrisak:
Draft implementation of a graphql_scalar proc macro. There are many similarities with the corresponding derive macro and they can probably share a lot of the codegen implementation.


integration_tests/juniper_tests/src/codegen/impl_scalar.rs
```diff
@@ -0,0 +1,42 @@
+use fnv::FnvHashMap;
+#[cfg(test)]
+use juniper::{
+    self, parser::ScalarToken, DefaultScalarValue, GraphQLType, InputValue, ParseScalarResult,
+    ParseScalarValue, Value,
+};
+
+struct UserId(String);
+
+// TODO Trait that the macro handles, move to proper location! Naming?
```
LegNeato, line 10: I don't think we actually need a trait..we can just generate the implementation directly (the current macro doesn't use a trait, right?). For example, there is no `GraphQLObject` trait and instead it just implements `GraphQLType` if I recall correctly.

integration_tests/juniper_tests/src/codegen/impl_scalar.rs
```diff
@@ -0,0 +1,42 @@
...
+trait ParseCustomScalarValue<T, S = DefaultScalarValue> {
+    fn resolve(&self) -> Value;
+    fn from_input_value(value: &InputValue) -> Option<T>;
+    fn from_str<'a>(value: ScalarToken<'a>) -> ParseScalarResult<'a, S>;
+}
+
+#[juniper::graphql_scalar2(name = "MyCustomName", description = "My custom description")]
```
LegNeato, line 17: You can just delete the `graphql_scalar!`, rename this, and macro and update all usages in tests. It might find some regressions as well.

integration_tests/juniper_tests/src/codegen/impl_scalar.rs
```diff
@@ -0,0 +1,42 @@
...
+impl ParseCustomScalarValue<UserId> for UserId {
```
LegNeato, line 18: For the existing proc macros this isn't a *real* trait for the impl I think. So this can just be `GraphQLScalar`. See https://github.com/graphql-rust/juniper/blob/master/integration_tests/juniper_tests/src/codegen/impl_union.rs#L37 for an example...there is no `GraphQLUnion` trait.

andrisak:
No, I thinks it's ready to merge now.

LegNeato:
Thank you so much for the high-quality PR! This makes Juniper much easier to use and much more idiomatic 🍻

PR #617, opened 2020-04-10
Implementation of some websocket utilities in juniper_subscriptions

engylemure:
Has the objective of storing some useful implementation that can be used in the subscriptions implementation over websocket.

### DIFF
--- a/juniper_subscriptions/src/lib.rs
+++ b/juniper_subscriptions/src/lib.rs
@@ -19,6 +19,24 @@ use juniper::{
     BoxFuture, ExecutionError, GraphQLError, GraphQLSubscriptionType, GraphQLTypeAsync, Object,
     ScalarValue, SubscriptionConnection, SubscriptionCoordinator, Value, ValuesStream,
 };
+/// Utilities for the implementation of subscriptions over WebSocket
+///
+/// This module provides some utilities for the implementation of Subscriptions over
+/// WebSocket, such as the [`GraphQLOverWebSocketMessage`] that contains the messages that
+/// could be sent by the server or client and a [`SubscriptionStateHandler`] trait that allows
+/// the user of the integration to handle some Subscription Life Cycle Events, its based on:
+///
+/// - [Subscriptions Transport over WS][SubscriptionsWsProtocol]
+/// - [GraphQL Subscriptions LifeCycle Events][GraphQLSubscriptionsLifeCycle]
+///
+/// [SubscriptionsWsProtocol]: https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md
+/// [GraphQLSubscriptionsLifeCycle]: https://www.apollographql.com/docs/graphql-subscriptions/lifecycle-events/
+/// [`GraphQLOverWebSocketMessage`]: GraphQLOverWebSocketMessage
+/// [`SubscriptionStateHandler`]: SubscriptionStateHandler
+#[cfg(feature = "ws-subscriptions")]
+pub mod ws_util;
+#[cfg(feature = "ws-subscriptions")]
+pub use ws_util::*;
 
 /// Simple [`SubscriptionCoordinator`] implementation:
 /// - contains the schema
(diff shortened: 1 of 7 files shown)


juniper_subscriptions/src/lib.rs
```diff
@@ -135,6 +135,121 @@ where
...
+            Ok(())
+        }
+    }
+
+    /// A Empty Subscription Handler
+    #[derive(Default)]
+    pub struct EmptySubscriptionHandler;
+
+    impl<Context> SubscriptionStateHandler<Context, std::io::Error> for EmptySubscriptionHandler where
+        Context: Clone + Send + Sync
```
LegNeato, line 249: I don't think you need clone here if it is `Send + Sync` ?

juniper_subscriptions/src/lib.rs
```diff
@@ -135,6 +135,121 @@ where
...
+        /// The subscription is on the end of a operation before sending the GQL_COMPLETE
+        /// message to the client.
+        OnOperationComplete(&'a Context),
+        /// The Subscription is terminating the connection with the client.
+        OnDisconnect(&'a Context),
+    }
+
+    /// Trait based on the SubscriptionServer
+    /// https://www.apollographql.com/docs/graphql-subscriptions/lifecycle-events/
+    pub trait SubscriptionStateHandler<Context, E>
```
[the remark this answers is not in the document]
engylemure: So i was considering this for authorization or log purposes into the subscription, since the websocket protocol doesnt handle authorization the same way as the http does. I saw that in the apollo client it seems that the authorization token or user information is sent as a payload in the ConnectionInit message, so i was trying to think in something that could enable the insertion of user information into the GraphQL Context or something that could enable easier ways to do it.

juniper_subscriptions/src/lib.rs
engylemure: When i was trying to develop the integration between actix web i have found some small questions about some things that seems to be possible with this integration, such as the same websocket connection for a client can have more than a subscription connection with the graphql server in a way that the apollo-client for example reuses the same websocket connection to handle more than a subscription. So i had this question, "each one of these subscription connections should share the same context or they should have isolated ones?".
Since i had considered that they should have isolated ones, i had defined that the Context in the subscriptions in actix-web should be Clone and the same i did for Context in the StateHandler, but thinking about it now it seems that they could share the same Context.

juniper_subscriptions/src/lib.rs
```diff
@@ -19,6 +19,11 @@ use juniper::{
     BoxFuture, ExecutionError, GraphQLError, GraphQLSubscriptionType, GraphQLTypeAsync, Object,
     ScalarValue, SubscriptionConnection, SubscriptionCoordinator, Value, ValuesStream,
 };
+#[cfg(feature = "unstable-ws-subscriptions-transport")]
+/// Utilities for the implementation of subscriptions over websocket
+pub mod ws_util;
+#[cfg(feature = "unstable-ws-subscriptions-transport")]
```
LegNeato, line 25: Not sure, but it isn't just subscriptions, it is all GraphQL over websockets right?

juniper_subscriptions/src/lib.rs
```diff
@@ -19,6 +19,11 @@ use juniper::{
     BoxFuture, ExecutionError, GraphQLError, GraphQLSubscriptionType, GraphQLTypeAsync, Object,
     ScalarValue, SubscriptionConnection, SubscriptionCoordinator, Value, ValuesStream,
 };
+#[cfg(feature = "unstable-ws-subscriptions-transport")]
+/// Utilities for the implementation of subscriptions over websocket
+pub mod ws_util;
```
LegNeato, line 24: I'd like this to be more descriptive / align more with the proposed spec naming.

juniper_subscriptions/src/lib.rs
engylemure: Yeah, it seems to be, since i didnt found no other protocol specification about graphql over websocket other than https://github.com/apollographql/subscriptions-transport-ws this one, that still to be a work in progress, i doesnt know what else could i use to actually base this.

engylemure:
@LegNeato I moved the implementation of PR #603 subscriptions from juniper_actix here, I'm not sure if that was what you had proposed to me or if I should put this implementation somewhere else.

LegNeato:
Do the changes on latest master supercede this PR?

Issue #586, opened 2020-03-24
graphql_union from tagged union of structs


arlyon:
**Describe the solution you'd like**

It would be cool to be able to define GraphQL unions like so:

```rust
#[derive(GraphQLObject)]
struct Square {
    /// length of one side
    pub length: i32,
}

#[derive(GraphQLObject)]
struct Circle {
    /// radius of the circle
    pub radius: i32,
}

#[juniper::graphql_union]
enum Shape {
    Square(Square),
    Circle(Circle),
}
```

I understand the current implementation is quite similar already, however it would be cool if we could pull the boilerplate into the macro somehouw..

**Describe alternatives you've considered**

This would be an alternative (and more ergonomic) alternative to the current implementation. Here is it in use (imagine a query that returns a `FieldResult<Shape>`):

```graphql
query {
    getShapes {
        ...on Square { length }
        ...on Circle { radius }
    }
}

LegNeato:
Oooh, I like it!

PR #618, opened 2020-04-12
Derive macro for tagged enums (GraphQLUnion)

jmpunkt:
Implements derive macros for tagged enums which are translated into GraphQLUnion types. The following features are provided

- The macro implements GraphQLAsyncType. 
- If all variants of the tagged enum have different types, then the derive macro implements `std::convert::From<T>`  converter for these types.

Fixes: https://github.com/graphql-rust/juniper/issues/586

### DIFF
--- a/docs/book/content/types/unions.md
+++ b/docs/book/content/types/unions.md
@@ -194,9 +194,7 @@ impl Character {
 This example is similar to `Enums (Impl)`. To successfully use the
 derive macro, ensure that each variant of the enum has a different
 type. Since each variant is different, the device macro provides
-`std::convert::Into<T>` converter for each _used_ variant. Fields
-which are skipped by using `#[graphql(skip)]` do not have a
-`std::convert::Into<T>` converter.
+`std::convert::Into<T>` converter for each variant.
 
 ```rust
 #[derive(juniper::GraphQLObject)]
--- a/docs/book/content/types/unions.md
+++ b/docs/book/content/types/unions.md
@@ -3,10 +3,11 @@
 From a server's point of view, GraphQL unions are similar to interfaces: the
 only exception is that they don't contain fields on their own.
 
-In Juniper, the `graphql_union!` has identical syntax to the [interface
-macro](interfaces.md), but does not support defining fields. Therefore, the same
-considerations about using traits, placeholder types, or enums still apply to
-unions.
+In Juniper, the `graphql_union!` has identical syntax to the
+[interface macro](interfaces.md), but does not support defining
+fields. Therefore, the same considerations about using traits,
+placeholder types, or enums still apply to unions. For simple
+situations, Juniper provides `#[derive(GraphQLUnion)]` for enums.
 
 If we look at the same examples as in the interfaces chapter, we see the
 similarities and the tradeoffs:
@@ -154,7 +155,7 @@ impl GraphQLUnion for Character {
 # fn main() {}
 ```
 
-## Enums
+## Enums (Impl)
 
 ```rust
 #[derive(juniper::GraphQLObject)]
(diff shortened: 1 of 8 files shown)


juniper/CHANGELOG.md
```diff
@@ -20,6 +20,11 @@ See [#419](https://github.com/graphql-rust/juniper/pull/419).
 
 See [#569](https://github.com/graphql-rust/juniper/pull/569).
 
+- GraphQLUnion derive support ("#[derive(GraphqQLUnion)]")
+  - implements GraphQLAsyncType
+
+See [#](https://github.com/graphql-rust/juniper/pull/).
```
LegNeato, line 26: Nit: We can fill this out now 😃

juniper_codegen/src/derive_union.rs
```diff
@@ -0,0 +1,99 @@
...
+    // Parse attributes.
+    let attrs = match util::ObjectAttributes::from_attrs(&ast.attrs) {
+        Ok(a) => a,
+        Err(e) => {
+            panic!("Invalid #[graphql(...)] attribute for enum: {}", e);
+        }
+    };
+
+    if !attrs.interfaces.is_empty() {
+        panic!("#[derive(GraphlQLUnion)] does not support interfaces");
```
LegNeato, line 24:
```suggestion
        panic!("#[derive(GraphQLUnion)] does not support interfaces");
```

juniper_codegen/src/derive_union.rs
```diff
@@ -0,0 +1,99 @@
...
+            let _type = match field.fields {
+                Fields::Unnamed(inner) => {
+                    let mut iter = inner.unnamed.iter();
+                    let first = match iter.next() {
+                        Some(val) => val,
+                        None => unreachable!(),
+                    };
+
+                    if iter.next().is_some() {
+                        panic!("#[derive(GraphlQLUnion)] all members must be unnamed with a single element e.g. Some(T)");
```
LegNeato, line 62:
```suggestion
                        panic!("#[derive(GraphQLUnion)] all members must be unnamed with a single element e.g. Some(T)");
```

juniper_codegen/src/derive_union.rs
```diff
@@ -0,0 +1,99 @@
...
+                    }
+
+                    first.ty.clone()
+                }
+                _ => panic!("#[derive(GraphlQLObject)] all fields of the enum must be unnamed"),
```
LegNeato, line 67:
```suggestion
                _ => panic!("#[derive(GraphQLUnion)] all fields of the enum must be unnamed"),
```

jmpunkt:
Thanks for the review. Everything you mentioned is fixed. During the testing i encountered some issues. Therefore, skip and documentation annotations are disabled for enumeration fields. Here are some thoughts about these decisions and the implementation. If something is unclear or wrong, let me know.

- Documentation of variations is not possible in GraphQL? (not 100% sure)
- Skipping fields works in general but throws panics during runtime. For example, if a query returns a skipped variant of the enum. The derive macro should provide additional safety (compared to the other union implementation). If the user requires this behavior he can fallback to the other union implementation.
- Currently the std::convert::Into<T> ensures that each variant has a unique type. Due to the GraphQL specification this must be done? (not 100% sure)

docs/book/content/types/unions.md
```diff
@@ -187,3 +188,34 @@ impl Character {
...
 # fn main() {}
 ```
+
+## Enums (Derive)
+
+This example is similar to `Enums (Impl)`. To successfully use the
+derive macro, ensure that each variant of the enum has a different
+type. Since each variant is different, the device macro provides
+`std::convert::Into<T>` converter for each _used_ variant. Fields
+which are skipped by using `#[graphql(skip)]` do not have a
```
LegNeato, line 198: This is no longer true, correct? It appears below you explicitly forbid skip for this (which feels right).

jmpunkt:
This PR should be ready to merge. Applied your feedback and added some minor tweaks.

LegNeato:
Awesome, thank you again!

PR #628, opened 2020-04-21
Added a normalization of subscriptions endpoint in graphiql_source

engylemure:
The objective here is to make the playground and graphiql have the same behavior with the endpoint arguments.


engylemure:
@tyranron Sure

Issue #630, opened 2020-04-24
GraphQLObject generates not specified code


jmpunkt:
**Describe the bug**
The [GraphQL Specification](https://spec.graphql.org/June2018/#sec-Objects) states that 

> 1. An Object type must define one or more fields.

Juniper allows generating objects without fields.

**To Reproduce**
The issue can be reproduced with both derive and impl macro.

```rust
#[derive(juniper::GraphQLObject)]
pub struct Query {}
```

**Expected behavior**
This should not compile.

Issue #627, opened 2020-04-20
Union behavior is different to GraphQL spec


jmpunkt:
**Describe the bug**
Juniper allows creating GraphQL unions with non-objects as variants. The GraphQL spec does not allow to create such unions. It is possible to compile the code, but the user receives and error messages on the playground website, like the following:
```json
{
  "message": "Expected String to be a GraphQL Object type.",
  "stack": "..."
}
```

**To Reproduce**
```rust
enum Character {
    Human(String),
}

#[juniper::graphql_union]
#[graphql(Scalar = juniper::DefaultScalarValue, Context = (), noasync)]
impl Character {
    fn resolve(&self) {
        match self {
            String => { match *self { Character::Human(ref h) => Some(h) } },
        }
    }
}

#[derive(juniper::GraphQLObject)]
#[graphql(Scalar = juniper::DefaultScalarValue,Context = (), noasync)]
pub struct Query {
    test: Character
}
```

Notice that there is no difference between using the `impl` macro or the derive macro for the union type. Both compile and create the same error.

**Expected behavior**
This should not compile. In the best case the proc macro should point out that `String` is not an GraphQL object.

jmpunkt:
I have notice that other types also do not follow the specification correctly. For example, objects (impl macro) can be empty which is not allowed by the specification.

The easiest way to prevent this, is defining failing code generation tests cases. A quick search suggests to use something like [trybuild](https://docs.rs/trybuild/1.0.25/trybuild/) to test Juniper's proc macros.

I would like to know how you(@LegNeato) think about this?

LegNeato:
For empty objects, we should be able to handle that in the proc macro and fail the compile. Please file a different issue!

For requiring a GraphQL object for unions, that will be harder. I like the idea of generating a test. We might also be able to generate a line that only compiles if it is an object but I am not sure. Right now it seems to require GraphQLType, we could have a marker trait just for objects and use trait bounds to enforce the Union's members implement it.

Issue #435, opened 2019-10-09
Use https://crates.io/crates/proc-macro-error


LegNeato:
It would be good to use https://crates.io/crates/proc-macro-error to improve error reporting for proc macros.

Basically we should look everywhere we `panic()` in our proc macros and add span info.

davidpdrsn:
Just wondering, how is that different from [`syn::Error`](https://docs.rs/syn/1.0.5/syn/struct.Error.html)?

davidpdrsn:
Ah I see it mentioned in the readme further down. Nvm

davidpdrsn:
https://github.com/graphql-rust/juniper/pull/441 lays the ground work for this. Once that is merged I’ll update the rest of code gen.

LegNeato:
This is done.

PR #631, opened 2020-04-24
Ensure Specification(June 2018) Compliance

jmpunkt:
# Motivation
Juniper currently produces code which is not allowed by the GraphQL specification. To ensure that these errors are fixed, a test suite is required. The test suite ensures that configurations which are not compliant to the specification, can not successfully compile.

# Implementation
Each GraphQL type has a `Type Validation` chapter. Therefore, each of the constraints must be checked. An isolated Rust source snippet is created for each of these constraints. An external library compiles these snippets. The library ensures that the output of the compilation process is equal to an expected output. With the process it is ensured that these problems can not reoccur _silently_   in the future.

# Goals (Progress)
- [ ] Provider better error messages
- [ ] Ensure that everything is specification conform
- [ ] Implement GraphQLTypeAsync for missing parts (required for the test suite)

# Related
- https://github.com/graphql-rust/juniper/issues/630
- https://github.com/graphql-rust/juniper/issues/627


jmpunkt:
Before I continue, I would like to get some feedback for the implemented error handling (@LegNeato). I wonder if the current design is acceptable or are there any problems?

#### Why?
A change is preferable because panics messages are not precise. Instead `syn::Error` allows specifying a context for an error. The following two errors are a result of that.

```
error: #[derive(GraphQLUnion)] expects at least one field/variant
 --> $DIR/derive_no_fields.rs:2:1
  |
2 | enum Character {}
  | ^^^^^^^^^^^^^^^^^
```
```
error: #[derive(GraphQLObject)] does not allow multiple fields/variants with the same name. There is at least one other field with the same name `test`
 --> $DIR/derive_fields_unique.rs:4:5
  |
4 | /     #[graphql(name = "test")]
5 | |     test2: String,
  | |_________________^
```
Since there are errors shared between the different implementations, there are shortcuts for shared errors (including no duplicates, not empty,  etc.). Errors which are unique are also propagated through the error framework. 

To let the error generator know what kind of macro is used, all macros must provide a context at the beginning. This context is used to generate prefixes like `#[derive(GraphQLObject)]`. Maybe this is not required at all since the current errors are pointing directly to the error.

LegNeato:
This is perfect and very close to what I was planning on doing when I got some time. Thank you!

I wonder if instead of referring to the derive we refer to "GraphQL" in general as most of these aren't requirement of the derive but instead a spec requirement.

So 

```rust
error: #[derive(GraphQLObject)] does not allow multiple fields/variants with the same name. There is at least one other field with the same name `test`
 --> $DIR/derive_fields_unique.rs:4:5
  ```

Becomes

```rust
error: GraphQL objects must not contain multiple fields/variants with the same name. There is at least one other field with the same name `test`
 --> $DIR/derive_fields_unique.rs:4:5
  |
```

What do you think?

LegNeato:
And maybe an optional link to the spec section?

jmpunkt:
Thanks for the response. That is a good idea.

I wonder if we should use https://docs.rs/proc-macro-error/1.0.2/proc_macro_error/struct.Diagnostic.html for errors. This allows to specify `notes` and `help` information. The underlying mechanism is not on stable yet. Therefore, not everything is supported. Information about the used GraphQL specification could be put in a `note`.

LegNeato:
That sounds like a good plan, we already were thinking about it:

https://github.com/graphql-rust/juniper/issues/435

jmpunkt:
The proc-macro-error crate is nice to use. With your previous feedback, an error looks like the following.

```
error: GraphQL object expects at least one field
  --> juniper_benchmarks/src/lib.rs:97:1
   |
97 | impl Mutation {}
   | ^^^^^^^^^^^^^^^^
   |
   = note: https://spec.graphql.org/June2018/#sec-Objects
```

As you can see there are errors are located in the benchmark module which tries to implement empty objects. **What should be done here? Remove the non-compiling snippets?**

The current implementation wraps all parsed attributes into a `SpanContainer<T>` which contains additional `Span` information for that attribute. This is required because all these information are lost during the parsing. Not sure if the `SpanContainer<T>` is the best approach. Maybe you have a better idea.

LegNeato:
I think for benches it is just an oversight. We can use `EmptyMutation` and `EmptySubscription` to fix those (and probably the "no fields on object" error should suggest those).

For `User`, we can just remove the empty proc macro and use the derive instead.

LegNeato:
`SpanContainer` sounds like a good solution for now. The error message looks great!

jmpunkt:
@LegNeato maybe you can resolve this question. Resolver functions for objects, e.g., `test(&self) -> String` can have various signatures. What is the smallest possible variant of this construct. The reason for this question is that after some refactoring in favor of the error messages, I found ~10 resolver functions which do not use `&self`. So which of the following functions must compile?

- `fn test() -> String` 
- `fn test(&self) -> String` 
- `fn test(&self, context: &C) -> String` 
- `fn test(context: &C) -> String`

LegNeato:
Oh weird! AFAIK we should have tests for this. The two middle should compile, the first and last should not. &self is required and context is optional (as is the executor) but if it is there it needs to be a reference.

LegNeato:
Ah, what I remember seeing is for the resolve function:

https://github.com/graphql-rust/juniper/blob/master/juniper_codegen/src/util/parse_impl.rs#L74

It looks like we don't require `&self`:

https://github.com/graphql-rust/juniper/blob/master/integration_tests/juniper_tests/src/codegen/derive_object.rs#L85


I think it makes sense to always require `&self` even though it is a breaking change, do you agree? I'm not sure what is more idiomatic.

jmpunkt:
Having the requirement for `&self` for each functions increase the consistency. For example,
https://github.com/graphql-rust/juniper/blob/master/integration_tests/async_await/src/main.rs sometimes has a `&self`. Therefore, introducing a `&self` makes it clear that the function requires an underlying object. `fn() -> T` could suggest that there are no requirements for the function.

An obvious downside is the breaking change which may result in changing a lot of functions. However, currently I tend to introduce the breaking but I am not fully convinced.

LegNeato:
This release is already breaking a ton so let's just do it.

jmpunkt:
This PR is getting a little bit larger. I had to refactor the GraphQLInputObject which allows applying my error scheme. Now the GraphQLInputObject has a similar form like GraphQLObject and the others. Since GraphQLInputObject was the last construct which used a different argument parse, it now does not allow using string scalars like "DefaultScalarValue". I mentioned this breaking change in the changelog.

One last thing, since GraphQLInputObject is missing the async trait, some of the tests fail with a different error message. So there are two possibilities:

- add GraphQLTypeAsync in this PR
- disabled tests until it is implemented

juniper_codegen/src/derive_object.rs
```diff
@@ -1,84 +1,135 @@
...
+            ) {
+                Ok(attrs) => attrs,
+                Err(e) => {
+                    proc_macro_error::emit_error!(e);
+                    return None;
+                }
+            };
+
+            if field_attrs.skip.is_some() {
+                None
```
LegNeato, line 48: This can be made an early return to prevent rightward drift.

jmpunkt:
Thanks for your review.

Additionally, I added a marker trait to detect non-object within the context of unions. Now it is not possible to use non-objects as values for unions.

Besides your recommendations, I added the `__` constraint for all fields, types and arguments. The only exception are internal objects, types, etc.

Finally there is support to rename an argument inside an object impl block. This was required to test the `__` for arguments. Notice that the `__` tests can only be tested with `#[graphql]` annotations. The function `to_camel_case` function can not return a name with a `__` prefix.

juniper_codegen/src/util/duplicate.rs
```diff
@@ -0,0 +1,46 @@
...
+    pub spanned: Vec<T>,
+}
+
+impl<T> Duplicate<T> {
+    pub fn find_by_key<'a, F>(items: &'a [T], name: F) -> Option<Vec<Duplicate<&'a T>>>
+    where
+        T: 'a,
+        F: Fn(&'a T) -> &'a str,
+    {
+        let mut mapping: HashMap<&str, Vec<&T>> = HashMap::with_capacity(items.len());
```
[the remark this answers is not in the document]
jmpunkt: The original idea was collecting all fields with the same name and highlight them in a single error message. This is possible with the `Diagnostics` API (not stable yet). I think we should keep it this and fix the error message if the API is stable.

jmpunkt:
I notice that interface testing is not possible at the moment, since it is still a declarative macro. Enforcing the requirements of the specification seems not possible. I disabled the test. After the macro is replaced the test should be re-enabled.

Maybe you can take a close look at the GraphQLTypeAsync impl for input objects. The input object now implements the async type. No functions are required to implement for the async trait since input objects are constructed by the other two traits (ToInputValue, FromInputValue)?!

jmpunkt:
The only way to switch the expected output is to have two crates with different expected output. But testing only on nightly should be fine. This way we notice possible changes since. Stable does not implement most of the features, thus the output should not change at all (until the `Diagnostic` API lands).

LegNeato:
Excellent! I think we just need to record the output again and we're good to merge!

juniper_codegen/src/util/duplicate.rs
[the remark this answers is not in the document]
LegNeato: Ah, makes sense

jmpunkt:
Implementing the constraints that arguments have the property `IsInputType(arg) == True` is bit tricky. I use a marker trait `IsInputType` which has an arbitrary method. For each  argument, 
`<#arg_type as IsInputType<#scalar>>::mark()` is generated. The compiler can check if each arg has the `IsInputType` implemented. This seems to work. However, this must be implemented for interfaces as well. Since interfaces are generated by declartive macros, there is no way to express the previous statement.

```rust
        $crate::__juniper_impl_trait!(
            impl<$($scalar)* $(, $lifetimes)* > GraphQLType for $name {
                fn mark() {
                    $(
                        <$return_ty as $crate::marker::IsOutputType<$($scalar)*>::mark();
                    )*
                }
            }
        );
```

The reason why this does not compile is because of the nested `$(...)*`. The only possible fix I see right now is to remove the declarative macro and replace it with a proc macro.

LegNeato:
Yeah, the plan is to remove all macros and replace them with proc macros in this release for just that reason. We just did so a couple weeks ago with `graphql_scalar`.

The declarative macros were from way back from when proc macros were not available on stable.

If you want to skip that check in this PR so we can land it I'm totally fine with that.

Also, I'd love to add you as a maintainer for Juniper, let me know if you would be interested. You're doing a ton of high-quality work.

juniper/src/types/marker.rs
```diff
@@ -11,8 +16,108 @@ use crate::{GraphQLType, ScalarValue};
...
 pub trait GraphQLObjectType<S: ScalarValue>: GraphQLType<S> {
-    #[allow(missing_docs)]
-    fn name(info: &Self::TypeInfo) -> Option<&str> {
-        <Self as GraphQLType<S>>::name(info)
-    }
+    /// An arbitrary function without meaning.
+    ///
+    /// May contain compile timed check logic which ensures that types
+    /// are used correctly according to the GraphQL specification.
+    fn mark() {}
```
LegNeato, line 23: I believe this can be empty and might slightly save on compile times. At least, the implementation of traits on https://doc.rust-lang.org/std/marker/index.html are empty. Could be done in a followup.

jmpunkt: The problem is that I did not found any other way of type checking except for `<T as U>::mark()`. Therefore, we require at least one function inside the trait to ensure that `T` implements `U`. Adding the trait bound as `where` clauses introduces a large overhead in the implementation and would require to specify lifetimes in struct like `Test<'a>`. In most internal structs lifetimes bound inside the struct can be ignored with the current approach.

LegNeato:
Thanks again!

Issue #602, opened 2020-04-04
Composing an error response


biot023:
Hi -- I'm validating an input object to an update mutation.
It would be really great if I could compose the `extensions` object of the `FieldError` to look something like this:

````
{
  "errors": [
    ["title", "Too short"],
    ["title", "Invalid characters"],
    ["description", "Invalid characters"]
  ]
}
````

My attempts to compose a `Value::Object` to match that were (I think) successful but I failed when I tried to substitute it to the `extensions` bit of `FieldError::new`.
Am I misunderstanding what the `extensions` bit is for?
And could someone advise me on the best way to achieve a list of errors the client can parse when composing a `FieldError`?
Thanks for any and all advice,
   Doug.

**Comment by LegNeato** [comment]
Does the example at https://graphql-rust.github.io/juniper/master/types/objects/error_handling.html#structured-errors help?

**Comment by biot023** [comment]
Hi @LegNeato, thanks for the reply.
Those docs don't really help, sorry. What I need to be able to do is to build up an error based on whether various input fields are deemed valid or not.
So something like:

````
if is_invalid(title) {
    errors.push(("title", "Invalid characters"));
}
if is_invalid(description) {
    errors.push(("description", "Invalid characters"));
}
// Now build the `FieldError` `extensions` with the errors `Vec<(&str, &str)>`.
````

The examples in the docs just show you how to build a static structure, I need to build one procedurally.
I have an ugly hack of just listing them in the error description for the time being but I'll really want to fix that before I open our API up!

**Comment by LegNeato** [comment]
You could do something like:

```rust
enum CustomInputError {
    TooShort(i32),
}

impl juniper::IntoFieldError for CustomError {
    fn into_field_error(self) -> juniper::FieldError {
        match self {
            CustomError::TooShort(x) => juniper::FieldError::new(
                "Too short",
                graphql_value!({
                    "min_length": x
                }),
            ),
        }
    }
}

struct Example {}

#[juniper::graphql_object]
impl Example {
    fn whatever(&self, title: MyTitle) -> Result<bool, CustomError> {
        if is_too_short(title) {
            Err(CustomError::TooShort(5))
        }
        // Do your logic or additional checks that return other errors.

        // Success case.
        Ok(true)
    }
}
```

As is standard in GraphQL and [mentioned in the book](https://graphql-rust.github.io/juniper/master/types/objects/error_handling.html#structured-errors :
[...]

PR #634, opened 2020-04-26
Documented Alternative Error Handling

jmpunkt:
Documented the process in https://github.com/graphql-rust/juniper/issues/602.

Would be nice if someone could proof read the additions. Please mark sections which are not **immediately** clear or have weird grammar.

### DIFF
--- a/docs/book/content/types/objects/error_handling.md
+++ b/docs/book/content/types/objects/error_handling.md
@@ -251,11 +251,12 @@ impl Mutation {
         if errors.is_empty() {
             GraphQLResult::Ok(Item { name, quantity })
         } else {
-            GraphQLResult::Err(errors)
+            GraphQLResult::Err(ValidationErrors { errors })
         }
     }
 }
 
+# fn main() {}
 ```
 
 Each function may have a different return type and depending on the input
@@ -306,7 +307,7 @@ pub struct Item {
 }
 
 #[derive(juniper::GraphQLObject)]
-pub struct ValidationErrorItem {
+pub struct ValidationError {
     name: Option<String>,
     quantity: Option<String>,
 }
@@ -314,7 +315,7 @@ pub struct ValidationErrorItem {
 #[derive(juniper::GraphQLUnion)]
 pub enum GraphQLResult {
     Ok(Item),
-    Err(ValidationErrorItem),
+    Err(ValidationError),
 }
 
 pub struct Mutation;
@@ -322,27 +323,28 @@ pub struct Mutation;
 #[juniper::graphql_object]
 impl Mutation {
     fn addItem(&self, name: String, quantity: i32) -> GraphQLResult {
-        let mut error = ValidationErrorItem {
+        let mut error = ValidationError {
             name: None,
             quantity: None,
         };
 
         if !(10 <= name.len() && name.len() <= 100) {
-            error.name(Some("between 10 and 100".to_string()));
+            error.name = Some("between 10 and 100".to_string());
         }
 
         if !(1 <= quantity && quantity <= 10) {
-            error.quantity(Some("between 1 and 10".to_string()));
+            error.quantity = Some("between 1 and 10".to_string());
         }
 
-        if error.name.is_none() && error.field.is_none() {
+        if error.name.is_none() && error.quantity.is_none() {
             GraphQLResult::Ok(Item { name, quantity })
         } else {
-            GraphQLResult::Err(errors)
+            GraphQLResult::Err(error)
         }
     }
 }
 
+# fn main() {}
 ```
 
 ```graphql
@@ -375,6 +377,8 @@ and would generate errors. Since it is not common for the database to
 fail, the corresponding error is returned as a critical error:
 
 ```rust
+# #[macro_use] extern crate juniper;
+
 #[derive(juniper::GraphQLObject)]
 pub struct Item {
     name: String,
@@ -421,21 +425,22 @@ impl Mutation {
         };
 
         if !(10 <= name.len() && name.len() <= 100) {
-            error.name(Some("between 10 and 100".to_string()));
+            error.name = Some("between 10 and 100".to_string());
         }
 
         if !(1 <= quantity && quantity <= 10) {
-            error.quantity(Some("between 1 and 10".to_string()));
+            error.quantity = Some("between 1 and 10".to_string());
         }
 
-        if error.name.is_none() && error.field.is_none() {
+        if error.name.is_none() && error.quantity.is_none() {
             Ok(GraphQLResult::Ok(Item { name, quantity }))
         } else {
-            Ok(GraphQLResult::Err(errors))
+            Ok(GraphQLResult::Err(error))
         }
     }
 }
 
+# fn main() {}
 ```
 
 ## Additional Material
--- a/docs/book/content/types/objects/error_handling.md
+++ b/docs/book/content/types/objects/error_handling.md
@@ -1,5 +1,14 @@
 # Error handling
 
+Error handling in GraphQL can be done in multiple ways. In the
+following two different error handling models are discussed: field
+results and GraphQL schema backed errors. Each approach has its
+advantages. Choosing the right error handling method depends on the
+requirements of the application. Investigating in both approaches is
+beneficial.
+
+## Field Results
+
 Rust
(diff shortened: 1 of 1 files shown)


docs/book/content/types/objects/error_handling.md
```diff
@@ -169,3 +178,293 @@ The specified structured error information is included in the [`extensions`](htt
...
+similar to a `FieldError`. The whole query is aborted and nothing can
+be extracted (except for error related information). Not all errors
+require this strict handling. Recoverable or partial errors can be put
+into the GraphQL scheme. This way the client knows what can usually
+happen. To implement this approach all errors must be partitioned into
+the two error classes. The difference to the model in Rust is that
+critical means, e.g., database error. On the other hand, recoverable
+errors can be fixed by the user itself. If the user changes the input,
+then the error might get away. The prime example for recoverable
+errors is input validation.
```
LegNeato, line 194:
```suggestion
be extracted (except for error related information).

Not all errors require this strict handling. Recoverable or partial errors can be put
into the GraphQL schema so the client can intelligently handle them.

To implement this approach, all errors must be partitioned into two error classes:

* Critical errors that cannot be fixed by the user (e.g. a database error).
* Recoverable errors that can be fixed by the user (e.g. invalid input data).
```

docs/book/content/types/objects/error_handling.md
```diff
@@ -169,3 +178,293 @@ The specified structured error information is included in the [`extensions`](htt
...
+
+The error class specifies where the error is seen. Critical errors are
+returned as `FieldErrors` (from the previous section). Non-critical
+errors are part of the GraphQL scheme. Similar to Rust, GraphQL allows
+to have a similar error model with unions (see Unions).
```
LegNeato, line 199:
```suggestion
Critical errors are returned from resolvers as `FieldErrors` (from the previous section). Non-critical errors are part of the GraphQL schema and can be handled gracefully by clients. Similar to Rust, GraphQL allows similar error models with unions (see Unions).
```

docs/book/content/types/objects/error_handling.md
```diff
@@ -169,3 +178,293 @@ The specified structured error information is included in the [`extensions`](htt
...
+
+### Example Input Validation (simple)
+
+In this example, a basic input validation is implemented with GraphQL
+types. Strings are used to identify the problematic field name. Errors
+for a particular field are also returned as a string. In this example,
+the string contains a localized error message. However, it is also
+possible to return a unique string identifier.
```
LegNeato, line 207:
```suggestion
In this example, basic input validation is implemented with GraphQL
types. Strings are used to identify the problematic field name. Errors
for a particular field are also returned as a string. In this example
the string contains a server-side localized error message. However, it is also
possible to return a unique string identifier and have the client present a localized string to the user.
```

docs/book/content/types/objects/error_handling.md
```diff
@@ -169,3 +178,293 @@ The specified structured error information is included in the [`extensions`](htt
...
+        }
+    }
+}
+
+```
+
+Each function may have a different return type. Depending on the input
+parameters a new result type is required. For example, adding a user
+require a new result type which contains the variant `Ok(User)`
+instead of `Ok(Item)`.
```
LegNeato, line 267:
```suggestion
Each function may have a different return type and depending on the input
parameters a new result type is required. For example, adding a user
requires a new result type which contains the variant `Ok(User)`
instead of `Ok(Item)`.
```

docs/book/content/types/objects/error_handling.md
```diff
@@ -169,3 +178,293 @@ The specified structured error information is included in the [`extensions`](htt
...
+
+Finally it is possible to send a mutation request and handle the
+result. The following example query shows how to handle the result.
```
LegNeato, line 270:
```suggestion
The client can send a mutation request and handle the
resulting errors as shown in the following example:
```

docs/book/content/types/objects/error_handling.md
```diff
@@ -169,3 +178,293 @@ The specified structured error information is included in the [`extensions`](htt
...
+        }
+      }
+    }
+  }
+}
+```
+
+A useful side effect of this approach is to have partially successful
+paths. Therefore, a if one of multiple paths fails, the result of the
+successful paths are not discarded.
```
LegNeato, line 292:
```suggestion
A useful side effect of this approach is to have partially successful
queries or mutations. If one resolver fails, the results of the
successful resolvers are not discarded.
```

docs/book/content/types/objects/error_handling.md
```diff
@@ -169,3 +178,293 @@ The specified structured error information is included in the [`extensions`](htt
...
+### Example Input Validation (complex)
+
+Instead of using strings do propagated error, it is possible to use
+GraphQL's type system to describe the errors more precisely. For each
+failable input variable a field in a GraphQL object is created. The
+field is set if the validation for that particular field fails. Notice
+that some kind of code generation reduces the unnecessary work. The
+amount of types which are required is significant larger than
+before. Each functions has their custom `ValidationResult` which
+contains only fields provided by the function.
```
LegNeato, line 303:
```suggestion
Instead of using strings to propagate errors, it is possible to use
GraphQL's type system to describe the errors more precisely.

For each fallible input variable a field in a GraphQL object is created. The
field is set if the validation for that particular field fails. You will likely want some kind of code generation to reduce repetition as the number of types required is significantly larger than
before. Each resolver function has a custom `ValidationResult` which
contains only fields provided by the function.
```

docs/book/content/types/objects/error_handling.md
```diff
@@ -169,3 +178,293 @@ The specified structured error information is included in the [`extensions`](htt
...
+        name
+        quantity
+      }
+    }
+  }
+}
+```
+
+Expected errors are handled directly inside the query. Even more, all
+non-critical errors are known in advance.
```
LegNeato, line 369:
```suggestion
Expected errors are handled directly inside the query. Additionally, all
non-critical errors are known in advance by both the server and the client.
```

docs/book/content/types/objects/error_handling.md
```diff
@@ -169,3 +178,293 @@ The specified structured error information is included in the [`extensions`](htt
...
+
+### Example Input Validation (complex with critical error)
+
+So far only non-critical errors occurred in the examples. Providing
+errors inside the GraphQL schema still allows to return critical
+errors. In the following example, a theoretical database could fail
+and would generate errors. Since it is not common for the database to
+fail, the corresponding error is returned as a critical error.
```
LegNeato, line 377:
```suggestion
Our examples so far have only included non-critical errors. Providing
errors inside the GraphQL schema still allows you to return unexpected critical
errors when they occur.

In the following example, a theoretical database could fail
and would generate errors. Since it is not common for the database to
fail, the corresponding error is returned as a critical error:
```

docs/book/content/types/objects/error_handling.md
```diff
@@ -169,3 +178,293 @@ The specified structured error information is included in the [`extensions`](htt
...
+
+If your not convinced yet, then consider the following example. An
+application is available in different spoken languages. Localization
+should be handled on the frontend exclusively. The frontend requires
+detailed information about the error. For example, instead of return
+"string to short" in different language, the expected limits of that
+string (min length, max length) is returned. The frontend uses these
+information to build an appropriate message. Enforcing this requires
+are large amount of types, but hopefully increase the quality of the
+software.
```
LegNeato, line 470:
```suggestion
The first approach discussed above--where every error is a critical error defined by `FieldResult` --is easier to implement. However, the client does not know what errors may occur and must instead infer what happened from the error string. This is brittle and could change over time due to either the client or server changing. Therefore, extensive integration testing between the client and server is required to maintain the implicit contract between the two.

Encoding non-critical errors in the GraphQL schema makes the contract between the client and the server explicit. This allows the client to understand and handle these errors correctly and the server to know when changes are potentially breaking clients. However, encoding this error information into the GraphQL schema requires additional code and up-front definition of non-critical errors.
```

jmpunkt:
Thanks for detailed review. Now this PR should be in good shape and ready to merge.

LegNeato:
Thank you again! This will really help folks.

PR #635, opened 2020-04-27
Fix deprecated comments

Byeongjee:
Hi, this is my first contribution to juniper.

I think some comments are deprecated.
It seems that `build_object_type`, `build_union_type`, ... methods returned some builders in the past.
Now all such methods return `*Meta` type directly, but the comments are not changed yet.


LegNeato:
Thank you so much, totally missed that! I appreciate the contribution, let me know if you need any help with the codebase! 🍻

PR #636, opened 2020-04-27
Add meta type builders with custom parse functions

Byeongjee:
While using juniper dynamic schema in my project, I found a small problem.
Currently, when `Registry` builds a `ScalarMeta`, it requires the input type to implement `FromInputValue` and `ParseScalarValue` and makes use of the traits.
Therefore, when we create a type dynamically, we cannot use the dynamic type information given by `TypeInfo` to implement the traits.

That's why I added builder methods to build meta types with `try_parse_fn` and `parse_fn` as arguments.
Maybe it's better to add `TypeInfo` argument to those traits, but I'm not sure about this because I'm not familiar with the juniper codebase yet.

If you think the direction of change is fine but there is a better solution, please let me know. I'll fix it.


juniper/src/schema/meta.rs
```diff
@@ -582,6 +612,21 @@ where {
         }
     }
 
+    /// Build a new scalar type metadata with the specified name, input fields, and `try_parse_fn`
```
LegNeato, line 615: Input object

juniper/src/schema/meta.rs
```diff
@@ -499,6 +514,21 @@ where
         }
     }
 
+    /// Build a new scalar type metadata with the specified name, possible values, and `try_parse_fn`
```
LegNeato, line 517: Enum

LegNeato:
Thanks for the PR! Would love to have a test that uses all these so we know we do not break them in the future.

PR #638, opened 2020-04-27
Avoid unnecessary Vec allocation when executing GraphQLBatchRequest

tyranron:
This is a minor improvement which eliminates unnecessary `Vec` collection of `GraphQLRequest`s before feeding them into `future::join_all()`, which is OK with accepting `Iterator`s anyway.

Additionally:
- Makes `GraphQLBatchRequest`/`GraphQLBatchResponse` code a bit more laconic.


LegNeato:
Thank you!

PR #640, opened 2020-04-28
Remove direct usage of `serde_derive` crate

tyranron:
Another minor code improvement to remove direct usages of `serde_derive` crate (which is available via `derive` feature of `serde` crate).

Also, removes unnecessary `serde` deps in `Cargo.toml` of integration crates.


LegNeato:
🍻

PR #648, opened 2020-05-01
Exclude basic subscriptions example from workspace

mihai-dinculescu:
Tiny fix required in order to get the basic_subscriptions example to run.


LegNeato:
Thanks!

Issue #600, opened 2020-04-03
 Unable to add custom attributes on fields


Dispersia:
**Is your feature request related to a problem? Please describe.**
Unable to add custom proc_macro_attribute's to fields, so can't do any rewrite logic

**Describe the solution you'd like**
juniper::graphql_object to keep custom attributes, not just the graphql(description/deprecation) ones

**Describe alternatives you've considered**
Having the proc macro go over the entire impl, then defining at the top which fields get rewritten

**Additional context**
I'm not sure what happens to the field methods, so this may not be possible if it would be transformed in some way. I started going through the codegen, but it's quite a bit at first glance so forgive me if it's not possible.

 Example of how some things I would like to do:

```
struct Query;

#[juniper::graphql_object]
impl Query {
    #[authorize(Policy = policies::Whatever)]
    async fn foo() -> FieldResult<foo> {}
}
```
and being able to have a partial method rewrite that automatically handles 401/403 type responses

LegNeato:
Yep, we should totally do this.

EituKo:
Hi, we'd like to implement this too!
We already discussed it [here](https://github.com/graphql-rust/juniper/issues/447) and started looking at how it could be done but didn't go much further, it would be great if we could eventually collaborate to make this possible.

jmpunkt:
Maybe I'm wrong, but I don not think it is possible to append proc macro attributes. However, the most common use-case related to this, is some kind of `Guard` for a field. A `Guard` allows to intercept the resolution of the type. If the guard fails, the underlying type is not resolved. The following example shows how to define group checks:

```rust
trait GraphQLGuard<S: juniper::ScalarValue> {
    type Context: juniper::Context;
    type Error: juniper::IntoFieldError<S>;

    fn guard(&self, ctx: &Self::Context) -> Result<(), Self::Error>;
}

struct Group(String);

impl <S: juniper::ScalarValue> GraphQLGuard<S> for Group {
    type Context = ();
    type Error = juniper::FieldError<S>;

    fn guard(&self, ctx: &Self::Context) -> Result<(), Self::Error> {
        use juniper::graphql_value;

        match ctx.get_group() {
            Some(group) if self.0 == group => Ok(()),
            _ => juniper::FieldError::new(
                "User is not part of the group",
                graphql_value!({ "policy_error": "Wrong group" })
            ),
        }
    }
}

struct Test;

#[juniper::graphql_object]
impl Test {
[...]

Issue #447, opened 2019-10-31
Is it possible to add authorization system on Juniper


AurelienFT:
Hello,

I have my GraphQL API with Juniper and I want to add permissions to call a graphql_object/to get his result/to get some fields of the result.

With Type-GraphQL we can do something like that : 
```
class MyObject {
  @Field()
  publicField: string;

  @Authorized()
  @Field()
  authorizedField: string;

  @Authorized("ADMIN")
  @Field()
  adminField: string;

  @Authorized(["ADMIN", "MODERATOR"])
  @Field({ nullable: true })
  hiddenField?: string;
}
```

In Rust with Juniper I'll like to know if there is a way to do it ? I didn't find anything about it so I think about a good syntax to make it easy to use. There is something similar to this in Juniper ? :
```
impl Queries {
    #[authorized_call("ADMIN")]
    #[authorized_return(readPermissions)]
    fn user(
        context: &Context,
        id: Option<String>,
        email: Option<String>,
        username: Option<String>,
        role: Option<String>,
    ) -> FieldResult<Option<User>> {
                Ok(find_user(context, id, email, username, role))
    }
}
```

Thank you for your time

theduke:
This is not currently possible, but definitely desired.

The async migration will take precedence for now, but this would be terribly hard to implement. 

We could even implement this in a relatively type-safe way.

Currently your best bet is something more manual:

```rust
struct Context {
  user: Option<User>,
}

enum Permission {
  ReadPost,
  ReadOwnPost,
  WritePost,
  ...
}

fn authorize(ctx: &Context, permissions: &'static [Permission]) -> Result<(), FieldError> {
  // ...
}

#[juniper::object(Context = Context)]
impl Query {
  fn post(context: &Context, id: u64) -> Result<(), FieldError> {
    authorize(context, &[Permission::ReadPost])?;
    ...
  }
}


```

EituKo:
@theduke 
Hi and thank you for your answer!

We really appreciate your alternative but we would like to have a compilation-time, type-safe permissions system with the following characteristics:

- Permissions implementation would be checked at compilation time and generate errors / warning if not implemented or not implemented correctly
- A custom filter could be applied if the return is an iterable type

We started using Rust for this kind of usage and would like to use it to its full potential since we're not able to do something like this with other languages.
We also think that having automated permissions documentation would be a great future feature.

As  @AurelienFT mentioned it earlier, it would look like this:
```rust
impl Query {
[...]

Issue #156, opened 2018-03-12
Directives support?


ghost:
Authentication and other checks can be included in the generated schema using directives. Is it possible to do that juniper right now?

I'm trying to add auth checks for certain mutations using directives but seems impossible in juniper.
Something like this,
https://gist.github.com/LawJolla/132b99179b978cf76eadfd7a6d742b22#file-scratch-graphql
Only other alternative I see is using macros. isAuthenticate!() and hasRole!(role: “MANAGER”) within the mutations.

rokob:
Hey I want this too. It looks like the query parser supports directives, but I don't see a way to add new directives of your own or to handle resolution. I'll give it a go at taking this on. I'll dig into what this might look like and post back here so I don't go too far off the reservation before agreeing on a design.

rokob:
So the "official" js library doesn't support directives other than the two baked in ones, `skip` and `include`, and won't be any time soon, c.f. https://github.com/graphql/graphql-js/issues/41 https://github.com/graphql/graphql-js/issues/446

Thus adding support here would be deviating from the standard a bit and therefore is maybe why this hasn't been added yet.

You can create directives in your schema via something like `root_node.schema.add_directive`
https://github.com/graphql-rust/juniper/blob/master/juniper/src/schema/model.rs#L190

The AST has the notion of directives which is how skip and include work for a query. Adding support for extra field level directives seems to be pretty straight forward, it would just be generalizing skip/include to call into a directive resolver.

However, there are other notions of directive that I have seen, for example https://www.apollographql.com/docs/graphql-tools/schema-directives.html, which are much more complicated. For instance, they have an `@upper` directive which uppercases strings:

```
type Query {
  hello: String @upper
}
```

This would require resolving the field, then calling into some other resolver defined for the directive. Again that specific case isn't too hard. But this quickly gets out of control.

Solving this with an abstraction like the SchemaDirectiveVisitor from Apollo seems like the right solution if this is a feature you'd actually want to add to this library. As this library doesn't really work with the SDL I am not sure this is actually a desired feature here. It probably makes more sense to just write resolvers that handle the necessary logic. If this library was able to parse a SDL document to generate types and hook into resolvers then this might be a worthwhile feature, but as it stands it is kinda antithetical to how this library views type definitions.

AlbertMarashi:
Directives are quite important, especially for building stuff relating to authorisation, scopes & permissions. How can we do this?

MarcoFuykschot:
[...]

PR #650, opened 2020-05-04
Initial support for GraphQL path guards

jmpunkt:
# Overview
Juniper should support authenticated paths in some way. Instead of using external macros, this PR provides trait based guards which are evaluated before the actual field is resolved. If the guard fails, the whole evaluation fails and an `FieldError` is returned. Besides allowing early aborts, it is also possible to change the context. For example, if the user is in a privileged group, then we can exchange the context. The new context contains additional information and it could be ensured that no re-check of the **same** permission happens again. 

## How does the context change work?
Each object defines their context. The guards may require a specific context. If both contexts are compatible, then the guard can be used. The guard also defines how the output context look like, its is still possible to forward the old context. To annotate a field use `#[graphql(Guard = "..")]`. Notice that guards are only support for object fields, at least for now.

## Authentication
With this PR it is possible to define permission systems on top of Juniper. The current testing cases implement a basic role-based authentication system. Take a look at the source to get a feeling how this feature is used.

# Related
- https://github.com/graphql-rust/juniper/issues/600 
- https://github.com/graphql-rust/juniper/issues/447

# Progress
- [x] basic implementation
- [x] combinators (and_then)
- [ ] sync support?
- [ ] tests
- [x] documentation for this features
- [x] documentation for this authentication


jmpunkt:
Not fully done yet. Tests are missing, docs are not perfect, and the book examples do not compile. However, the basic constructs should be established. Would love to hear some initial feedback about this feature.

The only question which is is not solved now is: how to handle non-async. For now I ignored sync code since it should be removed anytime soon. As a result, the guards only support async resolvers and only provide async interface.

LegNeato:
My main concern here (and a lot of the other outstanding PRs) is this is making Juniper more opinionated and less of a library and more of a framework.

jmpunkt:
Not sure about this. The whole proc macros are kind of a framework which uses the underlying library? This PR modifies the code generation in a way which allows to "intercept" the evaluation. Maybe you can expand what you expect, or if such a feature should be possible or not.

The general problem I see right now, is that it is not possible to have custom attribute macros. The order of macros is not specified, thus we can not ensure that their macro is called before. Furthermore, the user must define two procs, like we do, for `impl` blocks and structs. As a result, proc macros can not be used as a plugin architecture for fields. In the end, this means the user has to use the `#[graphl_object]` macro for all structs, if the fields must be must run custom logic before the actual evaluation.

ccbrown:
To quote the Best Practices page on graphql.org for Authorization:

> Enforcing this kind of behavior should happen in the business logic layer.

The page goes on to give examples and details and makes it pretty clear that defining permissions systems on top of Juniper is not recommended.

That being the viewpoint of the GraphQL Foundation (and of myself), I don't think this sort of functionality should be considered.

LegNeato:
Agreed! If Juniper doesn't have the proper hooks, please open an issue on the missing extension / integration point.

Dispersia:
@ccbrown are you able to name another graphql library that doesn't? Apollo Server does, HotChocolate does, graphql-dotnet does, prisma recommends to use apollo servers auth, etc etc.

 Just saying, that spec typically doesn't follow reality, with every company (besides facebook apparently) even versioning their apis: https://developer.github.com/v4/breaking_changes/ .

 Would it be happier if this was a sister package? I would argue this is a basics that comes with most  other libraries. Just giving my experience from using other libraries, if this just isn't wanted that's completely in your right, but it would definitely lead to a lower adoption.

ccbrown:
The most popular / comparable library for Golang, [graphql-go/graphql](https://github.com/graphql-go/graphql), does not have any support for this. The only other popular one I've actually used at scale is [samsarahq/thunder](https://github.com/samsarahq/thunder), which also does not have any support for this. Without looking, my gut says it would not be very hard at all to find more and this probably isn't the kind of situation where lack of the feature is going to be a significant burden to adopters.

This might be reasonable as a sister package, but I would guess the ideal solution here is really to just implement custom directives support (https://github.com/graphql-rust/juniper/issues/156). That would make [Apollo Server-like authentication via custom directives](https://www.apollographql.com/docs/apollo-server/security/authentication/#authorization-via-custom-directives) possible without making Juniper more opinionated.

felipellrocha:
> My main concern here (and a lot of the other outstanding PRs) is this is making Juniper more opinionated

The irony being, that explicitly not supporting auth, Juniper does, indeed, become more opinionated... By definition...

PR #651, opened 2020-05-05
[WIP] Relax graphql_object

c410-f3r:
**WIP**

Relax `graphql_object` procedural macro to include `fn` items provided by external declarative macros.

This will allow the inclusion of definitions declared outside of the main project for more reusable and self-contained parts.

```rust
struct Query;

macro_rules! macro_from_another_project_exporting_items {
    () => {
        pub fn a() -> i32 {
            1
        }

        pub fn b(&self) -> i32 {
            2
        }
    }
}

#[juniper::graphql_object]
impl Query {
    macro_from_another_project_exporting_items!();
} 
```

Couldn't personally find another way to build compositions with `Juniper`. Please, tell me if this approach is too "hacky" to be included upstream.

### DIFF
--- a/integration_tests/juniper_tests/src/codegen/macros.rs
+++ b/integration_tests/juniper_tests/src/codegen/macros.rs
@@ -1,0 +1,18 @@
+struct Query;
+
+macro_rules! macro_from_another_project_exporting_items {
+    () => {
+        pub fn a() {
+            1
+        }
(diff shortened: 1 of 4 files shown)


c410-f3r:
Nah, NVM. Syn is a parsing library, not a engine that expands macros.

Issue #580, opened 2020-03-20
#[juniper::graphql_object] != #[derive(juniper::GraphQLObject)]


olvyko:
Code like this in master branch:
```rust
pub struct Data {
   pub value: String,
}

#[juniper::graphql_object]
impl Data {
    pub fn value(&self) -> String {
        self.value
    }
}

#[derive(juniper::GraphQLObject)]
pub struct User {
   pub username: String,
   pub values: Vec<Data>,
}
```
Produce an error: 
```sh
the trait bound `Data: juniper::types::async_await::GraphQLTypeAsync<__S>` is not satisfied
required because of the requirements on the impl of `juniper::types::async_await::GraphQLTypeAsync<__S>` for `std::vec::Vec<Data>`
required because of the requirements on the impl of `juniper::types::async_await::GraphQLTypeAsync<__S>` for `&std::vec::Vec<Data>`

`match` arms have incompatible types
expected enum `std::result::Result<juniper::value::Value<__S>, juniper::executor::FieldError<__S>>`
   found enum `std::result::Result<_, juniper::executor::FieldError<juniper::value::scalar::DefaultScalarValue>>`
type parameters must be constrained to match other types
for more information, visit https://doc.rust-lang.org/book/ch10-02-traits.html#traits-as-parameters
```

But code like this works:
```rust
#[derive(juniper::GraphQLObject)]
pub struct Data {
   pub value: String,
}

#[derive(juniper::GraphQLObject)]
pub struct User {
   pub username: String,
   pub values: Vec<Data>,
}
```

LegNeato:
You need to add "value" as an async function.

This will be fixed when I rip out the sync resolvers or make the derive use sync rather than async resolvers. Happy to accept a patch for the latter.

LegNeato:
We are not going to be removing the sync code.

loafofpiecrust:
I'm having the same problem where I want to nest a type defined with `#[juniper::graphql_object]` with only `async` functions inside one defined with `
#[derive(juniper::GraphQLObject)]` and I get these same kinds of errors.
Is this something we want to fix or not?

LegNeato:
If you add `noasync` to the first case (a la https://github.com/graphql-rust/juniper/blob/4d77a1a9b9b0e60cbeb200527289bd45afec4141/juniper/src/schema/schema.rs#L134) then does it work?

tyranron:
[...]

PR #652, opened 2020-05-05
Make graphql_object macro supports deriving object field resolvers

zhenwenc:
This PR enhance the `graphql_object` macro to optionally derive field resolvers for fields defined in the object type struct.

```rust
#[derive(GraphQLObjectInfo)]
#[graphql(scalar = DefaultScalarValue)]
struct Obj {
    regular_field: bool,
}

#[juniper::graphql_object(derive_fields)]
impl Obj {
    fn custom_field_resolve() -> bool {
        true
    }
}
```

---

Related: #553

### DIFF
--- a/integration_tests/juniper_tests/src/codegen/impl_object_with_derive_fields.rs
+++ b/integration_tests/juniper_tests/src/codegen/impl_object_with_derive_fields.rs
@@ -1,0 +1,418 @@
+#[cfg(test)]
+use fnv::FnvHashMap;
+
+use juniper::{DefaultScalarValue, GraphQLObjectInfo};
+
+#[cfg(test)]
+use juniper::{
+    self, execute, EmptyMutation, EmptySubscription, GraphQLType, Registry, RootNode, Value,
+    Variables,
+};
+
+#[derive(Default)]
+struct Context {
+    value: String,
+}
+
+impl juniper::Context for Context {}
+
+#[derive(GraphQLObjectInfo)]
+#[graphql(scalar = DefaultScalarValue)]
+struct Obj {
+    regular_field: bool,
+
+    #[graphql(name = "renamedField")]
+    renamed_field_orig: i32,
+
+    #[graphql(skip)]
+    skipped_field: i32,
+}
+
+#[juniper::graphql_object(name = "MyObj", description = "obj descr", derive_fields)]
+impl Obj {
+    fn resolve_field() -> &str {
+        "obj::resolve_field"
+    }
+}
+
+#[derive(GraphQLObjectInfo)]
+#[graphql(scalar = DefaultScalarValue)]
+struct Nested {
+    obj: Obj,
+    nested_field: bool,
+}
+
+#[juniper::graphql_object(derive_fields)]
+impl Nested {
+    fn nested_resolve_field() -> &str {
+        "nested::resolve_field"
+    }
+}
+
+#[derive(GraphQLObjectInfo)]
+#[graphql(Context = Context, scalar = DefaultScalarValue)]
+struct WithContext {
+    value: bool,
+}
+
+#[juniper::graphql_object(Context = Context, derive_fields)]
+impl WithContext {
+    fn resolve_field(ctx: &Context) -> &str {
+        ctx.value.as_str()
+    }
+}
+
+// FIXME: Field with lifetime doesn't even work for derive GraphQLObject
+//        due to 'cannot infer an appropriate lifetime'.
+
+// #[derive(GraphQLObjectInfo)]
+// #[graphql(Context = Context, scalar = DefaultScalarValue)]
+// struct WithLifetime<'a> {
+//     value: &'a str,
+// }
+
+// #[juniper::graphql_object(Context = Context)]
+// impl<'a> WithLifetime<'a> {
+//     fn custom_field() -> bool {
+//         true
+//     }
+// }
+
+struct Query;
+
+#[juniper::graphql_object(Context = Context, scalar = DefaultScalarValue)]
+impl Query {
+    fn obj() -> Obj {
+        Obj {
+            regular_field: true,
+            renamed_field_orig: 22,
+            skipped_field: 33,
+        }
+    }
+
+    fn nested(&self) -> Nested {
+        Nested {
+            obj: Obj {
+                regular_field: false,
+                renamed_field_orig: 222,
+                skipped_field: 333,
+            },
+            nested_field: true,
+        }
+    }
+
+    fn with_context(&self) -> WithContext {
+        WithContext { value: true }
+    }
+
+    // fn with_lifetime(&self) -> WithLifetime<'a> {
+    //     WithLifetime { value: "blub" }
+    // }
+}
+
+#[tokio::test]
+async fn test_derived_object_fields() {
+    assert_eq!(
+        <Obj as GraphQLType<DefaultScalarValue>>::name(&()),
+        Some("MyObj")
+    );
+
+    // Verify meta info.
+    let mut registry = Registry::new(FnvHashMap::default());
+    let meta = Obj::meta(&(), &mut registry);
+
+    assert_eq!(meta.name(), Some("MyObj"));
+    assert_eq!(meta.description(), Some(&"obj descr".to_string()));
+
+    let doc = r#"
+        {
+            obj {
+                regularField
+                renamedField
+                resolveField
+            }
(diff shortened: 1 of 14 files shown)


zhenwenc:
Not sure how to add reviewers, would be great if anyone can help. Thanks! 

^^ @LegNeato @jmpunkt @Victor-Savu

---

While I was working on the PR, I also noticed a few potential issues. But I am not sure who to talk to or maybe they are actually intended or work-in-progress features:

1.  Looks like its a typo?
https://github.com/graphql-rust/juniper/blob/d13305f2029c12b11844b2fdf941f081e835a366/juniper_codegen/src/util/mod.rs#L649

2.  Derive `GraphQLObject` macro for struct with lifetime doesn't work. The following struct found in a test suite, which can reproduce the issue, but seems its not been used in any test case.
https://github.com/graphql-rust/juniper/blob/d13305f2029c12b11844b2fdf941f081e835a366/integration_tests/juniper_tests/src/codegen/derive_object.rs#L63-L66

3.  Derive `GraphQLObject` macro requires `scalar` attribute, otherwise generic `__S` will be used instead of the default `DefaultScalarValue`. For instance

```rust
#[derive(GraphQLObject)]
struct Object {
    value: i32,
}
```

should be the same as

```rust
#[derive(GraphQLObject)]
#[graphql(scalar = juniper::DefaultScalarValue)] // <---- here
struct Object {
    value: i32,
}
```

jmpunkt:
1. You can fix the typo.
2. Do you know why this does not work? Is the lifetime parameter missing at the generics definition?
3. There is an issue related to this (https://github.com/graphql-rust/juniper/issues/580). Not sure how to deal with this, I would ignore it for now. A workaround for this is to specify `#[graphql(Scalar = juniper::DefaultScalarValue)]`

zhenwenc:
Thanks @jmpunkt !

Re 2:

Given the following code:
```rust
mod test {
    #[graphql(scalar = juniper::DefaultScalarValue)]
    struct WithLifetime<'a> {
        value: &'a str,
    }

    struct Query;

    #[juniper::graphql_object(scalar = juniper::DefaultScalarValue)]
    impl<'a> Query {
        fn with_lifetime(&self) -> WithLifetime<'a> {
            WithLifetime { value: "blub" }
        }
    }
}
```

Here is the compile error, which I think it is related to the async resolver (but can be wrong):

```
error[E0495]: cannot infer an appropriate lifetime due to conflicting requirements
  --> integration_tests/juniper_tests/src/lib.rs:20:5
   |
20 |     #[juniper::graphql_object(scalar = juniper::DefaultScalarValue)]
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
note: first, the lifetime cannot outlive the lifetime `'a` as defined on the impl at 21:10...
  --> integration_tests/juniper_tests/src/lib.rs:21:10
   |
21 |     impl<'a> Query {
   |          ^^
note: ...so that the expression is assignable
  --> integration_tests/juniper_tests/src/lib.rs:20:5
   |
20 |     #[juniper::graphql_object(scalar = juniper::DefaultScalarValue)]
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   = note: expected  `foo::WithLifetime<'_>`
              found  `foo::WithLifetime<'a>`
note: but, the lifetime must be valid for the lifetime `'b` as defined on the method body at 20:5...
  --> integration_tests/juniper_tests/src/lib.rs:20:5
   |
20 |     #[juniper::graphql_object(scalar = juniper::DefaultScalarValue)]
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
note: ...so that the expression is assignable
  --> integration_tests/juniper_tests/src/lib.rs:20:5
   |
20 |     #[juniper::graphql_object(scalar = juniper::DefaultScalarValue)]
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   = note: expected  `std::pin::Pin<std::boxed::Box<(dyn core::future::future::Future<Output = std::result::Result<juniper::value::Value, juniper::executor::FieldError>> + std::marker::Send + 'b)>>`
              found  `std::pin::Pin<std::boxed::Box<dyn core::future::future::Future<Output = std::result::Result<juniper::value::Value, juniper::executor::FieldError>> + std::marker::Send>>`
```

Here is the generated `GraphQLTypeAsync` for `Query`:
```rust
    impl<'a> juniper::GraphQLTypeAsync<juniper::DefaultScalarValue> for Query
    where
        juniper::DefaultScalarValue: Send + Sync,
        Self: Send + Sync,
    {
        fn resolve_field_async<'b>(
            &'b self,
            info: &'b Self::TypeInfo,
            field: &'b str,
            args: &'b juniper::Arguments<juniper::DefaultScalarValue>,
            executor: &'b juniper::Executor<Self::Context, juniper::DefaultScalarValue>,
        ) -> juniper::BoxFuture<'b, juniper::ExecutionResult<juniper::DefaultScalarValue>>
        where
            juniper::DefaultScalarValue: Send + Sync,
        {
            use futures::future;
            use juniper::GraphQLType;
            match field {
                "withLifetime" => {
                    let res: WithLifetime<'a> = (|| WithLifetime { value: "blub" })();
                    let res2 = juniper::IntoResolvable::into(res, executor.context());
                    let f = async move {
                        match res2 {
                            Ok(Some((ctx, r))) => {
                                let sub = executor.replaced_context(ctx);
                                sub.resolve_with_ctx_async(&(), &r).await
                            }
                            Ok(None) => Ok(juniper::Value::null()),
                            Err(e) => Err(e),
                        }
                    };
                    use futures::future;
                    future::FutureExt::boxed(f)
                }
                _ => {
                    {
                        ::std::rt::begin_panic_fmt(&::core::fmt::Arguments::new_v1(
                            &["Field ", " not found on type "],
                            &match (
                                &field,
                                &<Self as juniper::GraphQLType<juniper::DefaultScalarValue>>::name(
                                    info,
                                ),
                            ) {
                                (arg0, arg1) => [
                                    ::core::fmt::ArgumentV1::new(arg0, ::core::fmt::Display::fmt),
                                    ::core::fmt::ArgumentV1::new(arg1, ::core::fmt::Debug::fmt),
                                ],
                            },
                        ))
                    };
                }
            }
        }
    }
```

LegNeato:
While I like the idea of this in general, I think it massively complicates the UX for folks to save on some typing. This reminds me a lot of Rust accessors and transparent newtypes. Rust itself requires boilerplate (`self.0`) but people who mind the repetition build a layer on top with proc macros and there are multiple libraries that do it based on various tradeoffs. I always assumed people would just use the proc macro with a bit of boilerplate and use `juniper_codegen` to make their own macros with their preferred tradeoffs if this was an issue.

Personally, I like how simple and explicit the UX is right now. Either it graphql projection is trivial and the derive suffices or it is complex and everything is right there in the `impl`.

I'm happy to entertain a patch if we can keep this simple and explicit. I'm not sure we can get there with rust proc macros as-is though 🤷

zhenwenc:
@LegNeato Thanks for your reply, I understand your concern and agree that Juniper should not be opinionated. 

However, since `juniper_codegen` already achieved a great effort on collecting necessary information to build GraphQL schema (such as fields for object / input types, field resolvers, etc), is it possible to expose these collected information to the user instead of generating the `GraphQLType` directly, so that user can use these information in their own macro to generate `GraphQLType`.

For example juniper can generate `GraphQLTypeInfo` that holds such information, similar to what this PR does. Then I can build my own `graphql_object` macro with the functionalities I need very easily.

Let me know what you think. Thanks!

smmoosavi:
can we have multiple `derive_fields`? it would be awesome if we can.

file1:
```rust
#[derive(GraphQLObjectInfo)]
#[graphql(scalar = DefaultScalarValue)]
struct Obj {
    regular_field: bool,
}

#[juniper::graphql_object(derive_fields)]
impl Obj {
    fn custom_field_resolve() -> bool {
        true
    }
}
```
file2:
```rust
#[juniper::graphql_object(derive_fields)]
impl Obj {
    fn other_field_resolve() -> bool {
        true
    }
}
```

PR #657, opened 2020-05-11
impl GraphQLScalar for NaiveTime

c410-f3r:


LegNeato:
Thanks for the PR! I'm actually not sure we want to land this. One can imagine different users needing different time representations (e.g. 1:00pm vs 13:00:00, 09:15:00 vs 09:15, etc). Need to think about it some more.

tyranron:
@LegNeato despite there is [nothing in spec about it yet](https://github.com/graphql/graphql-spec/issues/579), I suppose it's naturally to follow [RFC 3339](https://www.ietf.org/rfc/rfc3339.html#section-4) time representation for a default library scalar.

If somebody needs some custom time formatting, he always can go with a custom scalar.

LegNeato:
Yeah, we do that for other date/time stuff. I guess it is not clear what context `NaiveDateTime` is used in so it is less clear there is a default way to represent or even what the data `NaiveDateTime` represents in most contexts.

I'm inclined to merge this but I feel it moves Juniper to be more opinionated and less universal. Though I guess an argument can be made with `uuid` and dashes vs no dashes. But in that case at least the underlying data domain is the same.

c410-f3r:
If following an hypothetical or concrete standard is a hard constraint, then it is possible to write these "non-standard" implementations behind a feature flag, say "extra", "non-standard-impls" or "shinigamis". By doing so, you guys will take off the burden of someone that must create a wrapper and implement all by himself due to coherence rules, plus avoiding the spread of duplicated code across different projects.

LegNeato:
Let's put it behind a flag, off by default. Let's  use the prefix `scalar-` so future ones can follow the same format and people can turn on whatever combo they want.

c410-f3r:
@LegNeato Thank you

LegNeato:
Thanks for the PR and adding the flag and tests!

PR #662, opened 2020-05-14
Improve lookahead visibility for aliased fields

bumblepie:
Add a method to access the children of look ahead structs
Make the behaviour around accessing aliased lookahead fields more consistent


bumblepie:
Fixes issue #662

LegNeato:
Thank you! I wonder if smallvec would be better. Let's deprecate the other functions.

bumblepie:
Deprecation done, I couldn't see smallvec used already in Juniper, so I wasn't confident to introduce it as a dependency just for this. It might be a good optimisation in the future though.

LegNeato:
Thank you so much for the PR and including tests! 🍻

Issue #295, opened 2018-12-18
Replace macros with proc_macros


theduke:
Replace all macros used for defining GraphQL types with proc macro implementations.

* [x] graphql_interface  (#682, #1009)
* [x] graphql_object  (#181, #333)
* [x] graphql_scalar  (#1014, #1017, #1025)
* [x] graphql_union  (https://github.com/graphql-rust/juniper/commit/ee9a82a8177b50ccf59acaad1e9cce7a73ea9a85, #666)
* [x] ~~graphql_value~~ (not necessary,  #996)

This should probably be a  two-step process: first implement proc-macro alternatives for everything and have tests for functional equivalence.
Then remove the old macros with a new breaking version.

samuela:
Forgive me if this is a noob question, but what's the difference between macros and proc macros?

tyranron:
@samuela 

> Forgive me if this is a noob question, but what's the difference between macros and proc macros?

If it's still relevant, see the official reference:
- [Macros By Example](https://doc.rust-lang.org/reference/macros-by-example.html)
- [Procedural Macros](https://doc.rust-lang.org/reference/procedural-macros.html)

#### TL;DR

Declarative macros:
- declared via pattern-match syntax
- guaranteed to be hygienic
- restricted in expresiveness to certain patterns

Procedural macros:
- declared as a separate crate
- hygiene is up to an implementor
- accept and produce raw tokens, so can generate arbitrary kind of stuff

Issue #467, opened 2019-11-24
GraphQLType is not satisfied with juniper::object


jasonlor:
Currently using 0.14.1. I'm attempting to implement fields with the juniper::object macro.

```
#[juniper::object(
    Context = Context,
    Scalar = juniper::DefaultScalarValue,
)]
impl Event {

```
On compilation, i see the following error:
error[E0277]: the trait bound `models::Event: juniper::GraphQLType<__S>` is not satisfied

On this struct, I've implemented another trait that does not have the juniper::object macro.
impl TraitName for Event

I've looked through the docs and issues and have been unable to find a solution.

Thanks!

davidpdrsn:
Are you able to share a minimal reproduction script? It’s a bit hard to say what’s wrong from what you posted.

andy128k:
I have found similar issue. [Here](https://github.com/andy128k/juniper-issue-467) is minimal example which fails to build.

jasonlor:
Bump, any ideas?

RyuuGan:
Hello,

Any updates on this ?

```
#[derive(GraphQLObject)]
   |          ^^^^^^^^^^^^^ the trait `juniper::types::base::GraphQLType<__S>` is not implemented for `api::graphql::projects::GQLProject`
```

I also trying to use it as in @andy128k example

RyuuGan:
Hello,

I found a workaround, so it will compile and work as expected. The idea is that you don't use `#[juniper::object]` inside the `#[derive(GraphQLObject)]`, but what you can do is define all objects as `#[juniper::object]`. Yes it is a little bit ugly to define all the methods, but at least it works.

So instead of:

```rust
pub struct Article {
    pub title: String,
}

#[juniper::object(name = "Article")]
impl Article {
    fn title(&self) -> &str { &self.title }
}

#[derive(juniper::GraphQLObject)]
pub struct Blog {
    pub url: String,
    pub description: Vec<String>,
    pub articles: Vec<Article>,
}
```

You define: 

```rust
pub struct Article {
    pub title: String,
}

#[juniper::object(name = "Article")]
impl Article {
    fn title(&self) -> &str { &self.title }
}

pub struct Blog {
    pub url: String,
    pub description: Vec<String>,
    pub articles: Vec<Article>,
}

#[juniper::object(name = "Blog")]
impl Blog {
  // ...
}
```

zhenwenc:
[...]

PR #666, opened 2020-05-22
Codegen reimplementation for GraphQL unions

### LINKED ISSUE #549: Macro `#[juniper::graphql_union]` does not impl `GraphQLTypeAsync`

### LINKED ISSUE #663: Need to manually depend on `futures`
**Describe the bug**
I was starting to play with subscriptions, for which I needed to depend on the master branch

> Incidentally, a new release which includes subscriptions would be swell :)

My `Cargo.toml` therefore looks something like this:

```
[dependencies]
rocket = "0.4.4"

[dependencies.juniper]
# 2020-05-15 newer than 0.14.2
git = "https://github.com/graphql-rust/juniper.git"
rev = "91a335398313356fcd6627c00f4b9fc504714f48"

[dependencies.juniper_rocket]
# 2020-05-15 newer than 0.5.2
git = "https://github.com/graphql-rust/juniper.git"
rev = "91a335398313356fcd6627c00f4b9fc504714f48"
```

Suppose I then try to compile the following `main.rs`

```
#![feature(decl_macro, proc_macro_hygiene)]

#[derive(juniper::GraphQLEnum, Clone, Copy)]
enum Episode {
  NewHope,
  Empire,
  Jedi,
}

fn main() {}
```

I get

```
error[E0432]: unresolved import `futures`
 --> src/main.rs:3:10
  |
3 | #[derive(juniper::GraphQLEnum, Clone, Copy)]
  |          ^^^^^^^^^^^^^^^^^^^^ use of undeclared type or module `futures`
  |
  = note: this error originates in a derive macro (in Nightly builds, run with -Z macro-backtrace for more info)
```

I can work around this by adding `futures = "0.3.5"` to my `Cargo.toml` and everything compiles fine. Clearly somewhere a dependency on `futures` isn't being properly expressed.

c.f. also #534 

**Additional context**

Rust toolchain `nightly-2020-05-07`

tyranron:
Fixes #549
As side effect, also fixes #663
Also, related to #295 

## Background and motivation

__TL;DR__: current codegen for GraphQL unions is not much flexible and is too magic.

- `#[derive(GraphQLUnion)]` macro:
    - Applicable only to enums.
    - Doesn't support generics on type, even if they're not required for type resolution.
    - Cannot work with multiple `#[graphql]` attributes, just pickups first and silently ignores others.
    - Not cleverly enough generates `From` implementations, which leads to painful conflicts.
- `#[graphql_union]` macro:
    - The code being fed into macro [looks like a valid Rust code](https://github.com/graphql-rust/juniper/blob/5f777e8a6dc1c5b947b5fae578e9307ab2e24558/integration_tests/juniper_tests/src/codegen/impl_union.rs#L38-L43), but actually is not.
    - Operates with types/traits which don't really exist (`GraphQLUnion`), and declares `impl` blocks which don't exist after macro expansion too.

We definitelly want less magic, applying custom behavior in an ergonomic and convenient way and operate with valid Rust code for better IDE experience.

## New design

In a nutshell, it provides a single `#[derive(GraphQLUnion)]` macro which allows to modify resolution logic with custom functions. This is the same what `serde` does with `#[derive(Serialize)]` macro and `#[serde(with = "my_func")]` attribute.
```rust
#[derive(GraphQLUnion, derive_more::From)]
#[graphql(name = "TheCharacter", desc = "Superb character")]
#[graphql(context = MyContext)]  // <-- if not specified, then Context is `()`
#[graphql(scalar = juniper::DefaultScajarValue)] // <-- if not specified, then generated code is parametrized by `__S: ScalarValue`
#[graphql(on Jedi = Character::as_jedi)]
#[graphql(on Jedi = Character::as_better_jedi)]  // <-- emits compile-time duplication error
#[graphql(on Sith = Character::as_sith)]  // <-- resolution logic doesn't require the type to be a part of enum
pub enum Character<T> {
    One(Human),  // <-- will be trivially resolved as expected

    #[graphql(with = Character::as_droid)]
    Two(Droid),  // <-- will be resolved with `Character::as_droid` function

    #[graphql(ignore)]
    #[from(ignore)]
    Three<T>,   // <-- will be skipped

    #[graphql(with = Character::as_jedi)]  // <-- emits compile-time duplication error
    #[from(ignore)]
    Jedi,  // <-- if custom resolver specified, the variant doesn't have to containt the single type value
}

impl<T> Character<T> {
    // We can use in fuctions as context any type which implements `FromContext<MyContext>`.
    fn as_droid(&self, _: &()) -> Option<&Droid> {
        if let Self::Two(droid) = self {
            Some(droid)
        } else {
            None
        }
    }

    fn as_jedi(&self, ctx: &MyContext) -> Option<&Droid> {
       ctx.can_i_haz_the_jedi_please()
    }

    fn as_sith(&self, ctx: &SithContext) -> Option<&Sith> {
       ctx.deploy_new_sith_here()
    }
}
```

All the described above is applicable in the same manner to struct types too.
```rust
#[derive(GraphQLUnion]
#[graphql(name = "TheCharacter", desc = "Superb character")]
#[graphql(context = MyContext)]  // <-- if not specified, then Context is `()`
#[graphql(scalar = juniper::DefaultScajarValue)] // <-- if not specified, then generated code is parametrized by `__S: ScalarValue`
#[graphql(on Jedi = Character::resolve_jedi, on Sith = Character::deploy_sith)]`
struct TheCharacter<T>(PhantomData<T>);

impl<T> Character<T> {
    fn resolve_jedi(&self, ctx: &Context) -> Option<&Jedi> {
        ctx.can_i_haz_the_jedi_please()
    }

    fn deploy_sith(&self, ctx: &SithContext) -> Option<&Sith> {
        ctx.deploy_new_sith_here()
    }
}
```

Also, for enums, the innerhood `From` implementations are removed. It's better to used `derive_more::From` for that purpose, which does stuff in much more clever way, and will preserve a bit more of SRP for our macro.

However, to support traits, we still need to provide `#[graphql_union]` macro, as `proc_macro_derive` cannot be placed on trait declarations. And to avoid collisions with `#[graphql]` helper attribute (which will result in compilation error), we shoud use `#[graphql_union]` even for methods.
```rust
#[graphq_union(name = "Character", desc = "Traited character")]
#[graphq_union(context = MyContext)]  // <-- if not specified, then auto-inferred from methods, if possible, otherwise is `()`
#[graphq_union(scalar = juniper::DefaultScajarValue)] // <-- if not specified, then generated code is parametrized by `__S: ScalarValue`
#[graphq_union(on Jedi = custom_resolve_as_jedi)]  // <-- it's possible even for traits!
trait Character {  // <-- trait, however, needs to be object-safe as dynamic dispatch is involved
    fn as_human(&self) -> Option<&Human> { None } // <-- we can omit context here if we want
    fn as_droid(&self, _: &Context) -> Option<&Droid> { None }
    #[graphq_union(ignore)]
    fn unrelated_trait_method(&self) { }
}

impl Character for Human {
    fn as_human(&self) -> Option<&Human> {
        Some(self)
    }
}

impl Character for Droid {
    fn as_droid(&self, _: &Context) -> Option<&Droid> {
        Some(self)
    }
}

fn custom_resolve_as_jedi(ch: &dyn Character, ctx: &MyContext) -> Option<&Jedi> {
     ctx.can_i_haz_the_jedi_please(ch)
}
```

Also, this PR:
- Adds `juniper::marker::GraphQLUnion` trait, to not reffer in the code to unexistent trait/types.
- Relaxes `Sized` requirement on `GraphQLType` trait to allow implement it for bare trait objects.

## Features

- Fully valid Rust code. No magic at all. All macro are additive and only add new code without transforming/removing existing.
- Idiomatic and familiar way for handling custom logic in proc macros: `#[attr(with = func_name)]`.
- No need to choose between macros. There is a single macro for each situation, which still allows tweaking with custom behavior/logic.
- Multiple `#[graphql]` attributes are supported.
- Generics are supported.

## Checklist
- [x] Support parsing multiple `#[graphql]` attributes.
    - [x] Duplication check inside single attribute.
    - [x] Duplication check across multiple attributes.
- [x] Re-export `futures` crate inside `juniper` with `#[doc(hidden)]` for internal use in macros.
- [x] Make `GraphQLUnion` a real marker trait.
- [x] Relax `Sized` requirement on `GraphQLType` trait.
- [x] Enums (`#[derive(GraphQLUnion)]`):
    - [x] infer trivial implementation for enum variants;
    - [x] support ignoring variant with `#[graphq(ignore)]` attribute;
    - [x] support generics in type declaration;
    - [x] support custom context type with `#[graphql(context = Type)]` attribute;
    - [x] support custom `ScalarValue` type `#[graphql(scalar = Type)]` attribute;
    - [x] support custom GraphQL name and description with `#[graphql(name = "Type")]` and `#[graphql(description = "A Type desc.")]` attributes;
    - [x] support custom resolvers on type itself with `#[grapqh(on Type = resolver_fn)]` attribute;
    - [x] support custom resolver on a type variant with `#[grapqh(with = resolver_fn)]` attribute;
        - [x] duplication check between `#[grapqh(on Type = resolver_fn)]` and `#[grapqh(with = resolver_fn)]` attributes for the same type;
    - [x] remove auto-generated `From` implementations for enum variants.
- [x] Structs (`#[derive(GraphQLUnion)]`):
    - [x] support custom resolvers on type itself with `#[grapqh(on Type = resolver_fn)]` attribute;
    - [x] support generics in type declaration;
    - [x] support custom context type with `#[graphql(context = Type)]` attribute;
    - [x] support custom `ScalarValue` type `#[graphql(scalar = Type)]` attribute;
    - [x] support custom GraphQL name and description with `#[graphql(name = "Type")]` and `#[graphql(description = "A Type desc.")]` attributes.
- [x] Traits (`#[graphql_union]`):
    - [x] use trait methods with return `Option<&Type>` as union variant resolvers;
        - [x] allow omitting `Context` in methods;
        - [x] work correctly with parenthesized types;
        - [x] allow full import paths as types;
        - [x] validate signature for good error messages;
    - [x] support multiple attributes
        - [x] remove duplicated attributes on first expansion
    - [x] support custom resolvers on type itself with `#[graphql_union(on Type = resolver_fn)]` attribute;
    - [x] support custom context type with `#[graphql_union(context = Type)]` attribute;
    - [x] support custom `ScalarValue` type `#[graphql_union(scalar = Type)]` attribute;
    - [x] support custom GraphQL name and description with `#[graphql_union(name = "Type")]` and `#[graphql_union(description = "A Type desc.")]` attributes;
    - [x] ignore trait methods with `#[graphql_union(ignore)]` attribute.
- [x] Review tests and provide rich coverage of all situations
- [x] Documentation and Book are updated
- [x] Changelog is updated


tyranron:
And no... PR number doesn't have any subtle meanings 🙃

LegNeato:
I'll take a closer look but this looks awesome!

tyranron:
@LegNeato 

> Why is this `on` needed when below there is a `with`?

The `on` is chosen to repeat the [GraphQL syntax for dispatching union variants](https://spec.graphql.org/June2018/#example-f8163) in queries.

How do these two attributes differ:
- `#[graphql(on UnionVariantType = resolve_fn)]` is _always placed on top_ of the whole type declaration and _always specifies the `UnionVariantType`_ that is intented to be resolved with the custom `resolve_fn`. This allows to specify custom resolvers for types, that don't exist in enum type declaration at all. That's why this attribute is applicable to structs and traits too:
    ```rust
    #[derive(GraphQLUnion]
    #[graphql(on Jedi = Character::resolve_jedi, on Sith = Character::deploy_sith)]`
    struct Character; // <-- no fields at all here
    
    impl Character {
        fn resolve_jedi(&self, ctx: &Context) -> Option<&Droid> {
            ctx.can_i_haz_the_jedi_please()
        }

        fn deploy_sith(&self, ctx: &SithContext) -> Option<&Sith> {
            ctx.deploy_new_sith_here()
        }
    }
    ```

- `#[graphql(with = resolve_fn)]` is _applicable only to enum variants_, where the resolving `UnionVariantType` is _part of enum type declaration_. It's just a short, but ergonomic form, applicable to enum variants only.
    ```rust
    #[derive(GraphQLUnion)]
    enum Character {
        First(Human),
        #[graphql(with = custom_droid_resolver)]
        Second(Droid),
    }
    ```

tyranron:
@LegNeato ok, I've stuck a bit with `proc_macro_attribute` and can't decide what is better. I need an advise.

#### Synopsis

It turns out that the following is totally OK when using `proc_macro_derive`, as it allows to register helper `graphql` attribute, but is impossible when using `proc_macro_attribute`:
```rust
#[graphq_union]
#[graphql(name = "MyCharacter", desc = "Traited character")]
trait Character {
    fn as_human(&self) -> Option<&Human>;
}
```
The problem is that `proc_macro_attribute` has no ability to register helper attributes, so `#[graphql(...)]` is unknown and compiler complains:
```
error: cannot find attribute `graphql` in this scope
  --> integration_tests/juniper_tests/src/codegen/union_attr.rs:19:3
   |
19 | #[graphql(description = "A Collection of things")]
   |   ^^^^^^^
```

We could have been define such attribute separately, but it raises 2 other major downsides:
- library user should import `graphql` attribute (via `use juniper::graphql`), which is not obvious and ergonomic;
- when `proc_macro_derive` and `proc_macro_attribute` are used in the same module, the `graphql` attributes will conflict.
    ```
    error[E0659]: `graphql` is ambiguous (derive helper attribute vs any other name)
       --> integration_tests/juniper_tests/src/codegen/union_derive.rs:169:3
        |
    169 | #[graphql(Context = CustomContext)]
        |   ^^^^^^^ ambiguous name
        |
    ```

I see the following possible solutions...

#### 1. Use only `graphql_union` attribute
```rust
use juniper::graphql_union;

#[graphql_union(name = "MyCharacter", desc = "Traited character")]
#[graphql_union(scalar = juniper::DefaultScajarValue)]
#[graphql_union(on Jedi = custom_resolve_as_jedi)]
trait Character {
    fn as_human(&self) -> Option<&Human>;
    #[graphql_union(ignore)]
    fn some();
}
```
- 👍 No ambiguity on attribute names.
- 👍 No additional imports required.
- 👎 Differs too much with `#[derive(GraphQLUnion)]`.
- 👎 `graphql_union` is quite long name, which makes definition noisy.

#### 2. Provide another helper attribute
```rust
use juniper::{graphql_union, gql};

#[graphql_union]
#[gql(name = "MyCharacter", desc = "Traited character")]
#[gql(scalar = juniper::DefaultScajarValue)]
#[gql(on Jedi = custom_resolve_as_jedi)]
trait Character {
    fn as_human(&self) -> Option<&Human>;
    #[gql(ignore)]
    fn some();
}
```
- 👍 More laconic naming, so cleaner definition, similar to the one we have with `proc_macro_derive`.
- 👎 Necessity to additionally import `gql` macro.
- 👎 Too much different namings provided by `juniper` crate.

-----------------------------

Writing it up, I can conclude that the option 1 is the most sane one (despite it feels not perfect one).

But comments, considerations and ideas are welcome!

tyranron:
I've bootstrapped traits support basing on option 1.

As an another side effect:  
Previous `graphql_union!` implementation was generating code for `&'a dynTrait` trait object. I've relaxed `Sized` requirement on `GraphQLType` and now it generates implementation directly for `dyn Trait + 'a`.

The description is updated properly.

LegNeato:
Ugh, that's a bummer. Neither are ideal. Let me play around and see if I can come up with something.

LegNeato:
https://github.com/rust-lang/rust/issues/65823

LegNeato:
That issue seems to think they should be supported by the attribute proc macro and not throwing an error?

tyranron:
@LegNeato 

> That issue seems to think they should be supported by the attribute proc macro and not throwing an error?

Yup. If we could have something like proposed in rust-lang/rust#65823, the described above would be non-issue, and the initial design would be possible.

At the moment I think the option 1 is quite enough. It's not ideal, but does the job relatively obviously without subtleties. The most regular form (90% cases) in the code will be something like that:
```bash
#[graphq_union(
    name = "Character", 
    description = "Traited character",
    context = MyContext,
)]
trait Character {
    fn as_human(&self) -> Option<&Human> { None }
    fn as_droid(&self, _: &Context) -> Option<&Droid> { None }
}
```
I think users will hit the necessity to use multiple attributes for traits quite rare, in practice.

And, also, once Rust will land something like rust-lang/rust#65823 (or allow `proc_macro_derive` on traits), it will be quite easy to refactor.

jmpunkt:
For me it is not clear what happens in the following case.

```rust
#[derive(GraphQLUnion)]
pub enum Character {
    One(Human), 
    Two(Droid), 
    #[graphql(ignore)]
    Three(Three),
}

#[graphql_object]
impl Test {
     fn test(&self) -> Character {
         Character::Three(todo!()) // ignore the todo!()
     }
}
```
The resolver would fail and return an error? If `Three` should not be part of the scheme, then why is it inside of the enum. Personally Im not a huge fan of skipping variants within enums. Enums should be only contain variants which are relevant. If the user accidental did something wrong, then this error will only be visible if the it is queried which seems odd when using enums especially in the context of Rust. Or is there something Im missing?

Im wondering if it is possible to have a custom resolver with the signature `fn (&self, _: &Ctx) -> Result<A, B>`?

tyranron:
@jmpunkt 

> The resolver would fail and return an error? 

At the moment it [will panic](https://github.com/graphql-rust/juniper/blob/4a235383724946c9287e40551632c2ef4769c898/juniper_codegen/src/graphql_union/mod.rs#L466-L470). This is inherited from the [existing implmentation](https://github.com/graphql-rust/juniper/blob/558eae91df20c07566a4bd43acf1248a5a75a100/juniper_codegen/src/util/mod.rs#L1532), and, I'm afraid, is not possible to change without changing the the design of `GraphQLType` itself.

> If Three should not be part of the scheme, then why is it inside of the enum. Personally Im not a huge fan of skipping variants within enums. Enums should be only contain variants which are relevant.

It shouldn't. If you don't want to make it a part of enum, then you're not pushed to.

The reason of having `#[graphql(ignore)]` attribute is not to advertise bad practices, but to give the library user more codegen abilities and the flexibility in the cases where he may have no other options.

I can quite easily imagine the situation, when library user should deal with some enum or trait, where not all methods should be exposed in schema, but they need to be there to express some additional invariants or something. Making the user in such situation to roll out manual `GraphQLType` implementation with all the nuances will be a huge ergonomic cost. Instead, we give him the flexibility to exclude undesired things and have all the shiny generated code with  an effort of one additional helper attribute only.

> If the user accidental did something wrong, then this error will only be visible if the it is queried which seems odd when using enums especially in the context of Rust.

The point is that this is "opt-in" action, not "opt-out". So to do this accidentially, the library user should explicitly put the `#[graphql(ignore)]` attribute. To make sure that "he knows what he's doing" I think we can describe this issue in docs clearly with a huge warning.

> Im wondering if it is possible to have a custom resolver with the signature fn (&self, _: &Ctx) -> Result<A, B>?

At the moment, no. But, in theory, we could have been support `fn (&self, _: &Ctx) -> Result<Option<&A>, B>` signature. `Option<&UnionVariantType>` is mandatory due to current `GraphQLType` and resolving design, which uses that `Option` to understand whether variant should be resolved or not.

In this PR I'm trying to re-introduce capabilities of existing code generation in a better usage form. To extend those capabilities, I think we should go with a separate PR, as there are a lot questions to bikeshed and discuss.

LegNeato:
Awesome work, I'll take a look in-depth today.

I too am a bit suspect about `ignore` and would prefer to leave it out unless we have a concrete usecase. For example, with `skip` right away we knew folks would likely have a `pasword_hash` in a user struct they would not want to expose. I'd also prefer to rename `ignore` to `skip` to stay consistent with derives and serde if we keep the functionality.

tyranron:
@LegNeato both `skip` and `ignore` are supported, they are [synonyms](https://github.com/graphql-rust/juniper/blob/4feb1415bf2f94b784ae78a06dfdb6755f147aed/juniper_codegen/src/graphql_union/mod.rs#L232). I've provided support for both, because `ignore` is more widely used in Rust ecosystem (`derive_more`, for example), and `skip` is used by `juniper` for a long time.

As for other macros, I'm planning to refactor/rework them too in near future. So, at the end, with a new release they will be fully consistent, I swear 😉

> I too am a bit suspect about `ignore` and would prefer to leave it out unless we have a concrete usecase.

Actually, I've provided that, because there is a use case for ignoring in our closed codebase. This usecase involves binding some type parameters, which are not really used, but required for comfortable type-level "magic".  
Something like that:
```rust
enum UserCreationResult<T> {
    Ok(User),
    Err(UserCreationError),
    #[graphql(ignore)]
    _State(PhatomData<T>),
}
```
where `T` is used for expressing some type state patterns in resolvers code.

LegNeato:
Cool, makes sense. As long as we have a real usecase.

tyranron:
Now, the codegen implementation is complete and polished.

It's only the matter of adding more tests and updating user docs.

tyranron:
The [initial Book docs for unions](https://github.com/graphql-rust/juniper/blob/23015f61b06a3b1ab2303f000a2b15934ce57535/docs/book/content/types/unions.md) have landed.

@LegNeato @jmpunkt it would be nice to have some feedback for users docs. Is it enough or something should be described in more detailes? Did I miss something vital?

docs/book/content/types/unions.md
```diff
@@ -1,75 +1,197 @@
...
+    id: String,
+    primary_function: String,
+}
+
+#[derive(From, GraphQLUnion)]
+enum Character<S> {
+    Human(Human),
+    Droid(Droid),
+    #[from(ignore)]
+    #[graphql(ignore)]  // or `#[graphql(skip)]`, on your choice
```
LegNeato, line 74:
```suggestion
    #[graphql(ignore)]  // or `#[graphql(skip)]`, your choice
```

docs/book/content/types/unions.md
```diff
@@ -1,75 +1,197 @@
...
+    _State(PhatomData<S>),
+}
+#
+# fn main() {}
+```
+
+
+### Custom resolvers
+
+If some custom logic should be involved to resolve a [GraphQL union][1] variant, we may specify the function responsible for that.
```
LegNeato, line 84:
```suggestion
If some custom logic is needed to resolve a [GraphQL union][1] variant, you may specify a function to do so:
```

docs/book/content/types/unions.md
```diff
@@ -1,75 +1,197 @@
...
+#[derive(GraphQLUnion)]
+#[graphql(Context = CustomContext)]
+enum Character {
+    Human(Human),
+    #[graphql(with = Character::droid_from_context)]
+    Droid(Droid),
+}
+
+impl Character {
+    // NOTICE: The function signature is mandatory to accept `&self`, `&Context`
```
LegNeato, line 118:
```suggestion
    // NOTICE: The function signature must contain `&self` and `&Context` 
```

docs/book/content/types/unions.md
```diff
@@ -1,75 +1,197 @@
...
+    //         and return `Option<&VariantType>`.
+    fn droid_from_context<'c>(&self, ctx: &'c CustomContext) -> Option<&'c Droid> {
+        Some(&ctx.droid)
     }
 }
+#
+# fn main() {}
+```
+
+With a custom resolver we can even declare a new [GraphQL union][1] variant, which Rust type is absent in the initial enum definition (the attribute syntax `#[graphql(on VariantType = resolver_fn)]` follows the [GraphQL syntax for dispatching union variant](https://spec.graphql.org/June2018/#example-f8163)).
```
LegNeato, line 128:
```suggestion
With a custom resolver we can even declare a new [GraphQL union][1] variant where the Rust type is absent in the initial enum definition. The attribute syntax `#[graphql(on VariantType = resolver_fn)]` follows the [GraphQL syntax for dispatching union variants](https://spec.graphql.org/June2018/#example-f8163)).
```

docs/book/content/types/unions.md
```diff
@@ -1,75 +1,197 @@
...
 ```
 
-### Using an extra database lookup
 
-FIXME: This example does not compile at the moment
+
+
+## Structs
+
+Using Rust structs as [GraphQL unions][1] is very similar to using enums, with the nuance that specifying custom resolver is the only way to declare a [GraphQL union][1] variant.
```
LegNeato, line 181:
```suggestion
Using Rust structs as [GraphQL unions][1] is very similar to using enums, with the nuance that specifying a custom resolver is the only way to declare a [GraphQL union][1] variant.
```

docs/book/content/types/unions.md
```diff
@@ -80,49 +202,91 @@ struct Database {
...
+```
 
-impl Character for Droid {
-    fn id(&self) -> &str { self.id.as_str() }
+
+
+
+## Traits
+
+Sometimes it may seem very reasonable to use Rust trait for representing a [GraphQL union][1]. However, to do that, we should introduce a separate `#[graphql_union]` macro, because [Rust doesn't allow to use derive macros on traits](https://doc.rust-lang.org/stable/reference/procedural-macros.html#derive-macros) at the moment.
```
LegNeato, line 234:
```suggestion
To use a Rust trait definition as a [GraphQL union][1] you need to use the `#[graphql_union]` macro. [Rust doesn't allow derive macros on traits](https://doc.rust-lang.org/stable/reference/procedural-macros.html#derive-macros) so using `#[derive(GraphQLUnion)]` on traits does not work.
```

docs/book/content/types/unions.md
```diff
@@ -80,49 +202,91 @@ struct Database {
...
-impl<'a> GraphQLUnion for &'a dyn Character {
-    fn resolve(&self, context: &Database) {
-        match self {
-            Human => context.humans.get(self.id()),
-            Droid => context.droids.get(self.id()),
-        }
-    }
+#[graphql_union]
+trait Character {
+    // NOTICE: The function signature is mandatory to accept `&self`
```
LegNeato, line 256:
```suggestion
    // NOTICE: The function signature must contain `&self` 
```

docs/book/content/types/unions.md
```diff
@@ -80,49 +202,91 @@ struct Database {
...
+}
+#
 # fn main() {}
 ```
 
-## Placeholder objects
+
+### Custom context
+
+If a context is required in trait method to resolve a [GraphQL union][1] variant, we may just specify it in arguments.
```
LegNeato, line 276:
```suggestion
If a context is required in a trait method to resolve a [GraphQL union][1] variant, specify it as an argument. 
```

docs/book/content/types/unions.md
```diff
@@ -133,87 +297,138 @@ struct Database {
...
     droids: HashMap<String, Droid>,
 }
-
 impl juniper::Context for Database {}
 
-struct Character {
-    id: String,
+#[graphql_union(Context = Database)]
+trait Character {
+    // NOTICE: The function signature, however, may optionally accept `&Context`.
```
LegNeato, line 304:
```suggestion
    // NOTICE: The function signature may optionally contain `&Context`.
```

docs/book/content/types/unions.md
```diff
@@ -133,87 +297,138 @@ struct Database {
...
 
-# #[allow(dead_code)]
-enum Character {
-    Human(Human),
-    Droid(Droid),
+#[graphql_union]
+trait Character {
+    fn as_human(&self) -> Option<&Human> { None }
+    fn as_droid(&self) -> Option<&Droid> { None }
+    #[graphql_union(ignore)]  // or `#[graphql_union(skip)]`, on your choice
```
LegNeato, line 348:
```suggestion
    #[graphql_union(ignore)]  // or `#[graphql_union(skip)]`, your choice
```

docs/book/content/types/unions.md
```diff
@@ -133,87 +297,138 @@ struct Database {
...
 
-## Enums (Derive)
 
-This example is similar to `Enums (Impl)`. To successfully use the
-derive macro, ensure that each variant of the enum has a different
-type. Since each variant is different, the device macro provides
-`std::convert::Into<T>` converter for each variant.
+### Custom resolvers
+
+And, of course, similarly to enums and structs, it's not mandatory to use trait methods as [GraphQL union][1] variants resolvers, the custom functions may be specified as well.
```
LegNeato, line 368:
```suggestion
Similarly to enums and structs, it's not mandatory to use trait methods as [GraphQL union][1] variant resolvers. and instead custom functions may be specified:
```

docs/book/content/types/unions.md
```diff
@@ -133,87 +297,138 @@ struct Database {
...
+}
+impl juniper::Context for Database {}
+
+#[graphql_union(Context = Database)]
+#[graphql_union(
+    on Human = DynCharacter::get_human,
+    on Droid = get_droid,
+)]
+trait Character {
+    #[graphql_union(ignore)]  // or `#[graphql_union(skip)]`, on your choice
```
LegNeato, line 400:
```suggestion
    #[graphql_union(ignore)]  // or `#[graphql_union(skip)]`, your choice
```

integration_tests/juniper_tests/src/codegen/union_derive.rs
```diff
@@ -91,31 +164,31 @@ impl DroidCompat {
...
 
-#[derive(juniper::GraphQLUnion)]
+#[derive(GraphQLUnion)]
 #[graphql(Context = CustomContext)]
 pub enum DifferentContext {
     A(DroidContext),
     B(Droid),
 }
 
 // NOTICE: this can not compile due to generic implementation of GraphQLType<__S>
```
LegNeato, line 174:
```suggestion
// NOTICE: this does not compile due to generic implementation of GraphQLType<__S>
```

juniper/CHANGELOG.md
```diff
@@ -29,10 +29,21 @@ See [#618](https://github.com/graphql-rust/juniper/pull/618).
 
 - Derive macro `GraphQLEnum` supports custom context (see [#621](https://github.com/graphql-rust/juniper/pull/621))
 
+- Reworked `#[derive(GraphQLUnion)]` macro ([#666]):
+    - applicable to enums and structs;
+    - supports custom resolvers;
+    - supports generics;
```
LegNeato, line 35:
```suggestion
- Reworked `#[derive(GraphQLUnion)]` macro ([#666]):
    - Applicable to enums and structs.
    - Supports custom resolvers.
    - Supports generics.
```

juniper/CHANGELOG.md
```diff
@@ -29,10 +29,21 @@ See [#618](https://github.com/graphql-rust/juniper/pull/618).
...
+    - supports multiple `#[graphql]` attributes.
+- New `#[graphql_union]` macro ([#666]):
+    - applicable to traits;
+    - supports custom resolvers;
+    - supports generics;
+    - supports multiple `#[graphql_union]` attributes.
```
LegNeato, line 41:
```suggestion
- New `#[graphql_union]` macro ([#666]):
    - applicable to traits;
    - Supports custom resolvers.
    - Supports generics.
    - Supports multiple `#[graphql_union]` attributes.
```

juniper/CHANGELOG.md
```diff
@@ -45,10 +56,10 @@ See [#618](https://github.com/graphql-rust/juniper/pull/618).
 
 - Remove deprecated `ScalarValue` custom derive (renamed to GraphQLScalarValue)
 
-- `graphql_union!` macro removed, replaced by `#[graphql_union]` proc macro
+- `graphql_union!` macro removed, replaced by `#[graphql_union]` proc macro and custom resolvers of `#[derive(GraphQLUnion)]` macro.
```
LegNeato, line 59:
```suggestion
- `graphql_union!` macro removed, replaced by `#[graphql_union]` proc macro and custom resolvers for the `#[derive(GraphQLUnion)]` macro.
```

juniper/CHANGELOG.md
```diff
@@ -45,10 +56,10 @@ See [#618](https://github.com/graphql-rust/juniper/pull/618).
...
+- `#[derive(GraphQLUnion)]` macro doesn't generate `From` impls for enum variants anymore, consider [`derive_more`](https//docs.rs/derive_more)) crate to do that ([#666]).
```
LegNeato, line 60:
```suggestion
- The `#[derive(GraphQLUnion)]` macro doesn't generate `From` impls for enum variants anymore. Consider using the [`derive_more`](https//docs.rs/derive_more)) crate directly ([#666]).
```

juniper/src/lib.rs
```diff
@@ -111,6 +113,13 @@ extern crate uuid;
 #[cfg(any(test, feature = "bson"))]
 extern crate bson;
 
+// These ones are required for use by the code generated with `juniper_codegen` macros.
```
LegNeato, line 116:
```suggestion
// These are required by the code generated via the `juniper_codegen` macros.
```

docs/book/content/types/unions.md
```diff
@@ -1,75 +1,197 @@
...
+#
+# fn main() {}
+```
 
-impl Character for Human {
-    fn as_human(&self) -> Option<&Human> { Some(&self) }
+
+### Ignoring enum variants
+
+In some rare situations we may want to omit exposing enum variant in GraphQL schema.
```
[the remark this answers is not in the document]
tyranron: @LegNeato hmm... I tried to use `we` instead `you` everywhere to be "on the reader's side" during this "journey", rather than "command him" 🙃

If it's not OK, I'll change the style to use `you` everywhere.

tyranron:
@LegNeato thanks for the corrections, and sorry for my bad English 🙃

> Looks like the CI failures are real.

Yup, tests aren't there yet. Working on it...

jmpunkt:
While reading the documentation I found the behavior of the following example odd.

```rust
#[derive(GraphQLUnion)]
#[graphql(Context = CustomContext)]
#[graphql(on Ewok = Character::ewok_from_context)]
enum Character {
    Human(Human),
    Droid(Droid),
}
impl Character {
    fn ewok_from_context<'c>(&self, ctx: &'c CustomContext) -> Option<&'c Ewok> {
        Some(&ctx.ewok)
    }
}
```
Makes sense that we have custom resolver. But in this case it is not possible to have `Ewok` resolved ever, since it is not part of the enum. I was wondering if it is better to fail if there is no matching variant? The same structure which should compile would be

```rust
#[derive(GraphQLUnion)]
#[graphql(Context = CustomContext)]
#[graphql(on Ewok = Character::ewok_from_context)]
enum Character {
    Human(Human),
    Droid(Droid),
    Ewok,
}
impl Character {
    fn ewok_from_context<'c>(&self, ctx: &'c CustomContext) -> Option<&'c Ewok> {
        Some(&ctx.ewok)
    }
}
```

tyranron:
@jmpunkt thanks! Makes sense. I've [changed the example to reflect that](https://github.com/graphql-rust/juniper/blob/2287679f6b254d22e0d355d8c7920555ab11f3d3/docs/book/content/types/unions.md#custom-resolvers).

tyranron:
@LegNeato @jmpunkt I've also added "`ScalarValue` considerations" section to the union docs in the Book, because this nuance is quite unobvious for newcomers, which are not familiar with Juniper architecture, which results in confusions like [this one](https://github.com/graphql-rust/juniper/issues/467#issuecomment-636401104).

tyranron:
@LegNeato @jmpunkt at this point, I'm pretty statisfied with docs in Book and Rust sources.

Returning to tests to finish this...

tyranron:
Phew... almost there! 🎉

Existing tests are fixed and renewed.

Now, it's only the matter of adding new tests.

docs/book/content/types/unions.md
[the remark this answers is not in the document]
LegNeato: Ok, that works for me.

tyranron:
@LegNeato finally, done! Please review, and I'll squash-merge it once approved.

docs/book/content/types/unions.md
```diff
@@ -1,75 +1,205 @@
-# Unions
+Unions
+======
 
-From a server's point of view, GraphQL unions are similar to interfaces: the
-only exception is that they don't contain fields on their own.
+From a server's point of view, [GraphQL unions][1] are similar to interfaces: the only exception is that they don't contain fields on their own.
```
LegNeato, line 4:
```suggestion
From the server's point of view, [GraphQL unions][1] are similar to interfaces--the only exception is that they don't contain fields on their own.
```

docs/book/content/types/unions.md
```diff
@@ -1,75 +1,205 @@
...
+
+For implementing [GraphQL union][1] Juniper provides:
```
LegNeato, line 6:
```suggestion
For implementing [GraphQL unions][1] Juniper provides:
```

docs/book/content/types/unions.md
```diff
@@ -1,75 +1,205 @@
...
+- `#[derive(GraphQLUnion)]` macro for enums and structs;
```
LegNeato, line 7:
```suggestion
- `#[derive(GraphQLUnion)]` macro for enums and structs.
```

docs/book/content/types/unions.md
```diff
@@ -1,75 +1,205 @@
...
 
-If we look at the same examples as in the interfaces chapter, we see the
-similarities and the tradeoffs:
 
-## Traits
 
-### Downcasting via accessor methods
+## Enums
+
+Most of the time, we need just a trivial and straightforward Rust enum to represent a [GraphQL union][1].
```
LegNeato, line 15:
```suggestion
Most of the time, we just need a trivial and straightforward Rust enum to represent a [GraphQL union][1].
```

docs/book/content/types/unions.md
```diff
@@ -133,87 +306,173 @@ struct Database {
...
+
+impl Character for Human {
+    fn id(&self) -> &str { self.id.as_str() }
+}
+
+impl Character for Droid {
+    fn id(&self) -> &str { self.id.as_str() }
+}
+
+// Used trait object is always `Send` and `Sync`.
```
LegNeato, line 421:
```suggestion
// The trait object is always `Send` and `Sync`.
```

docs/book/content/types/unions.md
```diff
@@ -133,87 +306,173 @@ struct Database {
...
+    fn id(&self) -> &str { self.id.as_str() }
+}
+#
+# fn main() {}
+```
+
+
+### External resolver functions
+
+Similarly to enums and structs, it's not mandatory to use trait methods as [GraphQL union][1] variant resolvers, and instead custom functions may be specified:
```
LegNeato, line 377:
```suggestion
Similarly to enums and structs, it's not mandatory to use trait methods as [GraphQL union][1] variant resolvers. Instead, custom functions may be specified:
```

PR #674, opened 2020-06-04
Upgrade bson crate to 0.15.0

iancormac84:


juniper/src/integrations/bson.rs
```diff
@@ -30,19 +30,19 @@ where
     }
 }
 
-#[crate::graphql_scalar_internal(description = "UtcDateTime")]
-impl<S> GraphQLScalar for UtcDateTime
+#[crate::graphql_scalar_internal(description = "DateTime")]
+impl<S> GraphQLScalar for DateTime
```
tyranron, line 34: @iancormac84 this changes the name of this GraphQL scalar from `UtcDateTime` to `DateTime`. As the scalar name is too common, it may introduce problems, when mixed with other scalars from `chrono`, etc.

Let's keep the scalar name as `UtcDateTime`. This also more aligns with [the spec](http://bsonspec.org/spec.html#Notes).

> - UTC datetime - The int64 is UTC milliseconds since the Unix epoch.

I'd even suggest to this:
```rust
use bson::{oid::ObjectId, DateTime as UtcDateTime};
```

Issue #659, opened 2020-05-12
Async warp example doesn't run


allancalix:
**Describe the bug**
The async warp example server fails to build.

**To Reproduce**
```
# Tested revision
git checkout aedb2d1e
cargo clean
cd examples/warp-async
cargo run
```
**Expected behavior**
An example graphql server should successfully build and start serving on port 8080.

**Additional context**
* Build issue
```
error[E0433]: failed to resolve: could not find `document_join_macro` in `futures_util`
   --> /Users/allancalix/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-0.3.1/src/lib.rs:504:15
    |
504 | futures_util::document_join_macro! {
    |               ^^^^^^^^^^^^^^^^^^^ could not find `document_join_macro` in `futures_util`
```

* The warp version specified in the example manifest is a different major version from the example itself causing issues.

* Some interface changes

I fixed these issues on my fork:
https://github.com/allancalix/juniper/pull/1

PR #677, opened 2020-06-05
Update crate dependencies

jpgneves:
- reqwest 0.10.4 with rustls-tls to remove transitive OpenSSL
dependency
- futures 0.3.5
- proc-macro2 1.0.2

All tests pass with `cargo make ci-flow`.

### DIFF
--- a/juniper_codegen/Cargo.toml
+++ b/juniper_codegen/Cargo.toml
@@ -19,7 +19,7 @@ proc-macro = true
 
 [dependencies]
 proc-macro-error = "1.0.2"
-proc-macro2 = "1.0.1"
+proc-macro2 = "1.0.2"
 quote = "1.0.3"
 syn = { version = "1.0.3", features = ["full", "extra-traits", "parsing"] }
 
--- a/juniper_hyper/Cargo.toml
+++ b/juniper_hyper/Cargo.toml
@@ -14,11 +14,11 @@ url = "2"
 juniper = { version = "0.14.2", default-features = false, path = "../juniper"}
 tokio = "0.2"
 hyper = "0.13"
-futures = { version = "0.3.1" }
+futures = { version = "0.3.5" }
 
 [dev-dependencies]
 pretty_env_logger = "0.2"
-reqwest = "0.9"
+reqwest = { version = "0.10.4", features = [ "rustls-tls", "blocking" ] }
 
 [dev-dependencies.juniper]
 version = "0.14.2"
--- a/juniper_subscriptions/Cargo.toml
+++ b/juniper_subscriptions/Cargo.toml
@@ -10,7 +10,7 @@ edition = "2018"
 
 
 [dependencies]
-futures = "0.3.1"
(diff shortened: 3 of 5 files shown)


juniper_codegen/Cargo.toml
```diff
@@ -19,7 +19,7 @@ proc-macro = true
 
 [dependencies]
 proc-macro-error = "1.0.2"
-proc-macro2 = "1.0.1"
+proc-macro2 = "1.0.2"
```
tyranron, line 22: Cargo by default uses [carret version range](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements), so the specified version is not the exact version being picked up during deps resolution, but rather a most bottom possible one.

That's why there is no reason to bump `proc-macro2` to higher version here, unless:
- we want to use in the library code a feature available only in a newer version of dependency;
- we want to upgrade to a new major version of dependency, which is not covered with this carret version range, i.e. outside `>=1.0.2, <2.0.0`;
- a current the most bottom version of dependency is [yanked](https://doc.rust-lang.org/cargo/commands/cargo-yank.html).

juniper_subscriptions/Cargo.toml
```diff
@@ -10,7 +10,7 @@ edition = "2018"
 
 
 [dependencies]
-futures = "0.3.1"
+futures = { version = "=0.3.5", features = ["compat"] }
```
tyranron, line 13: The same applies here:
- we don't want to have high the most bottom version, unless we really need it in out code;
- nor we want to pin `futures` crate to the exact version `0.3.5`, so introduce any problems with transitive dependencies resolving for library users.

juniper_hyper/Cargo.toml
```diff
@@ -14,11 +14,11 @@ url = "2"
 juniper = { version = "0.14.2", default-features = false, path = "../juniper"}
 tokio = "0.2"
 hyper = "0.13"
-futures = { version = "0.3.1" }
+futures = { version = "0.3.5" }
```
tyranron, line 17: And that one is unnecessary to.

juniper_hyper/Cargo.toml
```diff
@@ -14,11 +14,11 @@ url = "2"
...
 
 [dev-dependencies]
 pretty_env_logger = "0.2"
-reqwest = "0.9"
+reqwest = { version = "0.10.4", features = [ "rustls-tls", "blocking" ] }
```
tyranron, line 21: And this one is good, but I'd rather suggest to use `0.10` as far as this is a dev dependency only, so we anyway will have the latest version executed for tests.

tyranron:
Finished this PR in https://github.com/graphql-rust/juniper/commit/f19d498254d3e5d2a008c483943facaf3ddcc4fe as didn't have permissions to push into this PR's branch.

Also, as side effect, fixed #659, incorporating changes, provided by @allancalix in https://github.com/allancalix/juniper/pull/1.

PR #681, opened 2020-06-14
fix unions with custom scalar values

ccbrown:
The changes introduced in #666 don't quite work for unions where custom scalar values are used.

Explicitly specifying a scalar value type only worked when that type was DefaultScalarValue. This fixes that and adds a test.


tyranron:
@LegNeato the GitHub Actions ICEs of rustc are weird... can't really understand why they happen.

As for nightly builds on Azure Pipelines, it seems that something is wrong with the latest nightly version. Locally, all the the code builds OK and tests pass.

I'm going to merge this anyway, and retry the stuff in `master` on the next nightly update. Seems to be irrelevant to this PR anyway...

Issue #658, opened 2020-05-11
`#[juniper::object]` interfaces attribute not documented


jacobh:
Hi,

this afternoon I've been porting some graphql types to using the attribute macro syntax. I was stuck trying to get my interfaces working correctly.

I think I've figured it out, with `interfaces` now being an attribute of the macro, but I couldn't see any mention of that syntax anywhere in the docs, and had to dig into the source.

Other than this I'm very much liking the attribute macro syntax! reads much more nicely, and rustfmt works correctly.

Old:
```rust
graphql_object!(DynamicQueueContext: Context as "DynamicQueue" |&self| {
    interfaces: [Box<dyn QueueContext>]

    // Queue interface fields
    field id() -> Uuid {
        ContextItem::id(self)
    }
    field name() -> String {
        self.name()
    }
    ...
}
```

New:
```rust
#[juniper::object(
    name = "DynamicQueue"
    Context = Context
    interfaces = [Box<dyn QueueContext>]
)]
impl DynamicQueueContext {
    // Queue interface fields
    fn id(&self) -> Uuid {
        ContextItem::id(self)
    }
    fn name(&self) -> String {
        self.name()
    }
    ...
}
```

tyranron:
Now [documented in Book along with interfaces](https://graphql-rust.github.io/juniper/master/types/interfaces.html).

Issue #605, opened 2020-04-06
Async support for Interfaces is not implemented on master


Ekzer:
Hi there!
I am getting troubles to make this simple piece of code working : 
``` 

#[derive(Debug, Serialize, Deserialize, Eq, PartialEq, juniper::GraphQLObject)]
pub struct Unit {
    id: String,
    duration: i32,
    productionNationalityCode: Vec<String>,
    format: Format
}

#[derive(Debug, Serialize, Deserialize, Eq, PartialEq, juniper::GraphQLObject)]
pub struct Format {
    quality: Vec<String>,
    dst: bool,
    audio: Vec<String>,
    drmType: Vec<String>,
    distTechnology: Vec<String>,
    functionalType: Vec<String>,
    version: Vec<String>,
    comMode: Vec<String>,
    qad: bool
}

trait ContentInterface {
    fn id(&self) -> &str;

    fn as_unit(&self) -> Option<&Unit> { None }
}

impl ContentInterface for Unit {
    fn id(&self) -> &str {
        self.id.as_str()
    }

    fn as_unit(&self) -> Option<&Unit> { Some(self) }
}

juniper::graphql_interface!(<'a> &'a ContentInterface: () as "ContentInterface" where Scalar = <S> |&self| {
    field id() -> &str { self.id() }

    instance_resolvers: |_| {
        // The left hand side indicates the concrete type T, the right hand
        // side should be an expression returning Option<T>
        &Unit => self.as_unit()
    }
});

```

I a getting this error : 
``` 
40 | / juniper::graphql_interface!(<'a> &'a ContentInterface: () as "ContentInterface" where Scalar = <S> |&self| {
41 | |     field id() -> &str { self.id() }
42 | |
43 | |     instance_resolvers: |_| {
...  |
47 | |     }
48 | | });
   | |___^ cannot infer type for type parameter `S`
```

Thank you for your help in advance !
Best regards

LegNeato:
Are you using master? Interfaces have yet to be ported to async master and thus don't currently work on master.

Ekzer:
Hi @LegNeato ,
I am using the async one yes.
`juniper = {git = "https://github.com/graphql-rust/juniper", features=["async"]}`

So I have to go back on classic master if I understood well ?

LegNeato:
Either you need to use the latest release on crates.io (which is sync but has interfaces and unions) or you need to wait until master has support for interfaces (happy to take a PR as well!).

[...]

Issue #421, opened 2019-08-29
Refactor object proc macro attributes to support parameter attributes


theduke:
Function parameter attributes are getting stabilized with `1.39`. (see https://github.com/rust-lang/rust/issues/60406#issuecomment-526327343).

This means we can refactor the argument customization of the `object` proc macro.

This is currently awkward due to the lack of above feature ( see https://graphql-rust.github.io/juniper/current/types/objects/complex_fields.html#customizing-arguments ).

* Supported arguments: `#[graphql(rename = "...", default = expr, description = "...")]`
* Old style should remain supported for now but with a deprecation warning (simple `println` in the macro)
* can be implemented right now with a feature flag, which can be removed once `1.39` is released

davidpdrsn:
I have started hacking in this. Progress is here https://github.com/davidpdrsn/juniper/tree/parameter-attributes

theduke:
@davidpdrsn awesome, please ask if you have any questions.

Also feel free to open a draft PR to give it visibility.

tyranron:
In #682 landed the implementation for GraphQL interfaces.

c410-f3r:
Nice

PR #682, opened 2020-06-15
WIP: 🔬 Make interfaces great again!

### LINKED ISSUE #544: Re-enable interface tests in the book
These were disabled in https://github.com/graphql-rust/juniper/pull/543. Once `juniper::graphql_interface` is implemented, re-enable.

tyranron:
Part of #295
Fixes #544, #605, #658, #748  
Related to #421, #2

...or not so great? 🤔

## Motivation

We want GraphQL interfaces which _just feel right like usual ones_ in many PLs.

~~They should be able to work with an open set of types _without_ requiring a user to specify all the concrete types an interface can downcasts into _on the interface side_.~~  
We're [unable](https://github.com/graphql-rust/juniper/pull/682#issuecomment-644771600) to do that, at the moment 😕

But we still could have a nice proc macros as convenient as they can be 🙃

### Traits

```rust
#[graphql_interface(enum = CharacterValue)]  // <-- use enum for dispatching
#[graphql_interface(for = [Human, Droid, Jedi])]  // <-- mandatory =( and should enumerate all implementors
#[graphql_interface(name = "Character", desc = "Traited character")]
#[graphql_interface(context = MyContext)]  // <-- if not specified, then auto-inferred from methods, if possible, otherwise is `()`
#[graphql_interface(scalar = juniper::DefaultScalarValue)] // <-- if not specified, then generated code is parametrized by `__S: ScalarValue`
#[graphql_interface(on Jedi = custom_downcast_as_jedi)]  // <-- custom external downcaster
trait Character {
    fn id(&self) -> &str;

    // Optional custom downcaster.
    #[graphql_interface(downcast)]
    fn as_droid(&self) -> Option<&Droid> { None }
}

#[graphql_interface(dyn = DynCharacter2, for = Human)]  // use trait object for dispatching
trait Character2 {
    fn id(&self) -> &str;
}

#[derive(GraphQLObject)]
#[graphql(implements = [CharacterValue, dyn Character2], Context = MyContext)]
struct Human {
    id: String,
    home_planet: String,
}
#[graphql_interface]
impl Character for Human {
    fn id(&self) -> &str {
        &self.id
    }
}
#[graphql_interface(dyn)]  // <-- this `dyn` is inevitable when dispatching via trait objects =(
impl Character2 for Human {
    fn id(&self) -> &str {
        &self.id
    }
}

#[derive(GraphQLObject)]
#[graphql(implements = CharacterValue, Context = MyContext)]
struct Droid {
    id: String,
    primary_function: String,
}
#[graphql_interface]
impl Character for Droid {
    fn id(&self) -> &str {
        &self.id
    }

    fn as_droid(&self) -> Option<&Droid> {
        Some(self)
    }
}


#[derive(GraphQLObject)]
#[graphql(implements = CharacterValue, Context = MyContext)]
struct Jedi {
    id: String,
    rank: String,
}
#[graphql_interface]
impl Character for Jedi {
    fn id(&self) -> &str {
        &self.id
    }
}
fn custom_downcast_as_jedi(ch: &CharacterValue, ctx: &MyContext) -> Option<&Jedi> {
     ctx.can_i_haz_the_jedi_please(ch)
}
```

The nearest analogue of GraphQL interfaces are, definitely, traits. However, they are a little different beasts than may seem at first glance. The main difference goes that in GraphQL interface type serves _both_ as an _abstraction_ and a _value dispatching to concrete implementers_, while in Rust, a trait is an _abstraction only_ and you need a _separate type to dispatch to a concrete implemeter_, like enum or trait object, because trait is not even a type itself. This difference imposes a lot of unobvious corner cases when we try to express GraphQL interfaces in Rust. This PR tries to implement a reasonable middleground.

So, poking with GraphQL interfaces will require two parts:
- declare the _abstraction_;
- operate ingerface's _values_.

Declaring an _abstraction_ is almost as simple as declaring trait:
```rust
#[graphql_interface(enum = CharacterValue)]  // <-- specify type ident to represent value of interface
#[graphql_interface(for = [Human, Droid, Jedi])]  // <-- specify implementers
trait Character {
    fn id(&self) -> &str;
}
```

And then, you need to operate in a code with a Rust implementing this GraphQL interface's value:
```rust
#[derive(GraphQLObject)]
#[graphql(implements = CharacterValue)]  // <-- we refer here to Rust type implementing GraphQL interface, not trait
struct Human {
    id: String,
}
#[graphql_interface]
impl Character for Human {  <-- and here we're implementing GraphQL interface for Human Graphql object as a regulat trait impl
    fn id(&self) -> &str {
        &self.id
    }
}
```

#### Trait objects

By default, an enum dispatch is used, because we should enumerate all the implementers anyway.

However, a dyn dispatch via trait is also possible to be used. As this requires a llitle bit more code transformation and aiding during macros expansion, we shoud specify it explicitly both on the interface and the implementers:
```rust
#[graphql_interface(dyn = DynCharacter)]  // <-- generates handy type alias for the trait object
#[graphql_interface(for = [Human, Droid, Jedi])]  // <-- specify implementers
trait Character {
    fn id(&self) -> &str;
}

#[derive(GraphQLObject)]
#[graphql(implements = dyn CharacterValue)]  // <-- `dyn` part is vital here, as an underlying trait is transformed
struct Human {
    id: String,
}
#[graphql_interface(dyn)]  // <-- `dyn` part is vital here, as an underlying trait is transformed
impl Character for Human {
    fn id(&self) -> &str {
        &self.id
    }
}
```

## Checklist

- [x] `derive(GraphQLObject)`/`graphql_object` macros:
    - [x] support `implements` attribute's argument to enumerate all traits this object implements.
- [x] Traits (`#[graphql_interface(dyn)]`):
    - [x] `trait` transformation:
        - [x] support multiple attributes and remove duplicated attributes on first expansion;
        - [x] transform trait to extend `AsGraphQLValue` in a object-safe manner;
            - [x] ensure trait is object safe;
        - [x] enumerate implementors with `#[graphql_interface(for = [Type1, Type2])]` attribute;
            - [x] ensure types are unique on type level;
            - [x] ensure types are GraphQL objects;
        - [x] support custom GraphQL name and description with `#[graphql_interface(name = "Type")]` and `#[graphql_interface(description = "A Type desc.")]` attributes;
        - [x] support custom GraphQL field deprecation with `#[graphql_interface(deprecated = "Text")]` attributes;
        - [x] support custom context type with `#[graphql_interface(context = Type)]` attribute;
            - [x] support custom names for `Context` argument and allow to mark it with `#[graphq_interface(context)]` attribute;
        - [x] support accepting executor in method arguments, optionally marked with `#[graphq_interface(executor)]` attribute;
        - [x] support custom `ScalarValue` type `#[graphql_interface(scalar = Type)]` attribute;
            - [x] allow explicit custom `ScalarValue` type patameter;
        - [x] ignore trait methods with `#[graphql_interface(ignore)]` attribute;
        - [x] support downcasting methods with `#[graphql_interface(downcast)]` attribute;
            - [x] allow omitting `Context` in methods;
            - [x] work correctly with parenthesized types;
            - [x] allow full import paths as types;
            - [x] validate signature for good error messages.
        - [x] support external downcasters on trait itself with `#[graphql_interface(on Type = downcast_fn)]` attribute;
        - [x] ~~support trivial deriving via `#[graphql_interface(derive(Type))]` attribute.~~ (will be implemented in a separate PR)
    - [x] `impl` transformation:
        - [x] ~~support custom context type with `#[graphql_interface(context = Type)]` attribute;~~ (not required)
        - [x] support custom `ScalarValue` type `#[graphql_interface(scalar = Type)]` attribute;
             - [x] allow explicit custom `ScalarValue` type patameter.
- [x] Enums (`#[graphql_interface(enum)]`):
    - [x] generate `enum` type of all implementers;
        - [x] generate `From` impls to this `enum` for each implementer;
        - [x] impl original trait by this `enum`.


tyranron:
@LegNeato @jmpunkt this, at the moment, isn't even a WIP, but rather a searching for a suitable design and a proof-of-concept manual GraphQLType implementation.

I'm feeling being gone too far with it, that's why I just need to share with you my results and discuss the directions...

Interfaces appeared to be _much more tough_ for implementation than [previous work](#666). Really, unions were quite easy. The main pain point about interfaces (applicable to traits only), is that __we don't know all the downstream types we want to resolve into__ at the point when we're implementing `GraphQLType`. This paint point raises 2 main challenges:
1. `GraphQLType` shoud be able to `resolve_field` for an unknown (erased) type. The only possibility to do this in Rust today is trait objects. So, `GraphQLType`, or some its part, _should be_ object safe to have that ability.
2. `GraphQLType::meta` implementation for interfaces should be able to register information in `Registry` about _all its downstream types_. We cannot use trait objects here, as `GraphQLType::meta` calls happen statically.

I've implemented a [PoC implementation here](https://github.com/graphql-rust/juniper/blob/0909a0dcb1868cb0bb065537a7078d96ddb35204/integration_tests/juniper_tests/src/codegen/interface_attr.rs), but it works only a half-way (you can use `cargo test -p juniper_tests` to see the result).

I has turned out, that challenge 1 is not too hard. I've [decoupled `GraphQLType` into two traits](https://github.com/graphql-rust/juniper/blob/0909a0dcb1868cb0bb065537a7078d96ddb35204/juniper/src/types/base.rs#L323-L338): one is object safe, and the other contains all the static stuff. This allows to [resolve `GraphQLType` values dynamically](https://github.com/graphql-rust/juniper/blob/0909a0dcb1868cb0bb065537a7078d96ddb35204/integration_tests/juniper_tests/src/codegen/interface_attr.rs#L194), and resolving an interface fields in tests "just works".

But, I've kinda stuck on the challenge 2.  
First, I've tried to use the `inventory` for meta registration, as I do it for `GraphQLIsObjectType::mark()` calls: [1](https://github.com/graphql-rust/juniper/blob/0909a0dcb1868cb0bb065537a7078d96ddb35204/integration_tests/juniper_tests/src/codegen/interface_attr.rs#L25), [2](https://github.com/graphql-rust/juniper/blob/0909a0dcb1868cb0bb065537a7078d96ddb35204/integration_tests/juniper_tests/src/codegen/interface_attr.rs#L130-L134).  
But... the problem is that `S: ScalarValue` is a generic and `Registry` is abstracted over it. So, our function, registered in `inventory`, should be able to work with an arbitrary `ScalarValue`, but we cannot simply put the generic into an `inventory`: all types should be known to do it.  
And starting from here I have 2 strategies to investigate:

1) Try to invent some any+downcast+lazy_init magic to be able to call `GraphQLType::meta` for all downstream types registered somewhere separately in an `inventory`.  
Still, haven't found a way to it, though... 🤔 And poking with Any and downcasting doesn't give much hope with that.

2) Try to remove `S: ScalarValue` type parameter from `Registry`.  
This, well, is a way _huge_ refactoring. I've investigated where this `S` comes from, and it turned out that it's only required by a [`default_value` field of `meta::Argument`](https://github.com/graphql-rust/juniper/blob/0909a0dcb1868cb0bb065537a7078d96ddb35204/juniper/src/schema/meta.rs#L190). Once we remove it, the whole `Registry` can drop an `S` type parameter from its type. This does mean, that instead of abstracting like `Registry<S>::get_type<T>()` we will be able to abstract like `Registry::get_type<T, S>()`, and so remain generic over `S` without polluting the whole `Registry` and dependent types' signatures with it. This should allow to register in an `inventory` a single function and call it, like we do it with `GraphQLIsObjectType::mark()`. In theory, this also will reduce the codebase complexity and compliation times.  
Another question is: how to do that properly? I'll share thought about that a little bit later after more investigations.

LegNeato:
Thanks for doing this! I've been thinking about this a lot and landed on something very similar. I think we need some sort of side channel to pass information at compile time to not have a central registry. I believe a combination of naming convention, trait bounds, dynamic dispatch, and static functions could be put together in some way to build that side channel, but I have yet to come up with it.

tyranron:
@LegNeato @jmpunkt meh... after playing/poking with it much more, I can conclude that challenge 2 is impossible to be resolved at the current state of Rust and its ecosystem. Even if we've found a way to deal with that generic `ScalarValue`, the main problem is that we just cannot use [`inventory`](http://docs.rs/inventory), because it doesn't work on WASM target at all, nor [`linkme` does](https://github.com/dtolnay/linkme/issues/6), nor [`ctor` does](https://github.com/mmastrac/rust-ctor/issues/14), which is [blocked by `wasm-bindgen`](https://github.com/mmastrac/rust-ctor/issues/14), which, in turn, blocked by `rustc` and "who knows when and whether it will be resolved at all". The price of dropping the whole WASM target is too high for having this feature. And without such things we cannot do any distributed implicit stuff registration.

So, the library user will have to enumerate all the interface's implementors on the interface declaration. I'll try to do it as much ergonomic as possible. I'll update the design in description shortly.  
If somebody will come up with a resonable and working solution for distributed implicit plugins registration, or `ctor`/`linkme` will become usable on WASM and less hacky/unstable, we will be able to drop this requirement in future.

Regarding the object safety, I'll decouple those changes into a separate PR, to not pollute this one with a quite big side-refactoring.

LegNeato:
https://github.com/dtolnay/inventory looks pretty sweet. How about we do one method for wasm (without ctor stuff and needing a central listing) and one for non-wasm?

LegNeato:
Just thinking out loud, I wonder...is there some way we can store similar info to inventory? The limitation with it is it requires runtime support for ctors (which wasm doesn't have). But we don't need runtime, we need to aggregate all implementors at compile-time and output code that is than compiled and run (and can be simple enums). Hmmm

LegNeato:
I know it was suggested before and it is gross, but reading/writing data to `CARGO_TARGET_DIR` via the proc macro might be an acceptable stopgap. It's gross but it is at compile-time and would be an implementation detail we later swap out.

LegNeato:
We're, but we are back to incremental compilation not working. Lame.

LegNeato:
More post-beer musings. What about using naming convention as a pre-determined sidechannel? Something like (massively simplified):

```rust


// #[juniper::graphql_interface]
trait Rideable {
    fn num_wheels() -> u8;
}


// Generated
trait crate::GraphQLInterfaceRideable: Rideable {
    fn as_graphql_rideable(&self) -> &Self {
        &self
    }
}

struct Bicycle;

// #[juniper::graphql_interface]
impl Rideable for Bicycle {
    fn num_wheels() -> u8 {
        2
    }
}

// Generated
impl crate::GraphQLInterfaceRideable for Bicycle {
    fn as_graphql_rideable(&self) -> &Self {
        &self
    }
}

// Juniper code assumes `crate::GraphQLInterfaceRideable` is implemented for any trait named "Rideable" outside of the type system.

```

So basically inverting control...the user's code would define a trait like `GraphQLInterface[X]` via proc macro codegen and juniper would just expect it is there using `Any`. While still runtime and not static checks, the contract would be between `juniper` and `juniper_codegen`'s output...both sides of which we control and could test for.

Anyway, pretty sure we run in some of the same problems above...but if we control both sides I feel like we should be able to do this is some way.

tyranron:
@LegNeato returning to this...

> How about we do one method for wasm (without ctor stuff and needing a central listing) and one for non-wasm?

Yeap, I guess we can hide such thing behind a feature flag. But, I'd rather do it as a separate PR on top of this one, which is going to contain only the base "works everywhere" implementation.

> But we don't need runtime, we need to aggregate all implementors at compile-time and output code that is than compiled and run (and can be simple enums).

However, `ctor`/`invetory`/`linkme` works in runtime only ("life before main").

> reading/writing data to CARGO_TARGET_DIR via the proc macro might be an acceptable stopgap. It's gross but it is at compile-time and would be an implementation detail we later swap out.

> But then we are back to incremental compilation not working.

That requires further investigation. Maybe there will be some way with invalidation. Can't say for sure atm.

> What about using naming convention as a pre-determined sidechannel?

Maybe I haven't understood your idea completely, but it seems to me that this only shifts the problem to another point (another interface declaration), rather than solves it.

The initial problem is that we should enumerate all the trait implementors statically, without specifying all of them manually in the definition. Those implementors potentially may live even in separate crates. I don't know the way to register such distributed implementors except the "distributed plugin registration" provided by `inventory`/`ctor`/`linkme`. Having convetionally named sub-trait is OK, but I don't see how it allows to enumerate its implementors.

tyranron:
Well, at this point, the implementation is very raw and unpolished, but the hard part is done and it just works:
- correctly transforms interface definition considering asyncness;
- generates `GraphQLType`/etc impls for trait object which fully resolve both concrete types and fields.

🎉🎉🎉🎉

Only the routine left: polishing, edge cases, tweaking with external functions, tests, docs, old code elimination.

And, except those, there is one more design feature to be done: allow to piggyback implementation dispathcing via enum, paired with trait (because dyn dispatch is not what we want always).

I'll update the PR description later with the progress...

LegNeato:
This is such great work!!!

LegNeato:
Is there some disconnected work outstanding I can help with? Don't want to get in your way but also want to help out where I can.

tyranron:
@LegNeato thanks for the offering, but I dunno, really.  At the moment it's quite a mess-in-the-middle, which itself will take quite a time to be partioned for delegation.

There are two main issues, which block our release:
1. this one with async interfaces;
2. and the issue with using `DefaultScalaraValue` in `#[graphql_object]` macro.

The second one is really annoying, and we have a lot of issues because of that, so I'm very sure (even insist) that it should be fixed before releasing.

If you could handle the second one, it would be very nice! But it's still a big one, though.

Despite that, I think we can review the current codebase, tests and docs and slightly polish them here and there before release. There can always be found something for improvement. Especially, I think subscriptions deserve much more love.

And, as a side node, after the release, there would be some subsequent breaking releases, because we need to finish the macros story and make it fully consistent, which will require BCs.

LegNeato:
Yeah, I have been slammed at work but planning on putting in some polish for the release. Ok, carry on and hopefully I can nibble at the edges to help.

tyranron:
Update!

Well... all the features are almost complete. Only left:
- executor passing to fields (if one will need lookahead stuff);
- piggybacking dispatch to enum.

After that, only the polishing is left:
- refactor and polish implementation code;
- eliminate old code;
- add codegen failure tests;
- add documentation and update the chapter in book.

tyranron:
Update!

All the features are implemented! The basic set of codegen test is written too.

Only left:
- code polishing;
- codegen failure tests;
- documentation and book.

tyranron:
Update!

Code polishing and code documentation is done!

Only left:
- codegen failure tests;
- documentation in book.

LegNeato:
Let's land this and play around with it? I'm worried about having to rebase while we iterate and this is materially better than what we currently have.

Thanks again for all your work!

LegNeato (approved):

tyranron:
@LegNeato that hasn't been finished yet! 😱 
The book is still weird and incomplete, and I've planned to add some more tests. It was a matter of 1-2 days, or so.

tyranron:
Fixed Book in https://github.com/graphql-rust/juniper/commit/b1a03661122c0fa5c8a4fc41418dee7f9ede6308

davidpdrsn:
Amazing work @tyranron! 🎊

PR #683, opened 2020-06-24
fix: endless poll on an errored ws stream

spacemeowx2:
The `sink_rx` split from `websocket` will output endless Err(_) when error happened. Causing 100% cpu usage.

Use `try_fold` to return when the first error occurred.

### DIFF
--- a/juniper_warp/src/lib.rs
+++ b/juniper_warp/src/lib.rs
@@ -420,7 +420,7 @@ pub mod subscriptions {
         },
     };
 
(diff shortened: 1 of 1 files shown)


spacemeowx2:
Thanks for pointing out. Will do

spacemeowx2:
And I think it's hard to test this function since this function accepts a `warp::ws::WebSocket`. The `warp::test::WsClient` can not send `Err` to the other side.

Issue #623, opened 2020-04-17
Field name is changed when prefixing variable with underscore.


arlyon:
**Describe the bug**

It is a common convention in rust code to prefix variables with an underscore if unused. When creating a `graphql_object`, I would like to prototype the interface and keep the compiler happy, so I prefix all the vars with an underscore.

```rust
struct Query;
#[juniper::graphql_object]
impl Query {
    fn blah() -> bool {
        true
    }
    fn fizz(_buzz: String) -> Option<bool> {
        None
    }
};
```

This changes the name of the `buzz` parameter in the schema. Without the underscore, the parameter is called `buzz`, but with it it is called `Buzz`. This is due to the way the `camelCase` is parsed here:

https://github.com/graphql-rust/juniper/blob/a05f4e55c41d4a5a912334a105a73bd33f0ff904/juniper_codegen/src/util/mod.rs#L252-L270

**Expected behavior**

I would expect that prefixing with an underscore would not change the variable name. Would fixing this cause any breakages?

arlyon:
Seems from the tests that that is the desired behaviour so I suppose the next step is understanding the motivations

LegNeato:
This is odd behavior, I'm inclined to believe it is not addressing any concrete functionality. dropping a leading "_" means you can't name your input object starting with "_" but I think that's probably ok and it is better to align to Rust's semantics...what do you think?

arlyon:
I agree and can do a PR (it's really quite a simple fix) but the only issue is that this may break code in the off-change people are using `_x` anywhere

LegNeato:
The next release changes everything (yay async!) so I am fine with that as long as we put it in the changeling.

arlyon:
I think the best course of action is to just add an additional function (in the same place) `to_graphql_identifier` which strips a single leading _ and passes it on to ``to_camel_case`, otherwise it may be a little misleading. PR coming soon.

arlyon:
Closed in #684

PR #684, opened 2020-06-25
Update to_camel_case to respect the rust convention of prefixing '_'

arlyon:
Rust (and clippy) advocate prefixing unused vars with an underscore. If doing that with juniper,
the expectation (at least personally) is that the variable name won't change. With the current
`to_camel_case` code a leading underscore capitalises. This prevents that by checking for and removing
it.

This is potentially (but unlikely) breaking. Anyone who _does_ have things with a leading underscore can simply capitalise the var to get the same effect, which is enough for me to call it a bug.

Some more discussion is in https://github.com/graphql-rust/juniper/issues/623

There was a similar (unmerged) PR for a different graphql project here: graphql-python/graphene#492 that had the same thought process.

### DIFF
--- a/juniper/src/util.rs
+++ b/juniper/src/util.rs
@@ -7,7 +7,13 @@ use std::borrow::Cow;
 pub fn to_camel_case(s: &'_ str) -> Cow<'_, str> {
     let mut dest = Cow::Borrowed(s);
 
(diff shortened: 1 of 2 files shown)


LegNeato:
This is great! Can we get it added to the changeling? Thanks again!

PR #688, opened 2020-06-29
Strip redundant Send/Sync bounds

tyranron:
## Background

At the moment, in `master` branch, we have `Send`/`Sync` requirements just on everything `async`, oftenly without any real necessity in that.

## What this PR does

Strips redundant `Send`/`Sync` bounds as much as possible, leaving them only there were they are really required.

Also, tries to lift such requirements, if possible, by refactoring the code.

Additionally, removes some boxing in `async` resolving code via swithicng to in-place created enum-future.


tyranron:
@LegNeato it turns out that this PR won't ship something really significant or useful. More likely a cosmetic refactoring, and a bit of `Box`es removing.

But doing this let me analyze `async` stuff implementation quite well, and now I see the ways how more `Send`/`Sync` stuff could be relaxed and `Box`ing avoided via some manual `Future`s/`Stream`s implementations. This will be quite a big task, so I'm going to left it for another PR, after codegen stuff is up.

PR #696, opened 2020-07-11
failure removed in favour of thiserror and anyhow

kunjee17:
Failure package is no more maintained. So, I have removed that in favor of thiserror and anyhow. Both package is suggested by Failure maintainer for replacement. 

This is my first PR for this project. So, feedback is always welcome.


kunjee17:
Thanks @LegNeato . I m little new to Rust and Juniper. Currently trying Juniper in couple personal project. Hope I could bring some goodies to Juniper in future from my experience in web development.

PR #707, opened 2020-07-18
Fix juniper_warp subscriptions

ccbrown:
* Previously, the `juniper_warp` crate would not even compile with the "subscriptions" feature due to a missing `use anyhow::anyhow`. This PR adds that.
* The `context` argument to `graphql_subscriptions` previously required `Clone`. That was unnecessary and made the function unusable with perfectly valid context types. This PR just removes that constraint.
* Previously, as soon as a subscription was stopped, the "got_close_signal" boolean would be set, and the WebSocket connection would be rendered unusable. This meant that multiple subscriptions on a single WebSocket connection were completely broken (and worse yet, completely silent). This PR adds a `HashMap` of subscription states (currently just an atomic boolean) to make it possible to stop subscriptions without interrupting others or breaking the entire connection.

This entire module (`juniper_warp::subscriptions`) probably needs a rewrite, but these fixes make it at least work and are hopefully uncontentious enough to be merged in some form sooner rather than later.

Also of note: If CI were to build the example code, it would have caught at least the first issue. :-/

### DIFF
--- a/juniper_warp/src/lib.rs
+++ b/juniper_warp/src/lib.rs
@@ -440,7 +440,7 @@ pub mod subscriptions {
         Mutation::TypeInfo: Send + Sync,
         Subscription: juniper::GraphQLSubscriptionType<S, Context = CtxT> + Send + 'static,
         Subscription::TypeInfo: Send + Sync,
-        CtxT: Clone + Send + Sync + 'static,
+        CtxT: Send + Sync + 'static,
         S: ScalarValue + Send + Sync + 'static,
     {
         let (sink_tx, sink_rx) = websocket.split();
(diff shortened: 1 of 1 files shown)


LegNeato:
Thanks so much 🍻! I don't use subscriptions myself so left the implementation up to the community. Happy to take PRs for fixes or tests or rewrites.

I want to get around to using these and polishing them but not sure when I will have time.

ccbrown:
Thanks for reviewing! I think I'm pretty likely to attempt a rewrite at some point if no one beats me to it. I've created a "wishlist" issue to document the things that should be addressed when that time comes: https://github.com/graphql-rust/juniper/issues/709

Issue #705, opened 2020-07-16
Subscriptions cannot be multiplexed over one websocket


jerel:
**Describe the bug**
Multiple subscriptions should be possible over one WebSocket connection, however the best I can prove this does not work with the Juniper implementation.

**To Reproduce**

Using the `warp_subscriptions` example you can set up a websocket connection in a Dev Tools console:
```
var ws = new WebSocket('ws://127.0.0.1:8080/subscriptions', 'graphql-ws')
ws.onmessage = (a) => console.log(a) 
ws.send(JSON.stringify({"type":"connection_init","payload":{}}))
ws.send(JSON.stringify({"id":"1","type":"start","payload":{"variables":{},"extensions":{},"operationName":null,"query":"subscription {\n users {\n id \n}\n}"}}))
ws.send(JSON.stringify({"id":"2","type":"start","payload":{"variables":{},"extensions":{},"operationName":null,"query":"subscription {\n users {\n name \n}\n}"}}))
```
In other frameworks (tested in Node) this begins streaming messages for each subscription id:

```
{"type":"data","id":"1","payload":{"data":{"users":{"id":1}}} }
{"type":"data","id":"2","payload":{"data":{"users":{"name":"stream user"}}} }
{"type":"data","id":"1","payload":{"data":{"users":{"id":2}}} }
```
but in Juniper only the first subscription appears to work:
```
{"type":"data","id":"1","payload":{"data":{"users":{"id":1}}} }
{"type":"data","id":"1","payload":{"data":{"users":{"id":2}}} }
{"type":"data","id":"1","payload":{"data":{"users":{"id":3}}} }
```

**Expected behavior**

I expect multiple subscriptions to be possible over a single websocket connection.

**Additional context**

I'm using the `master` branch.

LegNeato:
I believe https://github.com/graphql-rust/juniper/pull/707 fixed this on master. Please confirm!

PR #712, opened 2020-07-20
Add clone() back to request_id to fix compiler error in juniper_warp

jerel:
It's possible that I don't understand the original reasoning behind commit 54be035 removing `.clone()` here but this PR adding it back allows compilation to succeed and completes the fix of #705 via #707 for me. With this change multiple subscriptions now work on master.

Screen Shot 2020-07-20 at 9 50 05 AM


LegNeato:
Thanks! I'll add this to CI to prevent further breakage.

Issue #649, opened 2020-05-02
Implement Subscriptions for juniper-actix


mihai-dinculescu:
It would be awesome to have subscriptions in juniper-actix.

I'm willing to put effort into making it happen, but I will need some guidance.

LegNeato:
That would be great! What guidance do you need?

mihai-dinculescu:
Tbh I think that I'm OK for now.

So far I have:
- got a new examples/actix_subscriptions example up and running based on the [juniper_actix example](https://github.com/graphql-rust/juniper/blob/master/juniper_actix/examples/actix_server.rs)
- Followed the [Actix WS example](https://actix.rs/docs/websockets) to get a Juniper Subscription to connect to the Actix WS

Next, I will look into actually implementing the WS handler. `juniper_warp::subscriptions` and the actix ws examples should help me get there.

LegNeato:
Looks like this landed, thank you!

kamalkech:
@mihai-dinculescu not working correctly for me

tyranron:
@kamalkech this may be quite outdated, please, follow the [example of the latest `juniper_actix` version](https://docs.rs/juniper_actix/latest/juniper_actix/#examples).

PR #716, opened 2020-07-22
Impl subscriptions for juniper_actix

mihai-dinculescu:
Baisc implementation of #649.

I have migrated to juniper_graphql_ws, and interacting with it has been a breeze.

actix-web is not exposing a standard websocket, so we still have to use their ws actor. Far from ideal, but better than before. The actix team is aware of this limitation, and they are considering offering a standard websocket as part of their 3.x release.

- [x] added subscription support to **juniper_actix**
- [x] new **examples/actix_subscriptions**

Usage

```rust
async fn subscriptions(
    req: HttpRequest,
    stream: web::Payload,
    schema: web::Data<Schema>,
) -> Result<HttpResponse, Error> {
    let context = Database::new();
    let schema = schema.into_inner();
    let config = ConnectionConfig::new(context);

    subscriptions_handler(req, stream, schema, config).await
}
```

```rust
# ...
.service(web::resource("/subscriptions").route(web::get().to(subscriptions)))
```


LegNeato:
Just landed some changes that may have invalidated this, please take a look!

mihai-dinculescu:
> Just landed some changes that may have invalidated this, please take a look!

Yep, I will update it!

mbccrack:
what about pub_sub implementation in actix subscription ?

mihai-dinculescu:
> what about pub_sub implementation in actix subscription ?

Pub/sub is definitely a must have imho. But I don't think that it needs to be connected in any way to juniper or to juniper_webserver. Pub/sub logic will most likely sit in the application logic, e.g. in here: https://github.com/graphql-rust/juniper/pull/716/files#diff-651d4667b341d83d495f8eb350e83132R58

With that said, I think that it would be very useful if juniper would provide/expose a very basic, in memory, pub/sub implementation.

mihai-dinculescu:
I have migrated to juniper_graphql_ws, and interacting with it has been a breeze.

actix-web is not exposing a standard websocket, so we still have to use their ws actor. Far from ideal, but better than before. The actix team is aware of this limitation, and they are considering offering a standard websocket as part of their 3.x release.

@LegNeato, @ccbrown, I'd love to get some feedback on this.

juniper_actix/src/lib.rs
```diff
@@ -217,6 +217,265 @@ pub async fn playground_handler(
...
+    }
+
+    /// Errors that can happen while serving a connection.
+    #[derive(Debug)]
+    enum Error {
+        /// Errors that can happen in Actix while serving a connection.
+        Actix(ws::ProtocolError),
+
+        /// Errors that can happen while serializing outgoing messages. Note that errors that occur
+        /// while deserializing internal messages are handled internally by the protocol.
```
ccbrown, line 453: Looks like I made a typo. This was meant to say "while deserializing incoming messages" (now fixed on master):

```suggestion
        /// while deserializing incoming messages are handled internally by the protocol.
```

ccbrown: This comment probably doesn't actually apply at all anyway though as you're using this Error enum for a different purpose and this variant _is_ used for incoming messages.

mihai-dinculescu: Good point! I have simplified the error handling a bit. I hope that it makes more sense now.

examples/actix_subscriptions/src/main.rs
```diff
@@ -0,0 +1,140 @@
...
+    req: HttpRequest,
+    stream: web::Payload,
+    schema: web::Data<Schema>,
+) -> Result<HttpResponse, Error> {
+    let context = Database::new();
+    let schema = schema.into_inner();
+    let config = ConnectionConfig::new(context);
+    // set the keep alive interval to 15 secs so that it doesn't timeout in playground
+    // playground has a hard-coded timeout set to 20 secs
+    let config = config.with_keep_alive_interval(Duration::from_secs(15));
```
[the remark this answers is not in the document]
mihai-dinculescu: I thought about suggesting that, but then I've noticed that 30 seconds is the default for most servers. Playground is at fault here for using an arbitrary timeout value with no mechanism to change it. 

Playground has also been [sunsetted](https://foundation.graphql.org/news/2020/04/03/web-based-graphql-ides-for-the-win-how-why-playground-graphiql-are-joining-forces/), so I don't think it's worth the bother.

juniper_actix/src/lib.rs
```diff
@@ -217,6 +217,256 @@ pub async fn playground_handler(
...
+            let msg = msg.map(|r| Message(r));
+
+            match msg {
+                Ok(msg) => {
+                    let tx = self.graphql_tx.clone();
+
+                    async move {
+                        let mut tx = tx.lock().await;
+                        // Infallible - this should be safe to unwrap
+                        tx.send(msg).await.unwrap();
```
LegNeato, line 339: Let's put an `expect` in here just in case so we can track it down.

juniper_actix/src/lib.rs
```diff
@@ -217,6 +217,256 @@ pub async fn playground_handler(
...
+    impl<S: ScalarValue> std::convert::TryFrom<Message> for ClientMessage<S> {
+        type Error = Error;
+
+        fn try_from(msg: Message) -> Result<Self, Self::Error> {
+            match msg.0 {
+                ws::Message::Text(text) => {
+                    serde_json::from_slice(text.as_bytes()).map_err(|e| Error::Serde(e))
+                }
+                ws::Message::Close(_) => Ok(ClientMessage::ConnectionTerminate),
+                _ => Err(Error::UnexpectedClientMessage),
```
LegNeato, line 441: Are there some messages we should ignore?

From https://actix.rs/actix-web/actix_web_actors/ws/enum.Message.html it seems `nop` at least.

Should/can binary be supported?

mihai-dinculescu: Afaik GraphQL is a text-only protocol. If users want to handle binaries, they will have to base64 encode them as part of the JSON payload.

The `UnexpectedClientMessage` is sort of an ignore. It will produce a `{"type":"connection_error","payload":{"message":"unexpected message received from client"}}` messsage back, but [according to the spec](https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md#gql_connection_error) this will not disconnect the client.

If we want to have real ignores we'll have to bring this logic up into the Actor. Idm that, but maybe we can wait until there's a real use case for it?

juniper_actix/src/lib.rs
mihai-dinculescu: 👍

juniper_actix/src/lib.rs
LegNeato: Makes sense! Apologies as  I don't use actix or subscriptions so don't know the semantics here.

LegNeato:
(I think Github ate this comment previously)

Would love some tests so this doesn't regress. Looks like there is an example test for the websocket stuff at https://github.com/actix/actix-web/blob/master/actix-web-actors/tests/test_ws.rs.

mihai-dinculescu:
> 
> Would love some tests so this doesn't regress. Looks like there is an example test for the websocket stuff at https://github.com/actix/actix-web/blob/master/actix-web-actors/tests/test_ws.rs.

I was going to consider unit tests separately, as warp lacks them too and it will be nice to do something about that as well. Apart from unit tests, it would be nice to also have some sort of integration tests that are independent of the webserver implementation.

But I do get your point. It would be a shame if this PR didn't have at least one unit test. I'll look into adding some in a few days when I get back from camping :)

juniper_actix/src/lib.rs
```diff
@@ -217,6 +217,257 @@ pub async fn playground_handler(
...
+        ) -> Self::Result {
+            let msg = serde_json::to_string(&msg.message);
+
+            match msg {
+                Ok(msg) => {
+                    ctx.text(msg);
+                }
+                Err(e) => {
+                    // TODO: log the event when tracing is merged
+                    ctx.text(format!("error serializing response: {}", e));
```
ccbrown, line 415: I'm a bit inclined to say we should actually close the connection if this ever happens. It seems like it would generally be better to close the connection and let the client re-open it than leave the client waiting for a message that'll never arrive.

So maybe something like this would be appropriate:

```suggestion
                    ctx.close(ws::CloseCode::Error, format!("error serializing response: {}", e));
```

mihai-dinculescu: Yep, I agree. I've just done it. It looks like sending the close frame also closes the connection and stops the actor 👍

PR #717, opened 2020-07-22
Unsubscribe fix

jerel:
The current implementation of `juniper_warp` will kill the websocket at the first `stop` call or if a deserialization error is encountered. This means that any running operation (such as a subscription) that is stopped will require the client to reconnect and set up its subscriptions again.

This fix does two things:
1. ACKs a `connection_init` command
1. Only closes the websocket if the client sends a `connection_terminate` command

Both are modeled after the Apollo implementation https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md


jerel:
@LegNeato yes, this PR is unnecessary due to #721. I'll close for now and if the new implementation from 721 still works this way when I get around to using it I'll re-open an issue or PR based on that.

Issue #709, opened 2020-07-18
juniper_warp::subscriptions rewrite wishlist


ccbrown:
The `juniper_warp::subscriptions` module has been pretty neglected and has enough shortcomings to warrant a complete rewrite imo.

Things that should be improved in such a rewrite:

* `connection_init` handling should parse parameters sent from the client, and these should be exposed in a way that allows users of the library to perform authentication based on them. (Related: https://github.com/graphql-rust/juniper/pull/597)
* The server should respond to `connection_init` with GQL_CONNECTION_ACK + GQL_CONNECTION_KEEP_ALIVE or GQL_CONNECTION_ERROR (e.g. if authentication failed).
* It should be possible to execute queries and mutations over the connection, in addition to subscriptions.
* The server should send periodic keep-alives to the client (possibly with a configurable interval).
* Stopping a subscription should have more well-defined timing characteristics. For example, the values stream should be dropped immediately, and we shouldn't send the GQL_COMPLETE message until we can guarantee that no more events will be sent for that subscription.
* It should probably be possible for the bulk of the code to be framework-agnostic. I.e. the graphql-ws protocol implementation should be usable with frameworks other than Warp.
* The module needs extensive tests.

I think I'm pretty likely to attempt a rewrite if no one else beats me to it, but regardless of who does it, these are the things I would say need to be addressed.

ccbrown:
Starting on this now. Plan is to introduce a new crate that implements the graphql-ws protocol in a reusable way. Then juniper_warp, juniper_rocket, etc. can use that to provide subscription functionality with just a few lines of code.

PR #719, opened 2020-07-25
Simplify SubscriptionConnection

ccbrown:
Previously `SubscriptionConnection` was a stream of `GraphQLResponse<'a, S>`, a newtype for:

```rust
Result<(Value<S>, Vec<ExecutionError<S>>), GraphQLError<'a>>
```

 This doesn't really make much sense because subscription events cannot have any sort of pre-execution errors and the `GraphQLResponse` is always `Ok`. This is a bit misleading and complicates lifetimes as the output of the stream can otherwise be `'static`.

This PR just makes `SubscriptionConnection` a stream of unwrapped tuples. This has the added benefit of allowing the user of the API to actually access the data / errors before serializing. Alternatively, it might make sense to create a new type for this, e.g. an `ExecutionResult` type.

This is something I came across while working on #709 that I figured could go in its own PR.

### DIFF
--- a/juniper/src/lib.rs
+++ b/juniper/src/lib.rs
@@ -186,8 +186,8 @@ pub use crate::{
         marker::{self, GraphQLUnion},
         scalars::{EmptyMutation, EmptySubscription, ID},
         subscriptions::{
-            GraphQLSubscriptionType, GraphQLSubscriptionValue, SubscriptionConnection,
-            SubscriptionCoordinator,
+            ExecutionOutput, GraphQLSubscriptionType, GraphQLSubscriptionValue,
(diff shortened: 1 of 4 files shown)


LegNeato:
`ExecutionResult` sounds like a good idea. In general, I personally prefer to use explicit types rather than tuples where both sides would used a similar number of times...but it isn't a requirement.

ccbrown:
Sounds good, will update shortly.

ccbrown:
@LegNeato Updated. Went with `ExecutionOutput` since it's not a `Result` type and `ExecutionResult` already exists.

PR #721, opened 2020-07-26
GraphQL-WS crate and Warp subscriptions update

### LINKED ISSUE #709: juniper_warp::subscriptions rewrite wishlist

ccbrown:
This add a new crate: `juniper_graphql_ws`. This crate implements the [Apollo graphql-ws protocol](https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md).

This is a replacement for the implementation that was previously in `juniper_warp`. A few of the differences:

* ~Simplifies `SubscriptionConnection`. There is a dedicated PR for this here: https://github.com/graphql-rust/juniper/pull/719. This can be reviewed separately, or that PR can be closed and everything can just be reviewed at once.~ (Edit: This was merged separately.)
* Errors are now correctly emitted for field resolution errors within subscription event streams. Previously they were just being silently dropped inside of `juniper_subscriptions::whole_responses_stream`. This and the first bullet are the only changes outside of the new crate, `juniper_warp::subscriptions` mod, and the `warp_subscriptions` example.
* Types for all graphql-ws message types have been added. This makes a host of things much easier than before, and makes serialization and deserialization much safer and more robust. For example, previously, if a client sent a "}" in a message id, we would do bad things because the implementation was crafting responses using `format!`.
* The server now responds to "connection_init" with the appropriate "ack" + "keep-alive".
* The server now sends keep-alives. These default to a 30 second interval, but are configurable.
* Queries and mutations can now be executed over the same WebSocket connection.
* Stopping a subscription now has well-defined timing. All work is immediately dropped and the "complete" message is not sent until we can guarantee that no other events will be sent.
* We now have more reasonable behavior for "weird things" such as clients using the same id for multiple subscriptions, not sending "init" packets, etc. Clients *must* send an "init" packet before anything else happens, and multiple subscriptions with the same id do not start multiple unstoppable background streams.
* We now correctly send "error" messages for pre-execution errors and "data" packets for errors that happen during execution.
* We now correctly send "connection_error" messages for issues not associated with operations such as malformed client messages.
* There is now a configurable maximum number of operations a single connection can have at any time.
* Parameters from the client's "connection_init" message are now parsed and can be used as parameters for context creation. Because we actually require the "connection_init" message now, no context is required until this point. This enables support for authentication via this message.
* Serialization and deserialization are defined outside of the crate, so users can serialize in the way that's most optimal for them. You don't even have to use JSON.
* There are now unit tests, and lots of them. Because the entire implementation is contained by a single `Sink + Stream` and there are proper message types, writing unit tests is a breeze.

I suspect there are more. The previous implementation was really rough. But on to the usage.

## Usage

I'm just going to copy a unit test:

```rust
#[tokio::test]
async fn test_query() {
    let mut conn = Connection::new(
        new_test_schema(),
        ConnectionConfig::new(Context(1)).with_keep_alive_interval(Duration::from_secs(0)),
    );

    conn.send(ClientMessage::ConnectionInit {
        payload: Variables::default(),
    })
    .await
    .unwrap();

    assert_eq!(ServerMessage::ConnectionAck, conn.next().await.unwrap());

    conn.send(ClientMessage::Start {
        id: "foo".to_string(),
        payload: StartPayload {
            query: "{context}".to_string(),
            variables: Variables::default(),
            operation_name: None,
        },
    })
    .await
    .unwrap();

    assert_eq!(
        ServerMessage::Data {
            id: "foo".to_string(),
            payload: DataPayload {
                data: Value::Object(
                    [("context", Value::Scalar(DefaultScalarValue::Int(1)))]
                        .iter()
                        .cloned()
                        .collect()
                ),
                errors: vec![],
            },
        },
        conn.next().await.unwrap()
    );

    assert_eq!(
        ServerMessage::Complete {
            id: "foo".to_string(),
        },
        conn.next().await.unwrap()
    );
}
```

It couldn't really be any easier.

The second argument to `Connection::new` is generic. Instead of `ConnectionConfig`, you can provide a closure:

```rust
let mut conn = Connection::new(new_test_schema(), |params: Variables| async move {
    Ok(ConnectionConfig::new(Context(1))) as Result<_, Infallible>
});
```

This allows you to defer the configuration until you receive the "connection_init" parameters. Any error returned by this closure is sent back the client.

The entire `juniper_warp::subscriptions` module is now just a few lines to compose the sinks and streams:

```rust
/// Serves the graphql-ws protocol over a WebSocket connection.
///
/// The `init` argument is used to provide the context and additional configuration for
/// connections. This can be a `juniper_graphql_ws::ConnectionConfig` if the context and
/// configuration are already known, or it can be a closure that gets executed asynchronously
/// when the client sends the ConnectionInit message. Using a closure allows you to perform
/// authentication based on the parameters provided by the client.
pub async fn serve_graphql_ws<Query, Mutation, Subscription, CtxT, S, I>(
    websocket: warp::ws::WebSocket,
    root_node: Arc<RootNode<'static, Query, Mutation, Subscription, S>>,
    init: I,
) -> Result<(), Error>
where
    Query: GraphQLTypeAsync<S, Context = CtxT> + Send + 'static,
    Query::TypeInfo: Send + Sync,
    Mutation: GraphQLTypeAsync<S, Context = CtxT> + Send + 'static,
    Mutation::TypeInfo: Send + Sync,
    Subscription: GraphQLSubscriptionType<S, Context = CtxT> + Send + 'static,
    Subscription::TypeInfo: Send + Sync,
    CtxT: Unpin + Send + Sync + 'static,
    S: ScalarValue + Send + Sync + 'static,
    I: Init<S, CtxT> + Send,
{
    let (ws_tx, ws_rx) = websocket.split();
    let (s_tx, s_rx) = Connection::new(ArcSchema(root_node), init).split();

    let ws_rx = ws_rx.map(|r| r.map(|msg| Message(msg)));
    let s_rx = s_rx.map(|msg| {
        serde_json::to_string(&msg)
            .map(|t| warp::ws::Message::text(t))
            .map_err(|e| Error::Serde(e))
    });

    match future::select(
        ws_rx.forward(s_tx.sink_err_into()),
        s_rx.forward(ws_tx.sink_err_into()),
    )
    .await
    {
        Either::Left((r, _)) => r.map_err(|e| e.into()),
        Either::Right((r, _)) => r,
    }
}
```

So adding support for the other frameworks should be pretty trivial now and they can all share one implementation.

## Related Issues

* Depends on / includes https://github.com/graphql-rust/juniper/pull/719
* Obsoletes https://github.com/graphql-rust/juniper/pull/717
* Obsoletes https://github.com/graphql-rust/juniper/pull/617
* Obsoletes https://github.com/graphql-rust/juniper/pull/597
* Simplifies https://github.com/graphql-rust/juniper/pull/716
* Simplifies https://github.com/graphql-rust/juniper/issues/649

* Closes #709

### DIFF
--- a/examples/warp_subscriptions/src/main.rs
+++ b/examples/warp_subscriptions/src/main.rs
@@ -2,10 +2,10 @@
 
 use std::{env, pin::Pin, sync::Arc, time::Duration};
 
-use futures::{Future, FutureExt as _, Stream};
+use futures::{FutureExt as _, Stream};
 use juniper::{DefaultScalarValue, EmptyMutation, FieldError, RootNode};
-use juniper_subscriptions::Coordinator;
-use juniper_warp::{playground_filter, subscriptions::graphql_subscriptions};
+use juniper_graphql_ws::ConnectionConfig;
+use juniper_warp::{playground_filter, subscriptions::serve_graphql_ws};
 use warp::{http::Response, Filter};
 
 #[derive(Clone)]
@@ -134,6 +134,15 @@ fn schema() -> Schema {
     Schema::new(Query, EmptyMutation::new(), Subscription)
 }
 
+async fn on_upgrade(ws: warp::ws::WebSocket, root_node: Arc<Schema>) {
+    serve_graphql_ws(ws, root_node, ConnectionConfig::new(Context {}))
+        .map(|r| {
+            if let Err(e) = r {
+                println!("Websocket error: {}", e);
+            }
+        }).await;
+}
+
 #[tokio::main]
 async fn main() {
     env::set_var("RUST_LOG", "warp_subscriptions");
@@ -151,28 +160,16 @@ async fn main() {
     let qm_state = warp::any().map(move || Context {});
     let qm_graphql_filter = juniper_warp::make_graphql_filter(qm_schema, qm_state.boxed());
 
-    let sub_state = warp::any().map(move || Context {});
-    let coordinator = Arc::new(juniper_subscriptions::Coordinator::new(schema()));
+    let root_node = Arc::new(schema());
 
     log::info!("Listening on 127.0.0.1:8080");
 
     let routes = (warp::path("subscriptions")
         .and(warp::ws())
-        .and(sub_state.clone())
-        .and(warp::any().map(move || Arc::clone(&coordinator)))
         .map(
-            |ws: warp::ws::Ws,
-             ctx: Context,
-             coordinator: Arc<Coordinator<'static, _, _, _, _, _>>| {
-                ws.on_upgrade(|websocket| -> Pin<Box<dyn Future<Output = ()> + Send>> {
-                    graphql_subscriptions(websocket, coordinator, ctx)
-                        .map(|r| {
-                            if let Err(e) = r {
-                                println!("Websocket error: {}", e);
-                            }
-                        })
-                        .boxed()
-                })
+            move |ws: warp::ws::Ws| {
+                let root_node = root_node.clone();
+                ws.on_upgrade(move |websocket| on_upgrade(websocket, root_node.clone()))
             },
         ))
     .map(|reply| {
--- a/examples/warp_subscriptions/src/main.rs
+++ b/examples/warp_subscriptions/src/main.rs
@@ -134,15 +134,6 @@ fn schema() -> Schema {
     Schema::new(Query, EmptyMutation::new(), Subscription)
 }
 
-async fn on_upgrade(ws: warp::ws::WebSocket, root_node: Arc<Schema>) {
-    serve_graphql_ws(ws, root_node, ConnectionConfig::new(Context {}))
-        .map(|r| {
-            if let Err(e) = r {
-                println!("Websocket error: {}", e);
-            }
-        }).await;
-}
-
 #[tokio::main]
 async fn main() {
     env::set_var("RUST_LOG", "warp_subscriptions");
@@ -166,12 +157,18 @@ async fn main() {
 
     let routes = (warp::path("subscriptions")
         .and(warp::ws())
-        .map(
-            move |ws: warp::ws::Ws| {
-                let root_node = root_node.clone();
-                ws.on_upgrade(move |websocket| on_upgrade(websocket, root_node.clone()))
-            },
-        ))
+        .map(move |ws: warp::ws::Ws| {
+            let root_node = root_node.clone();
+            ws.on_upgrade(move |websocket| async move {
+                serve_graphql_ws(websocket, root_node, ConnectionConfig::new(Context {}))
+                    .map(|r| {
+                        if let Err(e) = r {
+                            println!("Websocket error: {}", e);
+                        }
+                    })
+                    .await
+            })
+        }))
     .map(|reply| {
         // TODO#584: remove this workaround
         warp::reply::with_header(reply, "Sec-WebSocket-Protocol", "graphql-ws")
--- a/juniper_graphql_ws/Cargo.toml
+++ b/juniper_graphql_ws/Cargo.toml
@@ -6,7 +6,7 @@ license = "BSD-2-Clause"
 description = "Graphql-ws protocol implementation for Juniper"
 documentation = "https://docs.rs/juniper_graphql_ws"
 repository = "https://github.com/graphql-rust/juniper"
-keywords = ["graphql-ws"]
+keywords = ["graphql-ws", "juniper", "graphql", "apollo"]
 edition = "2018"
 
 [dependencies]
--- a/juniper_graphql_ws/Cargo.toml
+++ b/juniper_graphql_ws/Cargo.toml
@@ -1,0 +1,19 @@
+[package]
+name = "juniper_graphql_ws"
+version = "0.1.0"
+authors = ["Christopher Brown <ccbrown112@gmail.com>"]
+license = "BSD-2-Clause"
+description = "Graphql-ws protocol implementation for Juniper"
+documentation = "https://docs.rs/juniper_graphql_ws"
+repository = "https://github.com/graphql-rust/juniper"
+keywords = ["graphql-ws"]
+edition = "2018"
+
+[dependencies]
+juniper = { version = "0.14.2", path = "../juniper", default-features = false }
+juniper_subscriptions = { path = "../juniper_subscriptions" }
+serde = { version = "1.0.8", features = ["derive"] }
+tokio = { version = "0.2", features = ["macros", "rt-core", "time"] }
+
+[dev-dependencies]
+serde_json = { version = "1.0.2" }
--- a/juniper_graphql_ws/src/lib.rs
+++ b/juniper_graphql_ws/src/lib.rs
@@ -29,8 +29,7 @@ use juniper::{
         task::{Context, Poll, Waker},
         Sink, Stream,
     },
-    GraphQLError, RuleError, ScalarValue,
-    Variables,
+    GraphQLError, RuleError, ScalarValue, Variables,
 };
 use std::{
     collections::HashMap,
@@ -94,10 +93,6 @@ impl<S: ScalarValue, CtxT: Unpin + Send + 'static> Init<S, CtxT> for ConnectionC
 
 enum Reaction<S: Schema> {
     ServerMessage(ServerMessage<S::ScalarValue>),
-    Activate {
-        config: ConnectionConfig<S::Context>,
-        schema: S,
-    },
     EndStream,
 }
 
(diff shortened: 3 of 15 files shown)


examples/warp_subscriptions/src/main.rs
```diff
@@ -151,30 +151,24 @@ async fn main() {
...
-                            if let Err(e) = r {
-                                println!("Websocket error: {}", e);
-                            }
-                        })
-                        .boxed()
-                })
-            },
-        ))
+        .map(move |ws: warp::ws::Ws| {
+            let root_node = root_node.clone();
```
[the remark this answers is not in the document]
ccbrown: This seemed simpler to me, but if someone with more Warp expertise tells me it would be better for whatever reason (demonstration purposes?), I'm happy to change it.

LegNeato:
Thanks for this! Will look in-depth in a bit. One thing I am worried about is you can use subscriptions _without_ `graphql-ws`...it is not (yet?) part of the spec.

So I want to make sure we have "open-ended" subscription support for people to roll their won protocol as well as practical opinionated support like `graphql-ws`.

ccbrown:
Yep, I'm definitely not reducing functionality of the general purpose crates. You can still use `juniper_subscriptions` and `juniper_warp` without graphql-ws.

juniper_graphql_ws/src/lib.rs
```diff
@@ -0,0 +1,1068 @@
...
+                        if keep_alive_interval > Duration::from_secs(0) {
+                            s = s
+                                .chain(
+                                    Reaction::ServerMessage(ServerMessage::ConnectionKeepAlive)
+                                        .to_stream(),
+                                )
+                                .boxed();
+                            s = s
+                                .chain(stream::unfold((), move |_| async move {
+                                    tokio::time::delay_for(keep_alive_interval).await;
```
LegNeato, line 180: It is a huge bummer to use tokio just for a delay...AFAICT everything else is executor-agnostic. Is there a generic delay that only depends on `futures`?

juniper_graphql_ws/src/lib.rs
```diff
@@ -0,0 +1,1068 @@
...
+                )
+            }
+            Self::Terminated => (self, stream::empty().boxed()),
+        }
+    }
+
+    async fn start(id: String, params: ExecutionParams<S>) -> BoxStream<'static, Reaction<S>> {
+        // TODO: This could be made more efficient if juniper exposed functionality to allow us to
+        // parse and validate the query, determine whether it's a subscription, and then execute
+        // it. For now, the query gets parsed and validated twice.
```
LegNeato, line 303: Can you file an issue to track?

juniper_graphql_ws/src/lib.rs
```diff
@@ -0,0 +1,1068 @@
+/*!
+
+# juniper_graphql_ws
+
+This crate contains an implementation of the graphql-ws protocol, as used by Apollo.
```
LegNeato, line 5: Let's add a link to the proposed spec.

juniper_graphql_ws/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "juniper_graphql_ws"
+version = "0.1.0"
+authors = ["Christopher Brown <ccbrown112@gmail.com>"]
+license = "BSD-2-Clause"
+description = "Graphql-ws protocol implementation for Juniper"
+documentation = "https://docs.rs/juniper_graphql_ws"
+repository = "https://github.com/graphql-rust/juniper"
+keywords = ["graphql-ws"]
```
LegNeato, line 9: Let's add "juniper", "GraphQL", and "Apollo"

juniper_graphql_ws/src/lib.rs
```diff
@@ -0,0 +1,1071 @@
...
+{
+    type Error = E;
+    type Future = Fut;
+
+    fn init(self, params: Variables<S>) -> Fut {
+        self(params)
+    }
+}
+
+enum ConnectionState<S: Schema, I: Init<S::ScalarValue, S::Context>> {
```
LegNeato, line 136: This seems like a great opportunity to use the typestate pattern to make sure all state transitions are valid and checked at compile-time:

http://cliffle.com/blog/rust-typestate/#:~:text=The%20typestate%20pattern%20is%20an,in%20its%20compile-time%20type.

juniper_graphql_ws/src/lib.rs
```diff
@@ -0,0 +1,1071 @@
...
+                            stoppers.retain(|_, tx| !tx.is_canceled());
+
+                            if config.max_in_flight_operations > 0
+                                && stoppers.len() >= config.max_in_flight_operations
+                            {
+                                // Too many in-flight operations. Just send back a validation error.
+                                stream::iter(vec![
+                                    Reaction::ServerMessage(ServerMessage::Error {
+                                        id: id.clone(),
+                                        payload: GraphQLError::ValidationError(vec![
```
LegNeato, line 235: Should we add a new error type here? I guess it is technically validation...

juniper_graphql_ws/src/lib.rs
```diff
@@ -0,0 +1,1071 @@
...
+                    id: id.clone(),
+                    payload: DataPayload { data, errors },
+                })
+                .to_stream();
+            }
+            Err(GraphQLError::IsSubscription) => {}
+            Err(e) => {
+                return Reaction::ServerMessage(ServerMessage::Error {
+                    id: id.clone(),
+                    payload: unsafe { ErrorPayload::new_unchecked(Box::new(params.clone()), e) },
```
LegNeato, line 332: Please put a comment why this is needed and safe to do.

juniper_graphql_ws/src/lib.rs
ccbrown: Agreed. I looked for a bit before adding the tokio dependency, but couldn't find anything. :(

juniper_graphql_ws/src/lib.rs
ccbrown: Sure thing. Will do once merged.

juniper_graphql_ws/src/lib.rs
ccbrown: Done. Side note: It might be worth adding an owned version of `GraphQLError` or giving `GraphQLError` some getters that would allow outside crates to create an owned error structure. Doing so would let me eliminate all except one `unsafe` in this PR. I was trying to be conservative with changes to the `juniper` crate here though.

juniper_graphql_ws/src/lib.rs
ccbrown: 🤷 I don't really feel strongly either way.

juniper_graphql_ws/src/lib.rs
ccbrown: Hmm... I experimented with a few things and couldn't figure out how to increase compiler coverage while keeping this elegant. The pre-init state effectively contains a `FnOnce`, which already gives us just about all the compile-time guarantees we can make. In order to invoke that, we _have_ to consume the state and move to a different one, of which all are valid.

ccbrown:
@LegNeato mind taking another look over the last few commits? I added what appeared necessary to make CI happy, but I'm not super confident as I don't know much about the repo's release process.

If everything looks good to you, I think this can be merged now.

LegNeato:
Looks great! 🚀 🍾

tyranron:
@ccbrown does this GraphQL over WebSocket protocol implementation supports queries and mutations too? Or just subscriptions?

ccbrown:
> @ccbrown does this GraphQL over WebSocket protocol implementation support queries and mutations too? Or just subscriptions?

It supports queries and mutations too.

PR #730, opened 2020-08-01
Use the standard subscription keep-alive message

mihai-dinculescu:
Use the [standard subscription keep-alive message](https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md#gql_connection_keep_alive) so that the clients can handle it automatically.


mihai-dinculescu:
Hmm, that certainly complicates things. I see that the Python library deviates from the standard as well.

I have noticed the issue in the GQL Playground. The UI is not keeping the connection alive. With the change in this PR it does. Obviously, supporting Apollo and the Apollo compliant clients is more important than making the Playground happy so I'm gonna close this PR and look into other ways of making it work in the Playground.

PR #736, opened 2020-08-12
Actix subscriptions tests

mihai-dinculescu:
A tiny subscriptions tests framework. 

An array of messages (to be sent or to be expected) is created and then given to the runner. This should work nicely with any web server. An example can be found below.

```rust
async fn test_ws_simple_subscription<T: WsIntegration>(integration: &T) {
    let messages = vec![
        WsIntegrationMessage::Send(
            r#"{
                "type":"connection_init",
                "payload":{}
            }"#
            .to_owned(),
        ),
        WsIntegrationMessage::Expect(
            r#"{
                "type":"connection_ack"
            }"#
            .to_owned(),
            WS_INTEGRATION_EXPECT_DEFAULT_TIMEOUT,
        ),
        WsIntegrationMessage::Expect(
            r#"{
                "type":"ka"
            }"#
            .to_owned(),
            WS_INTEGRATION_EXPECT_DEFAULT_TIMEOUT,
        ),
        WsIntegrationMessage::Send(
            r#"{
                "id":"1",
                "type":"start",
                "payload":{
                    "variables":{},
                    "extensions":{},
                    "operationName":null,
                    "query":"subscription { asyncHuman { id, name, homePlanet } }"
                }
            }"#
            .to_owned(),
        ),
        WsIntegrationMessage::Expect(
            r#"{
                "type":"data",
                "id":"1",
                "payload":{
                    "data":{
                        "asyncHuman":{
                            "id":"stream id",
                            "name":"stream name",
                            "homePlanet":"stream home planet"
                        }
                    }
                }
            }"#
            .to_owned(),
            WS_INTEGRATION_EXPECT_DEFAULT_TIMEOUT,
        ),
    ];

    integration.run(messages).await.unwrap();
}
```

### DIFF
--- a/examples/actix_subscriptions/Cargo.toml
+++ b/examples/actix_subscriptions/Cargo.toml
@@ -13,7 +13,7 @@ actix-cors = "0.2.0"
 futures = "0.3.5"
 tokio = { version = "0.2", features = ["rt-core", "macros"] }
 env_logger = "0.7.1"
-serde = "1.0.114"
+serde = "1.0.115"
 serde_json = "1.0.57"
 rand = "0.7.3"
 
--- a/examples/actix_subscriptions/Cargo.toml
+++ b/examples/actix_subscriptions/Cargo.toml
@@ -17,6 +17,6 @@ serde = "1.0.115"
 serde_json = "1.0.57"
 rand = "0.7.3"
 
-juniper = { path = "../../juniper", features = ["expose-test-schema", "serde_json"] }
+juniper = { path = "../../juniper", features = ["expose-test-schema"] }
 juniper_actix = { path = "../../juniper_actix", features = ["subscriptions"] }
 juniper_graphql_ws = { path = "../../juniper_graphql_ws" }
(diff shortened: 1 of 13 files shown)


mihai-dinculescu:
> 
> 
> Thanks for doing this! The `http` mod in juniper is more than just for tests so we can't rely on a dev-only dep like `async_trait`. Is it annoying to use `BoxFuture` over `async_trait`?

Good point. `LocalBoxFuture` did the job nicely.

The same issue is with `anyhow`. Should I look for a workaround (prob defining a custom error type) or maybe it's ok to add anyhow to juniper deps?

LegNeato:
Hmmm, perhaps we should put it in `integration_tests` so it can specify different deps and features in the test's own `Cargo.toml`. What do you think?

mihai-dinculescu:
> 
> 
> Hmmm, perhaps we should put it in `integration_tests` so it can specify different deps and features in the test's own `Cargo.toml`. What do you think?

Ah, silly me. The problem was that anyhow wasn't actually in the dev-dependencies list. Adding it fixes the issue.

I'll still look into separating the test helpers thou.

mihai-dinculescu:
Hmm, it turns out that `test` arm of `#[cfg(any(test, feature = "expose-test-schema"))]` is not needed. Getting rid of it removes the need for `anyhow` in dev-dependencies.

I'm not sure if it's still worth splitting out the http test helpers, @LegNeato ?

PR #747, opened 2020-08-29
Remove unused dependencies

c410-f3r:
`juniper` brings **a lot** of weight and compilation times into my projects and this PR is an attempt to decrease this burden.

Local tests say that `cargo build/check/test` will now have ~8 less crates to download and the used technique can be further extended into other projects (juniper_actix, juniper_warp, etc) to decrease CI times.


LegNeato:
Thank you! This has been on my TODO.

PR #751, opened 2020-09-01
Re-enable markers static checking in generated code

tyranron:
Another portion of changes that is required for #682, but is better to be done as a separate PR.

## Synopsis

On the current master [the code responsible to statically check `marker::IsInputType`/`marker::IsOutputType`](https://github.com/graphql-rust/juniper/blob/2ab00f55d620c51b8732c97c90f8f3349c2e6a0b/juniper_codegen/src/util/mod.rs#L965-L984) impls is commented with the note about requiring interfaces support.

However, in practice, these things are not related. Exiting of this markers can be checked without interfaces support.

## Solution

This PR re-enables these checks back again for all GraphQL types, and adds missing implementations.

## Design changes

An associated types are added to `IntoResolvable` and `IntoResult` traits. They are only used to be able to name these types on type-level.

Also, this PR relaxes `Sized` requirement on implementations of `marker::IsInputType`/`marker::IsOutputType` traits to accept dyn-sized types in future.

And removes `GraphQLType` requirement from `marker::IsOutputType`. The whole purpose for this, is to be able to impl it for `Result<T: IsOutputType, E>`, so we can statically check that subscription return type (which is opaque) to imp `marker::IsOutputType`.  
Such relaxation has no harm for existing code, as `marker::IsOutputType` is used _only_ inside code generated by macros, and the requirement `T: GraphQLType` is fully checked by other parts of generated code.  
Also, it will be re-enabled back in #682, once interfaces codegen will be more compatible withh all the type-level stuff.

### DIFF
--- a/integration_tests/codegen_fail/fail/input-object/derive_incompatible_object.stderr
+++ b/integration_tests/codegen_fail/fail/input-object/derive_incompatible_object.stderr
@@ -1,3 +1,12 @@
+error[E0277]: the trait bound `ObjectA: juniper::marker::IsInputType<__S>` is not satisfied
+ --> $DIR/derive_incompatible_object.rs:6:10
+  |
+6 | #[derive(juniper::GraphQLInputObject)]
+  |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `juniper::marker::IsInputType<__S>` is not implemented for `ObjectA`
+  |
+  = note: required by `juniper::marker::IsInputType::mark`
+  = note: this error originates in a derive macro (in Nightly builds, run with -Z macro-backtrace for more info)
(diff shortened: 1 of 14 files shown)


LegNeato:
Looks like failures are legit.

tyranron:
@LegNeato yup. It's still in draft. Have some specialization troubles, unexpectedly. I'll request your review once it's ready.

LegNeato:
Ah sorry, didn't appear in draft on my phone!

tyranron:
@LegNeato phew! 😓

It was quite untrivial to figure out how to resolve the issue, but some trait-level ~~arcana magic~~ type concretization does the job.

PR #755, opened 2020-09-08
Fix version.

bostontrader:
I have found a number of issues with The Book.  I'm willing to make some pull requests to clean up said issues.  I am submitting this small and simple pull request as a hello-pull-request to get the ball rolling. If you all would rather I amend this to include more issues, then please let me know and I will.

### DIFF
--- a/docs/book/content/quickstart.md
+++ b/docs/book/content/quickstart.md
@@ -10,7 +10,7 @@ Juniper follows a [code-first approach][schema_approach] to defining GraphQL sch
 
 ```toml
 [dependencies]
-juniper = "0.14.2"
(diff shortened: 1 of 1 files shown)


LegNeato (approved):
Hello and thank you! Would love any changes here. This change is great, but I am worried without release automation when we cut a release and push updated docs we will forget to update this to the released version.

bostontrader:
I figure that we have the same problem with the original version string so we have at least not regressed.  Keeping the version string up to date here is part of the general offer to update these docs.  I think there are a lot of devils in the details about how to plug the version automatically that are worthy of further discussion.  Perhaps it would be better to start an issue for that topic.

PR #756, opened 2020-09-09
Reduce warp_subscriptions example to 15 seconds so that GQL clients don't disconnect

### LINKED ISSUE #754: juniper_graphql_ws silently closes the socket
**Describe the bug**
When running the latest `examples/warp_subscriptions` app locally the websocket silently dies after ~20 seconds. I suspect that there's an issue with the keepalive functionality but haven't been able to track it down. If `config.with_keep_alive_interval(0)` is used the socket stays open. If `config.with_keep_alive_interval(Duration::from_secs(15))` is used it also stays open, however if the default is left (30) or it's manually set to > 25 with `with_keep_alive_interval` the socket will die after a few seconds.

**To Reproduce**
1. Run the example app
1. Execute `subscription { users { id } }` in GraphiQL
1. After ~20 seconds the socket will disconnect

**Expected behavior**
The socket should remain open permanently. It would be nice if a log event was generated when the socket closes (crashes?).

jerel:
Closes #754

### DIFF
--- a/juniper_graphql_ws/src/lib.rs
+++ b/juniper_graphql_ws/src/lib.rs
@@ -63,7 +63,7 @@ impl<CtxT> ConnectionConfig<CtxT> {
         Self {
             context,
             max_in_flight_operations: 0,
-            keep_alive_interval: Duration::from_secs(30),
+            keep_alive_interval: Duration::from_secs(15),
         }
     }
 


LegNeato (changes requested):
Instead of updating the example, can you change the default?

LegNeato (approved):
Thank you! 🥂

PR #757, opened 2020-09-10
Fix LookAheadMethods lifetimes

tyranron:
Revealed from #682  

## Synopsis

The `LookAheadSelection` and `ConcreteLookAheadSelection` are mostly reference types: part of their data is a reference to some outer value.

The current `LookAheadMethods` trait (which abstracts public interface to these two selections) cannot express this notion, however. This result in the situation:
```bash
let name: &'static str = "my_string";
let executor: Executor<'static, _> = Executor::create_from(&name);  // just imitating

let field_name: &'static str = executor.look_ahead().field_name();  // fails to compile
```
Compilation fails because `field_name` has the lifetime of a temporary value, returned by `executor.look_ahead()` (which acts as a reference in this case). This happens because `LookAheadMethods::field_name` bounds returned value to the `&self` lifetime, not the lifetime that `Self` points to, as it should be.

## Solution

Introduce a lifetime parameter in `LookAheadMethods` trait, so we could express the referencing to some outer values when implementing this trait.


LegNeato (approved):
Nice catch!

Issue #201, opened 2018-06-22
Ability to create a snake_case field with the graphql_object! macro


z0mbie42:
Hi, as I wasn't able to find any way to do this I open this issue:

a way to create snake_case fields should be provided.

Now we can create an object
```rust
graphql_object!(Human: () |&self| {
    field id() -> Option<&str> {
        self.id.as_str()
    }

    field appearsIN() -> Vec<Episode> {
        vec![Episode::NewHope, Episode::Jedi]
    }
});
```

which results in 
``` graphql
query {
human(id: "1234") {
    id
    appearsIN
  }
}
```
but we can't create the following object due to the automatic field transformation
```graphql
query {
human(id: "1234") {
    id
    appears_in
  }
}
```
indeed
```rust
graphql_object!(Human: () |&self| {
    field id() -> Option<&str> {
        self.id.as_str()
    }

    field appears_in() -> Vec<Episode> {
        vec![Episode::NewHope, Episode::Jedi]
    }
});
```
will results in
```graphql
query {
human(id: "1234") {
    id
    appearsIn
  }
}
```

theduke:
Yup we should probably allow snake case fields even though they are against the convention.

z0mbie42:
the source of the problem come from constructions like this
```rust
$reg.field_convert::<$t, _, Self::Context>(
                &$crate::to_camel_case(stringify!($name)), $info)
                .deprecated($reason),
```
https://github.com/graphql-rust/juniper/blob/master/juniper/src/macros/object.rs#L280

but being not fluent enough in rust, I don't know how to modify the macro to allow snake_case call instead of `to_camel_case`

theduke:
We will need to add some kind of attribute that disables to to_camel_case conversion.

Maybe

```
#[juniper(keep_field_name)]
field appears_in() -> Vec<Episode> {
        vec![Episode::NewHope, Episode::Jedi]
    }
```

kestred:
I disagree that the current behavior should be changed.

Juniper should keep the current behavior (names are coerced into camel case).

Many new users create fields using `snake_case` intuitively, which is part of the current benefits of Juniper that your rust code looks more or less like rust, using rust conventions (instead of diverging completely into a custom DSL), and your GraphQL schema still uses GraphQL conventions.

While field names are always coerced to strings, and can be handled arbitrarily because of that, other parts of the `graphql_xyz!` macros are hard coded and cannot be changed according to preference. 

Example:

```rust
[...]

PR #765, opened 2020-09-16
Allow disabling case convertion

ip1981:
Add a new attribute `rename` which could be "none", "camelCase", "SCREAMING_SNAKE_CASE".
By default (when this attribute is not specified) the exiting rules apply: camelCase to structs, SCREAMING_SNAKE_CASE to enums.

Inspired by Serde & Structopts.

This does not change interfaces because I haven't figured how to do them yet.

Re https://github.com/graphql-rust/juniper/issues/201.

### DIFF
--- a/integration_tests/juniper_tests/src/codegen/derive_enum.rs
+++ b/integration_tests/juniper_tests/src/codegen/derive_enum.rs
@@ -16,6 +16,13 @@ enum SomeEnum {
     Full,
 }
 
+#[derive(juniper::GraphQLEnum, Debug, PartialEq)]
+#[graphql(rename = "none")]
+enum NoRenameEnum {
+    OneVariant,
+    AnotherVariant,
+}
+
 /// Enum doc.
 #[derive(juniper::GraphQLEnum)]
 enum DocEnum {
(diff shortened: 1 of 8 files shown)


LegNeato (changes requested):
Thanks for the PR, this looks great! A couple of changes / suggestions:

1. We'll need a couple of unit tests or integration tests to make sure this isn't broken in the future.
2. I'd prefer to match the API from http://serde.rs. that is, `rename` in a field is for a specific name set verbatim and the ability to use a rename strategy at the top-level is on the object definition.

ip1981:
I'm not sure I totally get the second point.

There are differences between juniper and serde:

1. Juniper in fact uses only one-way renaming, even when decoding input objects. This is why there is no point in having an attribute like `rename_all`, for example.
2. I am trying to keep existing behavior by default (camelCase for structs, SCREAMING_CASE for enums). That's why it is `Option<RenameRule>`.

Each struct member already can be renamed explicitly with the `name` attribute.

ip1981:
I will add some tests for sure :)

LegNeato:
Ugh, sorry...read that way too fast and was fried from work. You're right, ignore #2 :-)

ip1981:
Added some tests.

LegNeato:
Thank you! I'll land this in a couple of days.

mkpankov:
Hey! Any updates on this?

LegNeato (approved):

Issue #726, opened 2020-07-29
Add support for parsing and validating a query without executing it


ccbrown:
Currently, when getting a request that might contain a subscription, there's no way to know if it actually contains a subscription until you try to execute it. This is a problem for example in our graphql-ws implementation, which currently just has to invoke both `juniper::execute` and `juniper::resolve_into_stream` and see what works:

https://github.com/graphql-rust/juniper/blob/84c9720b535c37dfc1d8bc6b142bf2f63e5fd166/juniper_graphql_ws/src/lib.rs#L301-L303

It would be useful if `juniper` had a lower level API that could be used to parse and validate a query, determine what type of operation is requested, _then_ execute it.

LegNeato:
This would probably get rid of duplicate code between `execute` and `execute_sync` too.

zacck:
Hey @LegNeato I would love to take this on, but at  the moment I have no idea where to get started it is ok to ask for instructions on what needs to be done?

PR #773, opened 2020-10-04
WIP: Add parse and validate functions

mirosval:
Fix: #726

Split out parse and validate and then re-use them in all execute functions.

This is a draft, please let me know if this direction is acceptable. Also, the current approach does not actually work in case of `execute` and `resolve_into_stream` because of lifetime issues. 

I would be very happy to get some guidance on this, if this approach works, I'll add:

- [ ] Tests
- [ ] Docs
- [ ] Examples

### DIFF
--- a/juniper/src/lib.rs
+++ b/juniper/src/lib.rs
@@ -168,7 +168,7 @@ use crate::{
 };
 
 pub use crate::{
-    ast::{FromInputValue, InputValue, Selection, ToInputValue, Type},
+    ast::{Document, FromInputValue, InputValue, Operation, Selection, ToInputValue, Type},
     executor::{
         Applies, Context, ExecutionError, ExecutionResult, Executor, FieldError, FieldResult,
         FromContext, IntoFieldError, IntoResolvable, LookAheadArgument, LookAheadMethods,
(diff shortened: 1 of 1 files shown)


ccbrown:
Glad to see someone taking this on! At a high level, I have a few thoughts here, aimed mainly towards simplifying the API:

1. This is up for debate, but I personally don't think the `ValidDocument` type is actually necessary. If a user of the library wants to skip validation that should be fine.

    > An invalid request is still technically executable, and will always produce a stable result as defined by the algorithms in the Execution section

    – https://spec.graphql.org/June2018/#sec-Validation

    The spec goes on to say that execution "should" only occur for valid requests, but I'm not sure we need to enforce that.

2. It's seems pretty off to me for parsing to require the schema as an argument. In every other GraphQL implementation I've looked at, the parsing process just takes in a string and outputs an AST.

    Juniper's parser currently requires the schema to convert scalar tokens to custom scalar values, but I don't see why this has to be done in the parser. Maybe we could move that into the validator and executor instead? If we did that, I don't think `Document` would even require the `ScalarValue` parameter any more.

mirosval:
@ccbrown Thank you for looking into this!

1.  > don't think the ValidDocument type is actually necessary.
I was basing this on the fact that the execute functions are named `execute_validated_query*`. I would then suggest to have 2 APIs: 
* Validate + Execute validated (named just `execute`). This would have a type like `ValidDocument` in between. So we can enforce that you validate the doc up front. It would be the default use case.
* Just execute without validation. This would then be named `execute_unvalidated_query*` or similar. To suggest you're doing something unsafe (since the spec says you might end up with ambiguous or surprising results for unvalidated requests). This is probably not what you want most of the time.

2. Seems like a reasonable idea. I will have to look into it more in-depth, but I would make a separate PR for that.

ccbrown:
This is a lower-level API that only people who know what they're doing should be using to begin with, so I think the hand-holding is unnecessary, but if we do think a type like `ValidDocument` is needed, let's hold off on the second `execute_unvalidated_query` API. That can come in a later PR if someone has a use-case for it. My goal with the suggestion was to simplify the API, not complicate it.

ccbrown:
Btw, these changes don't actually have to be in the `juniper` crate's API. We could satisfy #726 with public APIs in `juniper::parser`, `juniper::validation`, and `juniper::executor`. That would keep the public `juniper` API narrow and make it less likely that someone would stumble upon these and try to use them for something inappropriate.

LegNeato:
I agree with these comments :-)

mirosval:
Okay, thanks! That makes sense to me. I'll close this and create new PRs for the 3 new public APIs.

I spent my evening yesterday investigating the `ScalarValue` type parameter and I understand now what you meant with removing it from `Document`. 
* I think you can delay the parsing and so remove the type parameter. 
* For validation, you should just be able to pass in a validator of sorts that would try to parse the unparsed scalars
* I believe it should also be possible to remove the parameter from execution as well, but I need to look deeper into it

Let me know what you think.

Issue #750, opened 2020-08-31
Expose the fields of GraphQLRequest


sessionboy:
**Describe the solution you'd like**

My database is [dgraph](https://github.com/dgraph-io/dgraph), which is a native graphql database.
I need to query the data of `dgraph graphql database` in the juniper server. This is not as easy as restful.

So i want to get `query`, `operation_name` and `variables` because I want to request data from `dgraph graphql database` based on them. But now they are private.

In addition, how can I obtain resolveInfo? Maybe I need it.

LegNeato:
Hmm, I thought operation name was public, perhaps it got dropped:

https://github.com/graphql-rust/juniper/pull/265/commits

sessionboy:
@LegNeato  I checked the source code, they are private and I cannot get them. Can you update and publish it ?

mkpankov:
I think I need this too. Say user wants a subfield `{ details { foo, bar } }`. I need a different query for PostgreSQL to fetch details. Right now I have a query argument `detail: bool`, but I really should determine if user asked for details and change logic automatically.

PR #776, opened 2020-10-06
Make parsing publicly accessible

mirosval:
Partial fix for #726 
Reference #773 

I've added `parse` as a public method on the `GraphQLRequest` type. I've documented and exposed `Document`, `Spanning` and `ParseError` to allow working with the resulting type.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,7 @@ members = [
   "examples/warp_async",
   "examples/warp_subscriptions",
   "examples/actix_subscriptions",
+  "examples/simple",
   "integration_tests/juniper_tests",
   "integration_tests/async_await",
   "integration_tests/codegen_fail",
--- a/examples/simple/Cargo.toml
+++ b/examples/simple/Cargo.toml
@@ -1,0 +1,10 @@
+[package]
+name = "simple"
+version = "0.1.0"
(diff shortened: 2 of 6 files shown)


ccbrown:
I think we could kill two birds with one stone if instead of adding a `parse` method to `GraphQLRequest`, we just added a method to get the query. This would address part of https://github.com/graphql-rust/juniper/issues/750.

I this it's also work look into eliminating the schema / scalar template parameter requirement from parsing before making `Document` public so that we don't have to make a backwards-incompatible change later. If you'd like, I can probably take some time to investigate this further.

mirosval:
Thanks @ccbrown! I'm not sure I understand where you're suggesting to add the method. Are you suggesting that:
1. We move the parse method elsewhere (as a free fn) AND
2. Add a simple accessor method to get the query out of `GraphQLRequest`?

Regarding the `Scalar` in Document, how about I create another PR, address that and then we can come back to this one and revise it?

ccbrown:
> Are you suggesting that:
> 
> 1. We move the parse method elsewhere (as a free fn) AND
> 2. Add a simple accessor method to get the query out of GraphQLRequest?

Just 2. Then delete the parse method because we could just use the existing `parse_document_source` function with `GraphQLRequest`.

> Regarding the Scalar in Document, how about I create another PR, address that and then we can come back to this one and revise it?

Sounds good to me. 👍

PR #779, opened 2020-10-08
Default to generic ScalarValue in #[graphql_object] macro

### LINKED ISSUE #580: #[juniper::graphql_object] != #[derive(juniper::GraphQLObject)]

### LINKED ISSUE #769: Cannot derive GraphQLUnion with variants containing structs defined with graphql_object
Using latest master version of Juniper as of 2020/09/23.

**Describe the bug**
Trying to derive a union with variants containing structs defined using `#[juniper::graphql_object]` on their impls instead of deriving `juniper::GraphQLObject` on the structs fails.

Errors include the following:

1. trait `juniper::marker::GraphQLObjectType<__S>` is not implemented for `One`
2. trait `juniper::IsOutputType<__S>` is not implemented for `One`
4. expected enum `juniper::DefaultScalarValue`, found type parameter `__S`
5. expected an implementor of trait `juniper::GraphQLValueAsync<__S>`
6. `juniper::FieldError` cannot be formatted with the default formatter
7. expected an implementor of trait `juniper::GraphQLValueAsync<__S>`

This may be a mistake on my part, but I thought this would work based on explanations in the master book.

**To Reproduce**

```rust
#[derive(derive_more::From, juniper::GraphQLUnion)]                                                                                           
pub enum MyUnion {                                                                                                                            
    One(One),                                                                                                                                 
    Two(Two),                                                                                                                                 
}                                                                                                                                             
                                                                                                                              
pub struct One {                                                                                                                              
    a: String,                                                                                                                                
    b: i32,                                                                                                                                   
}                                                                                                                                             
                                                                                                                                              
#[juniper::graphql_object]                                                                                                                    
impl One {                                                                                                                                    
    fn a(&self) -> &str {                                                                                                                     
        &self.a                                                                                                                               
    }                                                                                                                                         
                                                                                                                                                
    fn b(&self) -> i32 {                                                                                                                      
        self.b                                                                                                                                
    }                                                                                                                                         
}                                                                                                                                             
                                                                                                                                        
pub struct Two {                                                                                                                              
    a: String,                                                                                                                                
    c: bool,                                                                                                                                  
}                                                                                                                                             
                                                                                                                                          
#[juniper::graphql_object]                                                                                                                    
impl Two {                                                                                                                                    
    fn a(&self) -> &str {                                                                                                                     
        &self.a                                                                                                                               
    }                                                                                                                                         
                                                                                                                                      
    fn c(&self) -> bool {                                                                                                                     
        self.c                                                                                                                                
    }                                                                                                                                         
}     
```


**Expected behavior**
Based on the examples from the master branch's book, I expected this to compile and work just like if I had defined the objects with `#[derive(juniper::GraphQLObject)]`.

tyranron:
Fixes #769   
Fixes #580

This PR makes `#[graphql_object]` macro to generate code which is generic over `S: ScalarValue` by default. Previously, `DefaultScalarValue` has been used.

## Checklist

- Tests updated
    - [x] integration tests
    - [x] codegen failure tests
    - [x] `juniper` crate tests
    - [x] `juniper_codegen` crate tests
    - [x] third-party integrations tests
    - [x] Book tests
- [x] Documentation in Book is updated
- [x] CHANGELOG entry is added


tyranron:
@LegNeato @ccbrown I've made it, but I'm unsure because of consequences it does. Please, review how examples were transformed. The main pain point goes from fallible fields, because `FieldError` is parametrized with `S: ScalarValue`. But if user uses custom type implementing `IntoFieldError<S>` then everythig works like a charm.

LegNeato (commented):
This is great work. I'm worried...it seems like this change bleeds into the developer experience heavily. Before, you basically had one place where you had to deal with scalar values (mixing the derive and the proc macro). Now, it shows up in a bunch of function signatures and turbofish calls by default.

ccbrown:
I think it should be possible to implement `IntoFieldError<S> for FieldError<DefaultScalarValue>`. `ScalarValue` requires `From` implementations for all of the `DefaultScalarValue` variants, so a `DefaultScalarValue` should be convertible to any other scalar value type. We currently have a reflexive implementation of `IntoFieldError` that I think would conflict, but if we got rid of `IntoFieldError` and just used `Into`, it should be possible.

Then users could always just return `FieldError` without the explicit parameter.

One of the weirdest things about this PR to me is that there's an increasing number of places where a variable named *Default*ScalarValue is no longer the default. That line in the quickstart that says "specify DefaultScalarValue if you *don't* want the default behavior" might be a bit of a head scratcher. Makes me wonder if a name like "StandardScalarValue" would be more appropriate.

tyranron:
@ccbrown @LegNeato yeah... you've confirmed my worries 🙃

> it may be possible to implement IntoFieldError<S> for FieldError<DefaultScalarValue>. ScalarValue requires From implementations for all of the DefaultScalarValue variants, so a DefaultScalarValue should be convertible to any other scalar value type.

What a nice idea! Thank you! I'll try to poke with it more.

tyranron:
@ccbrown yeah... it's very quickly hits into the specialization problem. But I've succeed to imitate the specialization in this situation with a [`TypeId` + `transmute` hack](https://github.com/graphql-rust/juniper/blob/c45104cc8327dac03f44363148281d35c396097c/juniper/src/value/mod.rs#L169-L173), which with a little help of optimizer should have no negligable performance impact (`TypeId` is a const info, so after monomorphization an optimizer should just cut off always irrelevant `if` branches).

Now I should re-refactor the changes in tests, examples, etc.

tyranron:
@ccbrown @LegNeato please look, now it's much more convenient. Fallible fields and subscription coerce nicely.

The only place left, where I cannot hide the type parameter is the  GraphQL interface, when the trait object is used. Though it's very rare case, and I may poke with it later.

ccbrown (approved):
Yeah this is definitely looking a lot nicer. 👍

LegNeato (approved):
Looks great!

Issue #183, opened 2018-05-16
Support where `Null` value came from


pepsighan:
I wanted to do PATCH operation in GraphQL but came to know that I had to do a workaround like #108 which in my honest opinion makes me write more code for little effect which is no fun.

In that issue, it was stated along with the workaround that :
> Lee Byron and friends (GraphQL designers and spec writers) recommend that you don't use null as a sentinel value for partial updates as most languages don't have the undefined/null distinction and because it limits implementors on how to treat values and objects/hashes.

I counter that with [The GraphQL Spec](http://facebook.github.io/graphql/October2016/#sec-Null-Value) which clearly dictates, though `undefined` is still not supported, an absence of `null` can be interpreted differently that explicit `null`.
> For example, these two field calls are similar, but are not identical:
>
> ```
>{
>  field(arg: null)
>  field
>}
>```
>The first has explictly provided null to the argument “arg”, while the second has implicitly not provided a value to the argument “arg”. 
> These two forms may be interpreted differently. For example, a mutation representing deleting a field vs not altering a field, respectively.

So, even though only `Null` value is supported, it would be true to the spec if we would be given the source of the said value, whether it was explicit or implicit.
Like: 
```
enum InputValue {
    Null(Null)
    ...
}

struct Null {
    explicit: bool
}
```

And may be a helper conversion to `Option<Option<T>>` where 
```
None => unspecified
Some(None) => null
Some(Some(value)) => value
```

mhallin:
Yes, it seems as if the GraphQL spec and authors changed their stance on `null` vs. `undefined` since we last looked at this.

I think the easiest change would be to add a tri-state enum like this:

```rust
enum GraphQLOption<T> {
    Absent,
    None,
    Some(T),
}
```

This would let you continue to use `Option<T>` where you don't care about null vs. absent, or this enum in the cases you do. It does require some infrastructure support though, the difference between those two cases are removed pretty early in processing today, as far as I can remember.

msdinit:
Hey,
Looking into trying to implement this feature, was wondering if anybody else is actively working on this, and if `GraphQLOption` is still the preferred way to handle this case?

ccbrown:
This is now supported in master via a `Nullable<T>` enum:

https://github.com/graphql-rust/juniper/blob/cd66bdb45063090164130eb970bec232008bdaa2/docs/book/content/advanced/implicit_and_explicit_null.md

PR #795, opened 2020-10-23
Add ability to distinguish between implicit and explicit null

ccbrown:
This adds the ability to distinguish between implicit and explicit null. This closes https://github.com/graphql-rust/juniper/issues/183.

The changes to Juniper to support this are pretty simple:

* `FromInputValue` now provides a `from_implicit_null` method. By default this just invokes `from_input_value(&InputValue::<S>::Null)` like Juniper did before, but types can provide their own implementation if they have a different representation for implicit nulls.
* `Arguments` no longer inserts `InputValue::Null`s into itself for missing arguments. This means `Arguments::get` will later return `None` for implicitly null arguments.
* Codegen has been modified to invoke `from_implicit_null` when arguments or fields are absent.

And finally, to make this functionality convenient for anyone to use without a lot of boilerplate, a `Nullable` type has been added:

```rust
pub enum Nullable<T> {
    ImplicitNull,
    ExplicitNull,
    Some(T),
}
```

This functions like `Option` except it has two empty variants, one for each type of null. It has all of the methods and traits you would expect from an `Option` type (with some exceptions as methods like `take` wouldn't have obvious behavior). It also has a few other methods that make sense such as `is_null`, `is_explicit_null`, `some`, etc.

### DIFF
--- a/juniper/src/ast.rs
+++ b/juniper/src/ast.rs
@@ -153,6 +153,13 @@ pub type Document<'a, S> = Vec<Definition<'a, S>>;
 pub trait FromInputValue<S = DefaultScalarValue>: Sized {
     /// Performs the conversion.
     fn from_input_value(v: &InputValue<S>) -> Option<Self>;
+
+    /// Performs the conversion from an absent value (e.g. to distinguish between implicit and
+    /// explicit null). The default implementation just uses `from_input_value` as if an explicit
+    /// null were provided. This conversion must not fail.
+    fn from_implicit_null() -> Self {
+        Self::from_input_value(&InputValue::<S>::Null).unwrap()
+    }
 }
 
 /// Losslessly clones a Rust data type into an InputValue.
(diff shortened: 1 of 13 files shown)


LegNeato:
Excellent, this will fix a lot of pain. Can you add docs to the book as this is a frequent request?

ccbrown:
@LegNeato added. :)

juniper/src/ast.rs
```diff
@@ -153,6 +153,13 @@ pub type Document<'a, S> = Vec<Definition<'a, S>>;
 pub trait FromInputValue<S = DefaultScalarValue>: Sized {
     /// Performs the conversion.
     fn from_input_value(v: &InputValue<S>) -> Option<Self>;
+
+    /// Performs the conversion from an absent value (e.g. to distinguish between implicit and
+    /// explicit null). The default implementation just uses `from_input_value` as if an explicit
+    /// null were provided. This conversion must not fail.
+    fn from_implicit_null() -> Self {
+        Self::from_input_value(&InputValue::<S>::Null).unwrap()
```
LegNeato, line 161: Small nit, better to use `expect` with a message so we can track down if this fails for some reason.

LegNeato (approved):
Thank you! Let's add this to the changelog as well.

PR #796, opened 2020-10-28
Implement `IntoFieldError` for `Infallible`

davidpdrsn:
Makes it possible to use `Result<T, Infallible>` as your return type from resolvers, which can be handy sometimes.


LegNeato:
Thank you! Can we get a quick test to make sure we don't regress this?

LegNeato:
(and adding it to the changelog)

davidpdrsn:
Of couse 😊

LegNeato (approved):
Thanks again and welcome back :-)

PR #803, opened 2020-11-07
Add juniper_warp/examples/Cargo.toml, so example server can be run easily

noamraph:
I added a Cargo.toml file to juniper_warp/examples, so I could easily compile and run the example server.


tyranron:
@noamraph you don't need a separate `Cargo.toml` for that, just run:
```bash
cargo run -p juniper_warp --example warp_server
```

You can read [about Cargo's CLI more here](https://doc.rust-lang.org/cargo/commands/cargo-run.html#package-selection).

Issue #737, opened 2020-08-13
Do a crates.io release with Websockets / Subscriptions support


leebenson:
@LegNeato, thanks for the excellent library. We're currently experimenting with it for Vector (https://vector.dev/).

Do you have an expectation for when a release may be cut with 'official' subscriptions / Websockets support, please? Our use-case is primarily streaming data, so this would be perfect.

I saw [your warning](https://github.com/graphql-rust/juniper/issues/669#issuecomment-640105347) about not using master in production, so I was curious if you had an ETA for the next release?

Thanks!

LegNeato:
Awesome, love vector and suggested work should use it 😄 .

Waiting for async interfaces (https://github.com/graphql-rust/juniper/pull/682) to firm up land and then intend to cut a new release.

If @tyranron can't get back to it soon I will likely pick it up.

tyranron:
@LegNeato I haven't recently budget to work on interfaces, but I hope during next month I'll finish them.

There is also some work with refactoring existing macros to have them all consistent and play well with `ScalarValue` defaults, but I think this can be postponed to next major release.

leebenson:
Are either of you available for commercial support?

We're keen on implementing Juniper in Vector ASAP. We're building an awesome new visualization UI for Vector, with Juniper as the GraphQL provider between a deployed instance, and the browser.

Websockets and subscription support is what we're after right now. Beyond that, there may be other issues or bug fixes that may make sense to prioritize.

I appreciate your time on the library and know it's hard to juggle between day jobs. If you have any available bandwidth to get the next release over the line, and tackle other issues ad hoc, we'd love to compensate your time.

Feel free to reach out to me by e-mail if that's more appropriate -- lee@timber.io.

Thanks!

tyranron:
@leebenson just because I think Vector is such an amazing piece of software, I've rescheduled my priorities for you and will finish #682 as soon as I'm capable to.

Regarding the commercial support, thank you very much for this offer, however, I can't accept it due to several reasons (including personal opinions about collaboration in open source).  
I'll help as much as I'll able to, but for no money.

leebenson:
@tyranron, that's an incredible gesture - thanks so much for your support.

I know how difficult it can be juggling multiple priorities. The offer for commercial support remains if it helps remove external pressure, but I certainly respect your position on open source. Thanks so much!

kunjee17:
#682 is in master. It's time to update the world. 😎

kunjee17:
[...]

PR #806, opened 2020-11-13
Fix attributes naming on methods and arguments for interfaces and unions

tyranron:
This PR corrects the naming of attributes in `#[graphql_interface]` and `#[graphql_union]` macros, placed on fields and field arguments to apply well along with the `#[graphql_object]` macro.

See https://github.com/graphql-rust/juniper/issues/737#issuecomment-727015731:
> ```rust
> #[graphql_object]
> impl Object {
>     #[graphql(name = "fancy")]
>     fn fancy_field() -> i32 { 0 }
> }
> 
> #[graphql_interface]
> trait Interface {
>     #[graphql_interface(name = "fancy")]  // <-- should be renamed as `#[graphql]`
>     fn fancy_field() -> i32 { 0 }
> }
> ```

Also, revives some tests in `juniper` crate, previously commented and skipped.

### DIFF
--- a/docs/book/content/types/interfaces.md
+++ b/docs/book/content/types/interfaces.md
@@ -167,7 +167,7 @@ use juniper::{graphql_interface, GraphQLObject};
 trait Character {
     fn id(&self) -> &str;
 
-    #[graphql_interface(ignore)] // or `#[graphql_interface(skip)]`, your choice
+    #[graphql(ignore)] // or `#[graphql(skip)]`, your choice
     fn ignored(&self) -> u32 { 0 }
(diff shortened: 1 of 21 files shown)


LegNeato (approved):
Great! Perhaps we should put a pointer in the changelog to the Pets changes in this PR to show folks how to go from the old `graphql_interface!` macro to the new proc macro?

tyranron:
@LegNeato done in https://github.com/graphql-rust/juniper/commit/8783496c68d5195be2c16802485debe6d817061e

PR #807, opened 2020-11-15
Allow different Scalar for GraphQLScalarValue

jmpunkt:
##### Motivation

Currently it is not possible to use scalar values with custom scalars. The following code fails.

```rust
#[derive(juniper::GraphQLScalarValue)]
pub struct LargeId(i64);
```

In order to allow custom scalars, scalars values must allow to set a different scalar value. Currently the implementation only uses generics. The following snippet works with this PR. Notice that `MyScalarValue` is a scalar implementation including the `i64` data type.

```rust
#[derive(juniper::GraphQLScalarValue)]
#[graphql(Scalar = MyScalarValue)]
pub struct LargeId(i64);
```


LegNeato (approved):
Looks great, thank you so much for the high quality PR with tests! 🍻

Issue #693, opened 2020-07-08
 ゛


toxeus:

LegNeato:
Great idea!

toxeus:
@zaksabeast do you plan to continue work on this PR?

toxeus:
I was able to resolve this by migrating to [async-graphql](https://github.com/async-graphql/async-graphql).

tyranron:
Leaving this still open, as something that could be improved.

PR #811, opened 2020-11-21
Add descriptive value validation errors

zaksabeast:
Fixes: #693 

This pull request adds descriptive errors when validating argument and default values.

For example, an input object missing a field would show `Error for argument "arg": "ExampleInputObject" is missing fields: "b"`.

If there's a different approach you'd like to take for this, I'd be happy to implement that as well.

### DIFF
--- a/juniper/src/executor_tests/enums.rs
+++ b/juniper/src/executor_tests/enums.rs
@@ -99,7 +99,7 @@ async fn does_not_accept_string_literals() {
     assert_eq!(
         error,
         ValidationError(vec![RuleError::new(
-            r#"Invalid value for argument "color", expected type "Color!""#,
+            r#"Error for argument "color": Invalid value ""RED"" for enum "Color""#,
             &[SourcePosition::new(18, 0, 18)],
         )])
     );
--- a/juniper/src/executor_tests/variables.rs
+++ b/juniper/src/executor_tests/variables.rs
@@ -1038,7 +1038,7 @@ async fn does_not_allow_missing_required_field() {
     assert_eq!(
         error,
         ValidationError(vec![RuleError::new(
-            r#"Invalid value for argument "arg", expected type "ExampleInputObject!""#,
+            r#"Error for argument "arg": "ExampleInputObject" is missing fields: "b""#,
             &[SourcePosition::new(20, 0, 20)],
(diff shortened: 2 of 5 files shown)


LegNeato:
Thanks for the PR! Are there perf implications of this change?

zaksabeast:
Sorry for the late response, and thank you for updating this branch!

The only performance implications I'm aware of would be for errors that are nested fairly deep within a graph.  For example, an error on a GraphQL object nested 10 levels deep would have the error bubble up to each of the 10 levels, and create a new error string each time to help pinpoint the error's location.

If this is undesirable (since it could create larger errors for larger graphs), an alternate solution could be to have an enum for the different types of validations, and not build upon errors for objects or lists.  This would ensure everything else would bubble up to an object or list, so the relative location of the error would be known, but the error wouldn't grow in size if it was deeply nested.

The second solution would be more performant since it wouldn't need to create as many strings on larger graphs, but at the cost of less information to locate an error.

Are there any performance implications you're wondering about specifically?

toxeus:
@zaksabeast do you plan to continue work on this PR?

zaksabeast:
@toxeus, sorry for the late response.  I'm up for continuing this PR, although I'm not quite sure what changes are needed.

haarts:
Perhaps @tyranron you can shine your light on this?

tyranron:
@haarts definitely not the first thing on my agenda, but I'll try to include this into new `0.16` major release.

toxeus:
@haarts @tyranron I think this would be a great addition for the next release. It feels inapt that `GraphQLScalar::from_input_value()` returns an `Option` instead of a `Result` which could carry detailed information about de-serialization errors that could (or rather should) be passed back to the client. Currently, I often have to reach out to the backend logs to find out why a certain GraphQL request failed. This doesn't feel right, especially if we consider that `juniper` can also be used to implement services for third parties that probably will not have access to backend logs.

tyranron:
> It feels inapt that `GraphQLScalar::from_input_value()` returns an `Option` instead of a `Result` which could carry detailed information about de-serialization errors that could (or rather should) be passed back to the client.

This was done in #987.

tyranron (approved):

PR #812, opened 2020-11-24
Allow raw identifier for field arguments

### LINKED ISSUE #786: Field argument descriptions don't get applied if using raw identifier in `arguments`
This is quite an obscure bug that I doubt many will run into, but I did hit while building juniper-from-schema.

**Describe the bug**
If you define an object like this:

```rust
#[graphql_object]
impl Query {
    #[graphql(
        arguments(
            a(description = "foo"),
            b(description = "bar"),
        )
    )]
    fn add(a: i32, b: i32) -> i32 {
        a + b
    }
}
```

Then the docs on the arguments show up just fine in GraphiQL.

However if you use raw identifiers

```rust
#[graphql_object]
impl Query {
    #[graphql(
        arguments(
            r#a(description = "foo"),
            r#b(description = "bar"),
        )
    )]
    fn add(r#a: i32, r#b: i32) -> i32 {
        a + b
    }
}
```

Then they don't show up

**To Reproduce**
Example warp app with GraphiQL that I used for debugging this

```rust
use juniper::*;
use warp::Filter;

type Schema = RootNode<'static, Query, EmptyMutation<()>, EmptySubscription<()>>;

fn schema() -> Schema {
    Schema::new(
        Query,
        EmptyMutation::<()>::new(),
        EmptySubscription::<()>::new(),
    )
}

struct Query;

#[graphql_object]
impl Query {
    #[graphql(
        arguments(
            r#a(description = "foo"),
            r#b(description = "bar"),
        )
    )]
    fn add(r#a: i32, r#b: i32) -> i32 {
        a + b
    }
}

#[tokio::main]
async fn main() {
    let state = warp::any().map(move || ());

    let graphql_filter = juniper_warp::make_graphql_filter(schema(), state.boxed());

    let graphiql_filter = warp::get()
        .and(warp::path("graphiql"))
        .and(juniper_warp::graphiql_filter("/graphql", None));

    warp::serve(graphiql_filter.or(warp::path("graphql").and(graphql_filter)))
        .run(([127, 0, 0, 1], 8080))
        .await
}
```

zaksabeast:
Fixes #786.

This pull request allows raw identifiers to be used for field arguments, and updates `multi_args_descr_trailing_comma` to include a trailing comma.

### DIFF
--- a/integration_tests/juniper_tests/src/codegen/impl_object.rs
+++ b/integration_tests/juniper_tests/src/codegen/impl_object.rs
@@ -95,8 +95,10 @@ mod fallible {
     }
 }
 
-mod raw_arguments {
-    use juniper::graphql_object;
+mod raw_argument {
+    use juniper::{
+        graphql_object, graphql_value, EmptyMutation, EmptySubscription, RootNode, Variables,
+    };
 
     struct Obj;
 
@@ -106,5 +108,32 @@ mod raw_arguments {
         fn test(&self, arg: String) -> String {
             arg
         }
+    }
+
+    #[tokio::test]
+    async fn named_correctly() {
+        let doc = r#"{
+            __type(name: "Obj") {
+                fields {
+                    args {
+                        name
+                    }
+                }
+            }
+        }"#;
+
+        let schema = RootNode::new(
+            Obj,
+            EmptyMutation::<()>::new(),
+            EmptySubscription::<()>::new(),
+        );
+
+        assert_eq!(
+            juniper::execute(&doc, None, &schema, &Variables::new(), &()).await,
+            Ok((
+                graphql_value!({"__type": {"fields": [{"args": [{"name": "arg"}]}]}}),
+                vec![],
+            )),
+        );
     }
 }
--- a/integration_tests/juniper_tests/src/codegen/impl_object.rs
+++ b/integration_tests/juniper_tests/src/codegen/impl_object.rs
@@ -94,3 +94,17 @@ mod fallible {
         }
     }
 }
+
+mod raw_arguments {
+    use juniper::graphql_object;
+
+    struct Obj;
+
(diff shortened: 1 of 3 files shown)


tyranron (changes requested):
@zaksabeast thank you!

Would you be so kind to re-check that for GraphQL interfaces and add appropriate tests to `integration_tests/juniper_tests/` too?

zaksabeast:
@tyranron Sure thing!

I'd be happy to do the re-check and add integration tests.

Could you please elaborate on the ask to re-check GraphQL interfaces?  I'm not super familiar with Juniper and might be misunderstanding the ask, so correct me if I'm wrong, but it looks like `arguments` can't currently be used with methods of a trait using the `graphql_interface` proc macro.

Would adding a test to `integration_tests/juniper_tests/src/codegen/derive_object_with_raw_idents.rs` make sense, or would a new file be preferred?

tyranron:
@zaksabeast 

> it looks like `arguments` can't currently be used with methods of a trait using the `graphql_interface` proc macro.

Hmmm, right. I didn't figure it out. So, `#[graphql_interface]` doesn't have such problem at all.

> Would adding a test to `integration_tests/juniper_tests/src/codegen/derive_object_with_raw_idents.rs` make sense, or would a new file be preferred?

It would be better to add `mod raw_arguments` to the `integration_tests/juniper_tests/src/codegen/impl_object.rs`, like [`fallible` does](https://github.com/graphql-rust/juniper/blob/a4871887bb6b30029bd4671af716c0649a0cc60c/integration_tests/juniper_tests/src/codegen/impl_object.rs#L85-L96).

Thanks!

zaksabeast:
Perfect, thank you for the help!

tyranron (approved):
@zaksabeast thanks! 🍻

PR #822, opened 2020-12-10
Mark `Arc` with `IsInputType` & `IsOutputType`.

obmarg:
We've just updated a service we work with to the latest release of
juniper (it's been on a fairly old master commit for some time).  In
this service we have some fields that are contained within `Arc`s, which
I added support for in #479.

Since then it seems the new marker traits of `IsInputType` &
`IsOutputType` have been added, but they do not support `Arc`, leading
to:

```
error[E0277]: the trait bound `Arc<menu::Menu>: IsOutputType<__S>` is not satisfied
  --> src/graphql.rs:36:1
   |
36 | #[juniper::graphql_object(Context=Context)]
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `IsOutputType<__S>` is not implemented for `Arc<menu::Menu>`
   |
   = note: required by `juniper::marker::IsOutputType::mark`
   = note: this error originates in an attribute macro (in Nightly builds, run with -Z macro-backtrace for more info)
```

This commit adds impls for `Arc`, mirroring the impls for `Box`.

I'm not sure if there's a good place to add a test for this
functionality so that regressions can be avoided in the future?

### DIFF
--- a/integration_tests/juniper_tests/src/arc_fields.rs
+++ b/integration_tests/juniper_tests/src/arc_fields.rs
@@ -1,0 +1,15 @@
+use std::sync::Arc;
+
+struct Query;
+
+#[juniper::graphql_object]
+impl Query {
+    fn ping() -> Arc<bool> {
+        Arc::new(false)
+    }
+}
+
+#[derive(juniper::GraphQLInputObject)]
+struct Ping {
+    expect_result: Arc<bool>,
+}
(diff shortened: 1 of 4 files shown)


LegNeato:
Ah, sorry about that! I'll cut a release with this when it lands. Do not feel strongly about where the test goes, anywhere it makes sense to you.

obmarg:
@LegNeato no problem, probably our fault for staying on an old commit for so long (and I really should have added a test in my initial PR).

Anyway, pushed up a test & a fix for Arc input values as well.  Let me know if there's anything else you think I should do - excited to be pulling juniper from crates.io

LegNeato:
Just released a patch release with this, thank you for the PR!

PR #826, opened 2020-12-13
Add EmptyQuery helper

Guara92:


LegNeato:
I think we are going to close this out because empty queries are not supported by the spec and I don't want to assume a fake one. Would totally take a pull request explaining this either in the module with the other empty types or in the book. Thank you so much for the pr!

PR #838, opened 2020-12-28
Add missing use declaration in Quickstart

bonnetn:
Seems like the example code in the quickstart guide is missing an use declaration. 

The code fails to compile: `error: cannot find macro `graphql_value` in this scope` with `rustc 1.48.0 (7eac88abb 2020-11-16)`


LegNeato:
Thank you!

PR #839, opened 2020-12-28
Make the actix example self contained

### LINKED ISSUE #804: Quickstart in book does not provide a working example
It is  painful to get started with Juniper due to the lack of a complete working example. The quickstart should provide a simple working example that compiles and runs out of the box. 

It would be really helpful to pick a server framework and provide a working example in the book, instead of referring to external examples. I would point out that there are a number of examples spread throughout the codebase, and none of them really address the beginner.

A couple of other observations:
 - mutations are not covered in enough depth. 
 - the lack of a rust client is conspicuously absent

EdorianDark:
A smaller example is esier to understand.
Fixes #804


EdorianDark:
If this is a good direction I would also adapt the others example, to make them uniform.

Issue #425, opened 2019-09-05
Add ?Send futures support in async-await branch


nWacky:
**Is your feature request related to a problem? Please describe.**
Sometimes it's necessary to run not `Send` future in juniper handler (for example to get data from database). However, when trying to run such future I get the following error:
```
error[E0277]: `dyn core::future::future::Future<Output = std::result::Result<User, Error>>` cannot be sent between threads safely
  --> mod.rs:15:1
   |
15 | #[juniper::object(name = "Mutations", Context = Context)]
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `dyn core::future::future::Future<Output = std::result::Result<User, Error>>` cannot be sent between threads safely    

```

**Describe the solution you'd like**
It would be nice to be able to use an option to use `!Send` futures in juniper, ideally with `?Send` option on `#[juniper::object]` macro (like in [async_trait](https://docs.rs/async-trait/0.1.11/async_trait/#non-threadsafe-futures) )

I have [tried removing Send bound from everywhere](https://github.com/instrumentisto/juniper/tree/async-await) and it works fine, but then it's not possible to use futures from juniper as `Send` futures (for example, I cannot call ` req.execute_async(schema, context).boxed().await`)

I don't think that it would be possible to have such option on `#[juniper::object]` macro because `Send` marker is defined in trait, not in it's implementantion. 
I think it could be done with a cargo.toml feature, but then it won't be possible to use both `Send` and `!Send` futures in one project

**Describe alternatives you've considered**
I have considered changing database connection function, but it decreases performance

theduke:
I don't think we'll be able to fit `!Send` futures into the design, even with a feature. Things are already a bit messy with `GraphQLTypeAsync` layered on top of `GraphQLType` (this might change a a bit in the future).

Your use case seems to be dealing with a synchronous database connection? 
You can't use those anyway in a async environment, since you are not supposed to block a executor thread. 

Longer synchronous operations need to be delegated to a thread pool.

Below is a super simple (suboptimal) code snippet as an example for such a threadpool.
I'm sure a good crate with a better implementation will emerge pretty soon.

```rust
#[derive(Clone)]
pub struct SyncThreadPool {
    pool: Arc<Mutex<threadpool::ThreadPool>>,
}

impl SyncThreadPool {
    pub fn new(threads: usize) -> Self {
[...]

Issue #329, opened 2019-03-05
Add GraphQLContext derive


LegNeato:
**Is your feature request related to a problem? Please describe.**
Contexts have some boilerplate, you have to implement `juniper::Context` which is usually an empty marker trait. The error message when you don't isn't obvious, see  https://github.com/graphql-rust/juniper/issues/327. 

**Describe the solution you'd like**
It might be good to have a `#[derive(GraphQLContext)` that does the boilerplate for you. It won't help the error message, but perhaps when people are copying and pasting or reading docs they will be less likely to miss it.

theduke:
Actually, I think we could just deprecate the `Context` trait completely and remove the trait bound.

There is no real reason for it to be there.

theduke:
I guess the only reason for it's existence would be that you can't use a type as context in the macros that isn't intended to be used as a context. 

But that would produce compile time errors anyway once you actually construct the schema.

LegNeato:
Do we want to reserve the trait for future extensibility? If we remove the requirement and later need it for implementation specifics, it would be a breaking change.

davidpdrsn:
@LegNeato what sort of future extensibility are you thinking of? I imagine adding methods to the trait would also be breaking change.

I'm also in favour of deprecating it.

theduke:
I don't see a reason to ever add methods on context.
More state keeping or functionality would be added to the `Executor`.

There is a consideration with futures support though: we will probably need a `Send + Sync + Clone` bound on context!
With that in mind, we may want to add those constraints onto the `Context` trait instead of litterling them around the code everywhere.

davidpdrsn:
I'm not that familiar with futures, but why `Clone`?

The app I'm working on has a context like this:

```rust
pub struct Context {
    db: DbCon,
    // more things...
}
```

Where `DbCon` is

```rust
pub struct DbCon(pub PooledConnection<ConnectionManager<PgConnection>>);
```

`PooledConnection` comes from [`r2d2`](https://docs.rs/r2d2/0.8.4/r2d2/struct.PooledConnection.html) and is _not_ `Clone`. So I can't easily make this context `Clone`.

I could instead store the connection pool itself in the context. The pool is `Clone` because it just [wraps and `Arc<SharedPool>`](https://docs.rs/r2d2/0.8.4/src/r2d2/lib.rs.html#316-318). That feels like it would create a lot of churn on the pool though. If each db query would have to first get a new connection.

[...]

Issue #828, opened 2020-12-14
session with actix-web


nebnes:
Hi, 
I am trying to set a session user in a login mutation. 
But I don't find how access by the context to the session.. 
Is it possible ? 
Thanks

tyranron:
@nebnes you make a `HttpRequest` your `Context`. Or roll out a custom context containg your session.

nebnes:
I try to pass the session in the context but I have this error : 
``` 
`std::rc::Rc<std::cell::RefCell<actix_session::SessionInner>>` cannot be shared between threads safely
```
Code : 
```
 let context = &Context {
        app_data: data.into_inner().clone(),
        pool: pool.get_ref().to_owned(),
        session : session
    };
   

    let res = request.execute(&schema, context).await;
```

The context :  
```
pub struct Context {
    pub app_data: AppState,
    pub pool: PgPool,
    pub session :Session
}

impl juniper::Context for Context {}
```

tyranron:
@nebnes yup [`SendWrapper`](https://docs.rs/send_wrapper/0.5.0/send_wrapper/) will do here, as long a you're using `actix-web` (it doesn't send `Future`s between threads).

See the [example I use](https://github.com/ALLATRA-IT/ephyr/blob/3e48d46f4ccc29d1acffa7b1011d652c45ee2c58/components/restreamer/src/api/graphql/mod.rs#L18-L52).

nebnes:
thanks @tyranron

PR #840, opened 2021-01-01
Support !Sync (and maybe !Send) Context

AzureMarker:
When upgrading juniper from 0.9 to 0.15, I found that context types now require `Send + Sync`. Previously, before juniper was async, everything was synchronous and the context could contain `!Sync` types such as database connections (ex. SQLite). This PR experiments with removing the `Sync` requirement (and `Send`, since the context is only ever referenced not owned).

There are some roadblocks which I'm not sure can be resolved at this time. The juniper futures now are `!Send` since they hold a reference to the `!Sync` context. This is fine in the core juniper library and juniper_actix, but web services based on tokio/hyper have issues with spawning `!Send` futures. Tokio supports spawning `!Send` futures, but only when using a `LocalSet`, which is only really feasible when using the single-threaded runtime. There is a (recently closed, but in-discussion) issue which addresses spawning `!Send` futures on the multi-threaded runtime: https://github.com/tokio-rs/tokio/issues/2545. Due to this issue, Hyper only supports `!Send` futures when using the single-threaded runtime: https://github.com/hyperium/hyper/blob/8861f9a7867216c81ea14ac6224c11a1303e7761/examples/single_threaded.rs. Warp does not support `!Send` yet: https://github.com/seanmonstar/warp/issues/528. It's not clear how Rocket 0.5 will handle `!Send`, but since it's based on hyper it won't be any better than what they provide.

Theoretically the future that runs the graphql resolver could be `Send` because the `!Sync` context is owned by the future and passed into juniper by reference. However, because GATs (Generic Abstract Types) are not implemented, traits with async methods must box their futures. This causes juniper's future to hold `LocalBoxFuture`s, which are not `Send`. Due to this, the overall future is not `Send`... If GATs were implemented, it may be possible to avoid the tokio `!Send` futures issue by convincing Rust that the overall juniper resolve future is `Send` since the context is owned by the calling code (the web handler which calls juniper) which is part of the same future object.

The main changes include:
* Removing `Send` and `Sync` bounds on context types.
* Replacing `BoxFuture` with `LocalBoxFuture` (similar for streams), due to the removal of `Context: Sync`.
* Updating `juniper_hyper` examples and tests to use a single-threaded runtime due to the above issues.
* Updating `juniper_warp` in an attempt to support `!Send` futures, but this fails at runtime due to the lack of support in warp for `!Send` futures (see issue link above).
* Move the config out of `ExecutionParams` so that the context is not captured in the error type, which would require the context to be `Send`. This is safe because the error object only refers to data in the request and schema, not the context (see the lifetimes on `juniper::execute`).

I don't really expect this PR to be merged any time soon, if at all, but it's worth bringing up the issue and showing how the core juniper library can support `!Sync` (and even `!Send`) context. The issues lie in the web server integration libraries, mainly due to the lack of GATs and issues in tokio with spawning `!Send` futures.

Related juniper issues:
* https://github.com/graphql-rust/juniper/issues/329#issuecomment-493672618
* https://github.com/graphql-rust/juniper/issues/425

### DIFF
--- a/examples/warp_subscriptions/src/main.rs
+++ b/examples/warp_subscriptions/src/main.rs
@@ -2,7 +2,7 @@
 
 use std::{env, pin::Pin, sync::Arc, time::Duration};
 
-use futures::{FutureExt as _, Stream};
+use futures::{FutureExt as _, Stream, TryFutureExt};
 use juniper::{
     graphql_object, graphql_subscription, DefaultScalarValue, EmptyMutation, FieldError,
     GraphQLEnum, RootNode,
@@ -162,14 +162,17 @@ async fn main() {
         .and(warp::ws())
         .map(move |ws: warp::ws::Ws| {
             let root_node = root_node.clone();
-            ws.on_upgrade(move |websocket| async move {
-                serve_graphql_ws(websocket, root_node, ConnectionConfig::new(Context {}))
-                    .map(|r| {
-                        if let Err(e) = r {
-                            println!("Websocket error: {}", e);
-                        }
-                    })
-                    .await
+            ws.on_upgrade(move |websocket| {
+                tokio::task::spawn_local(async move {
+                    serve_graphql_ws(websocket, root_node, ConnectionConfig::new(Context {}))
+                        .map(|r| {
+                            if let Err(e) = r {
+                                println!("Websocket error: {}", e);
+                            }
+                        })
+                        .await
+                })
+                .unwrap_or_else(|_| panic!("Failed to join handler"))
             })
         }))
     .map(|reply| {
--- a/juniper/src/executor/mod.rs
+++ b/juniper/src/executor/mod.rs
@@ -238,7 +238,7 @@ pub type ExecutionResult<S = DefaultScalarValue> = Result<Value<S>, FieldError<S
 
 /// Boxed `Stream` yielding `Result<Value<S>, ExecutionError<S>>`
 pub type ValuesStream<'a, S = DefaultScalarValue> =
-    std::pin::Pin<Box<dyn Stream<Item = Result<Value<S>, ExecutionError<S>>> + Send + 'a>>;
+    std::pin::Pin<Box<dyn Stream<Item = Result<Value<S>, ExecutionError<S>>> + 'a>>;
 
 /// The map of variables used for substitution during query execution
 pub type Variables<S = DefaultScalarValue> = HashMap<String, InputValue<S>>;
@@ -412,7 +412,6 @@ where
         'a: 'res,
         T: GraphQLSubscriptionValue<S, Context = CtxT> + ?Sized,
         T::TypeInfo: Sync,
-        CtxT: Sync,
         S: Send + Sync,
     {
         self.subscribe(info, value).await.unwrap_or_else(|e| {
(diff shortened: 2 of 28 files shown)


tyranron:
@Mcat12 so this PR makes `juniper` usable only with single-threaded runtimes, right? If so, I'm very unsure this chage worth such an impact. Maybe it's better to use [`send_wrapper`](https://docs.rs/send_wrapper) where you don't care about synchronization as where stated here: https://github.com/graphql-rust/juniper/issues/828#issuecomment-745486566?

Mcat12:
> @Mcat12 so this PR makes `juniper` usable only with single-threaded runtimes, right?

Not quite. I noted some of the current roadblocks in detail in the description, but basically juniper should be able to support `!Sync` context while still running on a multi-threaded runtime. My goal is to have the future returned by `juniper::execute` be `Send` if the context is `Sync` (keeping the current behavior), and be `!Send` only if the context is `!Sync`.

Juniper's core machinery is synchronous, operating within one future chain. The asynchrony comes from the user's resolver code, which can be async functions. This synchronous behavior supports having `!Sync` context, but proving that to the compiler right now is difficult.

The PR description notes some reasons why that is not the case currently, most notably that async functions cannot appear in traits due to a lack of GATs (Generic Associated Types). See for example [`GraphQLValueAsync::resolve_async`]. Another issue is that it is hard to spawn `!Send` futures in the multi-threaded tokio runtime. I've opened a PR to add a [`spawn_pinned`] API which would fix this shortcoming, but this limitation is why the hyper and warp juniper crates have issues in this PR. These roadblocks are being worked on, so eventually this idea should work out.

If you would like, I can mark this PR as a draft since it is unlikely to be merge-able in the short term.

[`GraphQLValueAsync::resolve_async`]: https://github.com/graphql-rust/juniper/blob/2c15ea798c1bd7df5bede24495fba541c1c84b2d/juniper/src/types/async_await.rs#L96
[`spawn_pinned`]: https://github.com/tokio-rs/tokio/pull/3370

tyranron:
@Mcat12 the biggest blocker is GAT, of course. The fact that `juniper` cannot be `Send`/`Sync` transparent is due to trait objects usage in traits. I've played with this multiple times last year and haven't figured out how it can be reasonable managed without GATs.

Things like whether runtime supports `!Send` futures or not are not quite `juniper` related, as it doesn't use runtimes directly. So, the best option is to be `Send`/`Sync` transparent.

> If you would like, I can mark this PR as a draft since it is unlikely to be merge-able in the short term.

Yeah, that would be nice.

> Edit: Regarding send_wrapper, that only works for actix (and other single-threaded runtimes). It doesn't fix the case of hyper and warp.

And it shouldn't, as for me. Running `!Send`/`!Sync` contexts on multi-threaded runtimes with work stealing should be prevented by compiler.

Mcat12:
> > Edit: Regarding send_wrapper, that only works for actix (and other single-threaded runtimes). It doesn't fix the case of hyper and warp.
> 
> And it shouldn't, as for me. Running `!Send`/`!Sync` contexts on multi-threaded runtimes with work stealing should be prevented by compiler.

Although it's somewhat unrelated to the core juniper library, this is where `spawn_pinned` in tokio comes in. This API would allow libraries like warp (which do not support `!Send` futures) to support `!Sync` context, making all of the supported web servers able to run with `!Sync` context.

AzureMarker:
Closing this PR since any future progress will require starting from scratch.

PR #847, opened 2021-01-12
Add missing marker trait impls for container types

### LINKED ISSUE #845: Union with Boxed enum variants
Hi, 

Upgrading from 0.14 to 0.15 is giving me some problems with unions that have some boxed enum variant, i.e.:
```
#[derive(GraphQLUnion)]
pub enum GqlResult {
    Ok(Box<Data>),
    Err(Error),
}
```

the docs specifies that container types are supported but the compiler complains about a missing impl of the trait `GraphQLObjectType`  for Box, I'm doing something wrong?

tyranron:
Fixes #845

### DIFF
--- a/integration_tests/juniper_tests/src/codegen/union_derive.rs
+++ b/integration_tests/juniper_tests/src/codegen/union_derive.rs
@@ -1476,8 +1476,6 @@ mod full_featured_struct {
 mod issue_845 {
     use std::sync::Arc;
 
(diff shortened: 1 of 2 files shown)


LegNeato (changes requested):
Seems to be failing CI

toxeus:
The CI is fixed in https://github.com/graphql-rust/juniper/pull/848

PR #848, opened 2021-01-15
juniper_rocket_async: fix tokio dependency

toxeus:
rocket async has recently bumped its tokio dependency
to 1.0.1 which breaks juniper_rocket_async. This can
be trivially fixed by using rocket's tokio re-export.


tyranron (approved):
Thanks! 🍻

Issue #853, opened 2021-01-22
Playground Broken in master


dt665m:
**Describe the bug**
@LegNeato Just a heads up, as of 2e24b55e99f7f7c38e70a640633f5b6479adeaf5, the graphql playground has borked because graphql-playground(js) 1.7.27 seems build into a different structure.  It no longer has the /build/static folder, so jsdelivr fails.

**To Reproduce**
- checkout master
- load the playground
- see the browser console.log where the middleware.js file can't be found from jsdelivr

**Expected behavior**
the playground works

dt665m:
seems like graphql-playground README recommends a static html file that doesn't specify the version , [here](https://github.com/graphql/graphql-playground/blob/master/packages/graphql-playground-html/withAnimation.html). Do we need to specify the version for a specific reason?  I just tried accessing the middleware.js file without the version from jsdelivr and it works.

dt665m:
PR submitted #854

dt665m:
#856 merged, closing

PR #854, opened 2021-01-22
remove jsdelivr versioning of graphql-playground

dt665m:
This is a fix for #853

### DIFF
--- a/juniper/src/http/playground.rs
+++ b/juniper/src/http/playground.rs
@@ -23,7 +23,7 @@ pub fn playground_source(
   <title>GraphQL Playground</title>
   <link rel="stylesheet" href="//cdn.jsdelivr.net/npm/graphql-playground-react/build/static/css/index.css" />
   <link rel="shortcut icon" href="//cdn.jsdelivr.net/npm/graphql-playground-react/build/favicon.png" />
-  <script src="//cdn.jsdelivr.net/npm/graphql-playground-react@1.7.27/build/static/js/middleware.js"></script>
+  <script src="//cdn.jsdelivr.net/npm/graphql-playground-react/build/static/js/middleware.js"></script>
 
 </head>
 


juniper/src/http/playground.rs
```diff
@@ -23,7 +23,7 @@ pub fn playground_source(
   <title>GraphQL Playground</title>
   <link rel="stylesheet" href="//cdn.jsdelivr.net/npm/graphql-playground-react/build/static/css/index.css" />
   <link rel="shortcut icon" href="//cdn.jsdelivr.net/npm/graphql-playground-react/build/favicon.png" />
-  <script src="//cdn.jsdelivr.net/npm/graphql-playground-react@1.7.27/build/static/js/middleware.js"></script>
+  <script src="//cdn.jsdelivr.net/npm/graphql-playground-react/build/static/js/middleware.js"></script>
```
tyranron, line 26: I don't think that removing the version at all is a good way to go, as it's the same as have `tokio = "*"` in `Cargo.toml`. You will never now when the things blew up on breaking change.

But!

We can just follow semver and specify `@1` version. So we catch up with recent updates but don't hit major version breakage.

juniper/src/http/playground.rs
```diff
@@ -23,7 +23,7 @@ pub fn playground_source(
   <title>GraphQL Playground</title>
   <link rel="stylesheet" href="//cdn.jsdelivr.net/npm/graphql-playground-react/build/static/css/index.css" />
   <link rel="shortcut icon" href="//cdn.jsdelivr.net/npm/graphql-playground-react/build/favicon.png" />
```
tyranron, line 25: And it seems that here and in the line above we should specify `@1` version too.

tyranron (changes requested):
@dt665m thanks for the contribution!

Zireael13:
Sorry didn't see this PR before I created #856. As I noted there, just using the latest version (currently 1.7.27) won't work as that is the one that is broken. Need to pin it to 1.7.26 which I did in the other PR. Can move the changes over here if you'd like @tyranron

dt665m:
> Sorry didn't see this PR before I created #856. As I noted there, just using the latest version (currently 1.7.27) won't work as that is the one that is broken. Need to pin it to 1.7.26 which I did in the other PR. Can move the changes over here if you'd like @tyranron

for some reason, leaving out the semver on jsdeliver resolves to a working version.  We're running off my PR here and it works.  🤷‍♂️

Zireael13:
> for some reason, leaving out the semver on jsdeliver resolves to a working version. We're running off my PR here and it works. 🤷‍♂️

https://cdn.jsdelivr.net/npm/graphql-playground-react/build/static/js/middleware.js

Huh you are right. That is really odd. jsDelivr must somehow resolve to a past version if the file isn't present in the new one, I guess.

I guess we'll leave it up to the repo maintainers if they want to go with the versionless one or pinned to 1.7.26.

LegNeato:
Generally I like explicit versioning as it helps reproducible builds as well as tracking down bugs. Thank you for the PR!

PR #861, opened 2021-01-25
Update juniper_codegen for syn 1.0.60

### LINKED ISSUE #858: Can't compile juniper_actix project.
**Describe the bug**
While I m trying to build project I m getting below error
```rs
 no variant or associated item named `__Nonexhaustive` found for enum `syn::Type` in the current scope
   --> juniper_codegen/src/common/parse/mod.rs:199:18
```

**To Reproduce**
(code example preferred)
Clone the latest code and try `cargo build` in `juniper_actix` folder.

**Expected behavior**
It should build without errors.

**Additional context**
Getting similar issue in project while pointing to git repo in cargo file.

obmarg:
syn 1.0.60 has updated it's `Type::__Nonexhaustive` to
`Type::TestExhaustive`, breaking juniper.  This updates juniper to use
the recommended idiom for doing exhaustive matching on `Type`, which
fixes this.

Not entirely clear if this needs exhaustive matching or if we could
just use a fallback, but this fixes the build at least.

I imagine this will break for users who are still on an old syn though.

Fixes #858

### DIFF
--- a/examples/basic_subscriptions/Cargo.toml
+++ b/examples/basic_subscriptions/Cargo.toml
@@ -13,5 +13,5 @@ serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 tokio = { version = "0.2", features = ["rt-core", "macros", "stream"] }
 
-juniper = { git = "https://github.com/graphql-rust/juniper" }
-juniper_subscriptions = { git = "https://github.com/graphql-rust/juniper" }
+juniper = { path = "../../juniper" }
+juniper_subscriptions = { path = "../../juniper_subscriptions" }
--- a/examples/warp_async/Cargo.toml
+++ b/examples/warp_async/Cargo.toml
@@ -6,8 +6,8 @@ publish = false
 authors = ["Christoph Herzog <chris@theduke.at>"]
 
 [dependencies]
-juniper = { git = "https://github.com/graphql-rust/juniper" }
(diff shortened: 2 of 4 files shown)


obmarg:
Also updated an example `Cargo.toml` to use path based deps for juniper as otherwise it pulled the broken code from github - let me know if that's not the right thing to do.

Binero:
For now this can be worked around by adding an explicit dependency to `syn = "=1.0.59` in your `Cargo.toml`. 

```toml
[dependencies]
actix-web = "3.3"
juniper = "0.15.2"
syn = "=1.0.59"
```

trevyn:
> I imagine this will break for users who are still on an old syn though.

We could just specify `syn` `1.0.60` in `juniper_codegen/Cargo.toml`, right? Technically that’s not a breaking change...

LegNeato:
Thank you for this! I think we should change it to 1.0.60 as that is now the min required version, correct?

LegNeato (changes requested):
As mentioned, I think we want cargo.toml changes

obmarg:
Updated the Cargo.toml

LegNeato:
Thank you! I'll get a crates.io to release out later tonight

kitsuneninetails:
Please push crate ASAP, as this completely derails our entire production-level application builds.  :pray: 

Thanks!

LegNeato:
This is now on crates.io

Issue #818, opened 2020-12-10
Switch to HashMap for the internal representation of object fields rather than a vector of tuples


LegNeato:
We store object fields and their types in a `Vec` of tuples. This means for field lookup we need to traverse the list. There are also some weird functions for merging two objects due to having to merge the two `Vec`s. 

Switching to some sort of `HashMap` for the internal representation of object fields rather than a `Vec` of tuples should probably be faster. This `HashMap` will be small and mostly read-only so there might be a better `HashMap` than the one in `std`

PR #872, opened 2021-02-09
Switch to HashMap for the internal representation of object fields

abrackx:
Hey there, first time contributing here. 

This PR updates Object's field key_value_list to use a `HashMap<String, Value<S>>` (#818).

Because HashMaps are unordered I had problems with a few tests that were saving values as strings and then comparing to other expected strings. The tests `object_query`, `display_object` and `batch_requests_work` were examples of that. I resolved these failures by doing some json object comparison but I'm not sure if that's the best approach. 

I am also not sure if I need to update the changelog for this so please let me know. 

I'll put this as a draft in case I completely missed the mark.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,7 +60,6 @@ jobs:
 
       - name: Build and run tests
         env:
-          RUSTFLAGS: "-C link-dead-code"
           CARGO_MAKE_RUN_CODECOV: true
         run: |
           cargo make workspace-ci-flow --no-workspace
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ your Schemas automatically.
 
 ## Guides & Examples
 
-- [Juniper + actix-web example](https://github.com/actix/examples/tree/master/juniper)
+- [Juniper + actix-web example](https://github.com/actix/examples/tree/master/graphql/juniper)
 
 ## API Stability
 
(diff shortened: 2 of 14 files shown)


LegNeato:
Sorry this took so long, this looks great! We have benchmarks in https://github.com/graphql-rust/juniper/tree/master/juniper_benchmarks, can you see if this improves or regresses?

LegNeato:
Note that the GraphQL specification requires result field order to stay the same as query field order...it might be best to use `IndexMap` here to preserve ordering. That will likely make the tests work as-is (assuming the issue is ordering).

Thank you again for the PR! 🍻

LegNeato (changes requested):
Let's preserve order.

abrackx:
Updated! IndexMap seemed to do the trick. 

I ran benchmarks on master, checked out this branch and then reran them. Here is the output: 
[after.txt](https://github.com/graphql-rust/juniper/files/6058142/after.txt) 

It looks like performance improved on `Sync vs Async - Users Flat - Instant/Sync/1` and `Sync vs Async - Users Flat - Instant/Async - Single Thread/10`. Otherwise performance did not change.

Let me know if there's anything else I missed. Thanks!

LegNeato (approved):

LegNeato:
Thank you for the change!  🍻

Sorry again it too so long to take a look at it, and welcome to Juniper!

LegNeato:

Issue #843, opened 2021-01-06
Persisted queries


shoooe:
**Is your feature request related to a problem? Please describe.**
Persisted queries aim at:
 - Improve the performance of a GraphQL server
 - Whitelist specific queries
 - Reduce bandwidth usage

**Describe the solution you'd like**
Something similar to [HotChocolate's persisted queries](https://chillicream.com/docs/hotchocolate/v10/execution-engine/persisted-queries).

**Describe alternatives you've considered**
I considered implementing a layer, between the HTTP request and the Juniper engine, that could convert an ID into a GraphQL query stored within the executable. 

That would solve the whitelisting problem and bandwidth usage but wouldn't improve the performance of the GraphQL server compared to just passing the query directly. 

For that Juniper would have to expose the parsing and executing as two different function calls. That way I could parse the whitelisted queries at server startup and just reuse those parsed queries time and time again.

**Additional context**
[Relay](https://relay.dev/docs/en/persisted-queries.html) describes persisted queries.
[Apollo](https://www.apollographql.com/blog/persisted-graphql-queries-with-apollo-client-119fd7e6bba5/) also seems to have support for those.

LegNeato:
We have an existing issue to split up the stages here: https://github.com/graphql-rust/juniper/issues/726

PR #874, opened 2021-02-13
Make the executor and validation APIs public to enable split parsing and execution

jerel:
Making these APIs public is based on my understanding of https://github.com/graphql-rust/juniper/pull/773#issuecomment-703783048 and https://github.com/graphql-rust/juniper/pull/773#issuecomment-704009918

This PR is related to #776, #726, and enables features similar to #843 to be implemented in user land code. In my project I'm currently doing this (pseudo code) in the first step:
``` rust
fn compile() {
  let document = parse_document_source(document_source, &root_node.schema)?;
  visit_all_rules(&mut ctx, &document);
  let operation = get_operation(&document, operation_name)?;
  Ok((document.clone(), operation.clone()))
}
```
so I can make use of the `operation` and `document` and then in the second step I'm using the [now public] `validation::validate_input_values` and `executor::execute_validated_query_async` or `resolve_validated_subscription` as needed.

Thoughts on these changes?

### DIFF
--- a/integration_tests/juniper_tests/Cargo.toml
+++ b/integration_tests/juniper_tests/Cargo.toml
@@ -8,6 +8,7 @@ publish = false
 derive_more = "0.99"
 futures = "0.3"
 juniper = { path = "../../juniper" }
+juniper_subscriptions = { path = "../../juniper_subscriptions" }
 
 [dev-dependencies]
 async-trait = "0.1.39"
--- a/integration_tests/juniper_tests/src/lib.rs
+++ b/integration_tests/juniper_tests/src/lib.rs
@@ -16,3 +16,5 @@ mod issue_371;
 mod issue_398;
 #[cfg(test)]
 mod issue_500;
+#[cfg(test)]
+mod pre_parse;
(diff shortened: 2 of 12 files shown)


jerel:
@LegNeato @ccbrown what are your thoughts on this proposed change?

Being able to precompile queries while checking them against the schema is a big requirement of a project I'm working on and ideally we could implement it in a way that stays aligned with this project.

ccbrown:
To sum up my thoughts from the linked issues:

* This is really valuable functionality that we should have in some form.
* Ideally parsing wouldn't require knowledge of the schema. Unfortunately this eliminating the parser's dependency on the schema isn't trivial. @mirosval made a valiant effort in https://github.com/graphql-rust/juniper/pull/782, but it fell through. Given the value of this functionality and the lack of a clear path towards eliminating the parser's schema dependency, I'm okay moving forward with exposing whatever's necessary for this functionality anyway.
* I have a preference for exposing this functionality in the inner crates (`juniper::parser`, `juniper::validation`, etc.) rather than the top level `juniper` crate, just for the sake of keeping the top level API narrow.

LegNeato (changes requested):
Sorry this took so long. Can you add an integration test that uses the API so we can be sure to not regress?

LegNeato (approved):
Thank you! 🍻

LegNeato:

mrtnzlml:
Hey @jerel! 👋 I have a question regarding this PR. Say you want to implement persistent queries, so in the first step, you parse and validate the query and save it to DB (so it can be accessed later while skipping validations). My question is: how should the validated query be stored in DB? Do you store it as a string (meaning that you must call `parse_document_source` later again)? Or is there a better way to skip the parsing on future requests? Thanks! :)

PR #883, opened 2021-02-25
Actix 4 incoming

kunjee17:
Currently in beta but works. Good news is that it is on Tokia 1.0. So, either we can have it with beta or wait for full version. I ll try to keep this branch up to date in any case.

### DIFF
--- a/juniper_actix/CHANGELOG.md
+++ b/juniper_actix/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 - Compatibility with the latest `juniper`.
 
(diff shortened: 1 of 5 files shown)


kunjee17:
I was unable to get subscription support to latest version. I might need some help to fix that. @engylemure  @mihai-dinculescu 

Current subscription support was in draft / beta mode. If things are alright we can get it as fully supported with latest actix 4.0 version.

kunjee17:
Closing this in favour of #912

PR #884, opened 2021-02-25
Fix CI

trevyn:
This adjusts a few tiny things that have changed in Rust nightly, and removes `-C link-dead-code` that was causing problems for the Windows GitHub Actions runners.

(I'm not quite sure where the `-C link-dead-code` part came from, but it does not appear to be present in the Azure script, so this should just bring us to parity and let the GHA runners succeed.)

This passes GHA on my branch: https://github.com/trevyn/juniper/actions/runs/598443306

(Is there anything preventing us from switching completely from Azure to GHA? This would make testing PRs in forks much easier.)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,7 +60,6 @@ jobs:
 
       - name: Build and run tests
         env:
-          RUSTFLAGS: "-C link-dead-code"
           CARGO_MAKE_RUN_CODECOV: true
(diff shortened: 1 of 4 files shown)


trevyn:
(spurious CI timeout failure after doc-only change)

LegNeato (approved):

LegNeato:
Thank you!

Issue #851, opened 2021-01-19
Support for `hyper 0.14` and `tokio 1.0` in `juniper_hyper`


alvskar:
`juniper_hyper` seems tied to the older versions of `hyper` and `tokio` at this time. Are there plans to update it? As I can see it now we'd need to fork it if we want to use the newer tokio (older hyper seems to rely on older tokio).

LegNeato:
https://github.com/graphql-rust/juniper/pull/849

Alexx-G:
@LegNeato @theduke @mhallin Is there a chance you'll have some bandwidth to take a look at this one? There are 2 open PRs ( #888 , #912 ) that close it.
In case you have concerns over stability, it's always possible to have a few "beta" versions to test the upgrade and collect feedback.

LegNeato:
This has now been pushed to crates.io

PR #849, opened 2021-01-18
Upgrade tokio 0.2 -> 1.0.2

akhramov:
Recently tokio got a first stable release and many libraries &
applications already migrated to the newest version.

This changes upgrades tokio version to 1.0.2:

* Tokio renamed some of its features, e.g `rt-util` and `rt-core` now
  combined into `rt`.

* `stream` feature got extracted to a separate crate
  [tokio-stream](https://docs.rs/tokio-stream/0.1.2/tokio_stream),
  waiting for eventual `Stream` landing to the Rust std
  library. [RFC](https://github.com/rust-lang/rfcs/pull/2996)

[TODO]

Actix's integration test_actix_ws_integration test still fails due to,
I guess, async rt is not being initialized. Apart from that, the tests
are green. Please feel free to take over this effort.

PR #888, opened 2021-02-26
Tokio upgrade

LegNeato:
This is a superset of https://github.com/graphql-rust/juniper/pull/849 because I don't have permission to update that branch's PR (or my git-fu isn't good enough).

### DIFF
--- a/juniper/Cargo.toml
+++ b/juniper/Cargo.toml
@@ -55,7 +55,7 @@ uuid = { version = "0.8", default-features = false, optional = true }
 bencher = "0.1.2"
 pretty_assertions = "0.7.1"
 serde_json = "1.0.2"
-tokio = { version = "1.0.2", features = ["macros", "rt", "time"] }
+tokio = { version = "1.0.2", features = ["macros", "rt-multi-thread", "time"] }
 
 [[bench]]
 name = "bench"
--- a/juniper/Cargo.toml
+++ b/juniper/Cargo.toml
@@ -55,7 +55,7 @@ uuid = { version = "0.8", default-features = false, optional = true }
 bencher = "0.1.2"
 pretty_assertions = "0.7.1"
 serde_json = "1.0.2"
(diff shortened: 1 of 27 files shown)


LegNeato:
The reason why the actix stuff is failing on CI is that CI runs `cargo test --all-features`, which enables the ws integration

juniper/Cargo.toml
```diff
@@ -54,7 +54,7 @@ uuid = { version = "0.8", default-features = false, optional = true }
 bencher = "0.1.2"
 pretty_assertions = "0.6.1"
 serde_json = "1.0.2"
-tokio = { version = "0.2", features = ["macros", "rt-core", "time"] }
+tokio = { version = "1.0.2", features = ["macros", "rt-multi-thread", "time"] }
```
samuela, line 57: Is there a reason to pin to 1.0.2 instead of the latest 1.2.0?

samuela:
This will close https://github.com/graphql-rust/juniper/issues/851 I think.

juniper/Cargo.toml
trevyn: > Is there a reason to pin to 1.0.2 instead of the latest 1.2.0?

This means "1.x.x >= 1.0.2", i.e. "^1.0.2", not "=1.0.2". It's best for libraries to refer to the lowest version required that is semver-compatible with latest, in case a host project needs to use an earlier, but still compatible version.

LukasKalbertodt:
Could someone explain what this PR is waiting for or blocked by? Can I somehow help to finish this and get this merged? What still has to happen here?

toxeus:
@LukasKalbertodt I think PR https://github.com/graphql-rust/juniper/pull/912 is in a better shape than this one. It's basically missing a rebase to incorporate the fixes for the nightly compiler. If it gets merged is a different question.

LegNeato:
Landed as https://github.com/graphql-rust/juniper/pull/912

Issue #819, opened 2020-12-10
Use a small and/or read-optimized string for internal field representation


LegNeato:
we store the name of each field as a string. Fields are likely on average pretty short (think "id", "name", etc) and are read more than written. We should investigate using a small and/or read-optimized string for internal field representation instead of `std`'s `String`.

PR #899, opened 2021-03-21
Use smartstring for name of field (#819).

andrisak:
Replaced name of field with a `smartstring` instead of `String`. Running `cargo bench --features expose-test-schema` indicates a slight improvement on the `query_type_name` test and slightly lower performance on the `introspection_query` test. Here is an example of two runs:

**Master**:
test introspection_query ... bench:     962,848 ns/iter (+/- 70,973)
test query_type_name     ... bench:      11,557 ns/iter (+/- 894)

**This PR/branch**:
test introspection_query ... bench:     978,076 ns/iter (+/- 70,523)
test query_type_name     ... bench:      11,354 ns/iter (+/- 1,471)


LegNeato (approved):
Thank you!

LegNeato:

PR #907, opened 2021-04-02
Fragment spreads in Interface types

### LINKED ISSUE #407: Panic when using fragment of a implementation in a interface
**Describe the bug**
Juniper panic when using a fragment for an implementation directly in a interface.

**To Reproduce**
```rust
use juniper::*;

struct Query;

#[juniper::object]
impl Query {
    fn things() -> Vec<Thing> {
        vec![Thing::Foo(Foo { foo: 1 }), Thing::Bar(Bar { bar: 2 })]
    }
}

enum Thing {
    Foo(Foo),
    Bar(Bar),
}

juniper::graphql_interface!(Thing: () where Scalar = <S> |&self| {
    instance_resolvers: |_| {
        &Foo => match *self { Thing::Foo(ref foo) => Some(foo), _ => None },
        &Bar => match *self { Thing::Bar(ref bar) => Some(bar), _ => None },
    }
});

struct Foo {
    foo: i32,
}

#[juniper::object(
    interfaces = [&Thing],
    scalar = S
)]
impl<S: ScalarValue> Foo {
    fn foo(&self) -> i32 {
        self.foo
    }
}

struct Bar {
    bar: i32,
}

#[juniper::object(
    interfaces = [&Thing],
    scalar = S
)]
impl<S: ScalarValue> Bar {
    fn bar(&self) -> i32 {
        self.bar
    }
}

type Schema = juniper::RootNode<'static, Query, EmptyMutation<()>>;

#[test]
fn test_fragment_in_interface_does_not_work() {
    let _ = juniper::execute(
        r#"
            query Query {
                things {
                    ...fooFields
                    ...barFields
                }
            }

            fragment fooFields on Foo {
                foo
            }

            fragment barFields on Bar {
                bar
            }
        "#,
        None,
        &Schema::new(Query, EmptyMutation::new()),
        &Variables::new(),
        &(),
    )
    .unwrap();
}

#[test]
fn test_fragment_in_interface_works() {
    let _ = juniper::execute(
        r#"
            query Query {
                things {
                    ...on Foo {
                        ...fooFields
                    }
                    ...on Bar {
                        ...barFields
                    }
                }
            }

            fragment fooFields on Foo {
                foo
            }

            fragment barFields on Bar {
                bar
            }
        "#,
        None,
        &Schema::new(Query, EmptyMutation::new()),
        &Variables::new(),
        &(),
    )
    .unwrap();
}
```

Error:
`thread 'fragment_in_interface::test_fragment_in_interface_does_not_work' panicked at 'Field foo not found on type Some("Thing")', juniper\src\types\base.rs:385:21`

**Expected behavior**
Both queries should be equivalent.

dyedgreen:
Closes #407

### DIFF
--- a/integration_tests/juniper_tests/src/issue_407.rs
+++ b/integration_tests/juniper_tests/src/issue_407.rs
@@ -54,23 +54,25 @@ type Schema = juniper::RootNode<'static, Query, EmptyMutation, EmptySubscription
 
 #[tokio::test]
 async fn test_fragments_in_interface() {
+    let query = r#"
+        query Query {
+            characters {
+                ...HumanFragment
+                ...DroidFragment
+            }
+        }
+
+        fragment HumanFragment on Human {
+            name
+        }
+
+        fragment DroidFragment on Droid {
+            serialNumber
+        }
+    "#;
+
     let (_, errors) = juniper::execute(
-        r#"
-            query Query {
-                characters {
-                    ...HumanFragment
-                    ...DroidFragment
-                }
-            }
-
-            fragment HumanFragment on Human {
-                name
-            }
-
-            fragment DroidFragment on Droid {
-                serialNumber
-            }
-        "#,
+        query,
         None,
         &Schema::new(Query, EmptyMutation::new(), EmptySubscription::new()),
         &Variables::new(),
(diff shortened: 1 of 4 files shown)


dyedgreen:
As a note: I'm not sure if there is a difference between the `async` and sync type resolution which I'm missing with this. The tests I added are async, so the fix might miss out on the sync case of this bug (if it exists).

tyranron:
@dyedgreen thanks for the contribution!

Actually, yes. The sync execution is implemented separately (it has similar name and flow). So, it would be nice to provide a fix for it too and cover it with tests.

LegNeato (approved):
Thank you so much for the PR and including tests! 🍻

LegNeato:

PR #915, opened 2021-04-15
Fixes Issue 914

### LINKED ISSUE #914: Multiple fragments on sub types override each other
**Describe the bug**
If a query has multiple fragments selecting the same sub type, the selected fields override each other

**To Reproduce**
Steps to reproduce the behavior: Run a query like
```graphql
query Me {
  loggedInUser {
    id
    ...PicOne
    ...PicTwo
  }
}

fragment PicOne on User {
  profilePicture {
    id
    width
    height
  }
}

fragment PicTwo on User {
  profilePicture {
    id
    url
  }
}
```

Returns only the fields of the later fragment.

**Expected behavior**
It should merge the selected sub fields and return both values.

dyedgreen:
Closes #914 

This fixes the issue raised and adds a test to prevent future regressions.

Note that this slightly changes to public api of `juniper::Object`'s: `add_field` will now check if a new field exists and is an object, and in that case merge the objects keys instead of replacing the old object.

It might also make sense to not return anything from that function (which is a breaking change of the public interface, but seems to compile without modification to the existing uses in the juniper crate, I think the behavior was originally copied from `std::collections::HashMap`).

### DIFF
--- a/integration_tests/juniper_tests/src/issue_914.rs
+++ b/integration_tests/juniper_tests/src/issue_914.rs
@@ -1,0 +1,85 @@
+use juniper::*;
+
+struct Query;
+
+#[derive(GraphQLObject)]
+struct Foo {
+    bar: Bar,
+}
+
+#[derive(GraphQLObject)]
+struct Bar {
+    a: i32,
+    b: i32,
+}
+
+#[graphql_object]
+impl Query {
+    fn foo() -> Foo {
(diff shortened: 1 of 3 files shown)


dyedgreen:
(The CI fail seems to be related to rocket; but I'm not sure if that was caused by my change 😬)

LegNeato:
Thank you for the fix! :beers: The breakage is indeed some Rocket changes that have broken us.

PR #912, opened 2021-04-12
Upgraded tokio, warp, hyper and actix

engylemure:
### Upgraded to

- `tokio = 1`
- `warp = 0.3`
- `hyper = 0.14`
- `actix = 4.0.0-beta.5`

PR #916, opened 2021-04-15
Pin rocket to specific revision 

toxeus:
Rocket is in active development with many breaking
changes. To prevent breaking the CI for unrelated
pull requests pin rocket down to a specific revision
instead of following master blindly.

This project is already struggling with breaking changes in the nightly compiler. That's the reason why this PR fails CI, too.

### DIFF
--- a/juniper_rocket_async/Cargo.toml
+++ b/juniper_rocket_async/Cargo.toml
@@ -14,7 +14,7 @@ repository = "https://github.com/graphql-rust/juniper"
 [dependencies]
 futures = "0.3.1"
 juniper = { version = "0.15.4", path = "../juniper", default-features = false }
-rocket = { git = "https://github.com/SergioBenitez/Rocket", branch = "master", default-features = false }
+rocket = { git = "https://github.com/SergioBenitez/Rocket", rev = "50c9e88cf91cbcc6298cbd04282aebf6fba69b88", default-features = false }
 serde_json = "1.0.2"
 tokio = { version = "1", features = ["macros", "rt"] }
 


toxeus:
Just noticed that there's a PR for that already.

LegNeato:
Thank you for this! I believe folks using Rocket master will generally want to track it. It is also easier to make changes along the way instead of waiting and doing massive changes. I have put up a diff to support the latest: https://github.com/graphql-rust/juniper/pull/919

Thank you again for doing this!

LegNeato:
(We stress the nightly compiler a lot and report bugs FWIW)

toxeus:
> I believe folks using Rocket master will generally want to track it.

That would be me then. I *am* tracking Rocket master. And I want `juniper_rocket_async` to track Rocket master. But in a controlled way. The current approach is a bit painful to work with to be honest.

I think an approach where we pin down Rocket and bump it once a month is a more stable one then blindly following master. Blindly following master breaks revision control: I can checkout an old version of `juniper_rocket_async` but its Rocket dependency will point to latest master and the build will fail. Then, I need to bisect through Rocket's master branch to find a commit that compiles. I think that's not how it's supposed to be.

> (We stress the nightly compiler a lot and report bugs FWIW)

I think it is very honorable to contribute to the Rust community and test and report nightly compiler bugs. My impression is just that recently PRs are piling up that fail the CI because issues with the nightly compiler (including your Rocket master fix above). And there seems to be a lack of resources to fix the nightly compiler issues. To most `juniper` users [landing the tokio 1.x bump](https://github.com/graphql-rust/juniper/pull/912) is more valuable than being able to compile on nightly. In fact, the update to tokio 1.x is a very painful one since all crates with a tokio dependency need to follow in one go and in my case `juniper` is the only blocker left for the tokio 1.x upgrade which is a bit upsetting. 

I hope that my feedback comes across as constructive. `juniper` is a great project that's why I'm investing time here to share feedback from a user's perspective to make things even better :)

PR #923, opened 2021-04-29
Fix fields on interface not being resolved when used with fragments

### LINKED ISSUE #922: Fields on interface fragment spreads don't get resolved
**Describe the bug**

When resolving fields defined on an interface, if the field is defined within a fragment spread, then it is omitted from the output.

I'm happy to work on a fix for this, but will need some guidance on where to look (and whatever other tips/hints you could give me).

**To Reproduce**

Using the schema defined in `juniper/src/tests/fixtures/starwars/` (I ran the `juniper_rocket_async` example to reproduce):

### Working

A simple query returns the expected result (`id` and `name` fields are populated).

```graphql
query {
  hero {
    id
    name
  }
}
```

```json
{
  "data": {
    "hero": {
      "id": "2001",
      "name": "R2-D2"
    }
  }
}
```

### Broken

But this query, using a fragment spread on the `Character` interface, is broken (`id` and `name` fields are omitted in the response):

```graphql
{
  hero {
    ... on Character {
      id
      name
    }
  }
}
```

```json
{
  "data": {
    "hero": {}
  }
}
```

**Expected behavior**

The second query above should return the exact same output as the first query.

**Additional context**

I encountered this problem using Relay in my UI, as parts of the query being generated by Relay use this pattern. Which leads me to believe that these queries should definitely be valid.

ArsileLuci:
Closes #922 

PR Checklist

 - [x] Ensure proper formatting
 - [x] Run all tests
 - [x] Update the CHANGELOG

### DIFF
--- a/integration_tests/juniper_tests/src/issue_922.rs
+++ b/integration_tests/juniper_tests/src/issue_922.rs
@@ -6,7 +6,6 @@ struct Query;
 impl Query {
     fn characters(executor: &Executor) -> Vec<CharacterValue> {
         executor.look_ahead();
-
         vec![
             Into::into(Human {
                 id: 0,
--- a/integration_tests/juniper_tests/src/issue_922.rs
+++ b/integration_tests/juniper_tests/src/issue_922.rs
@@ -96,14 +96,13 @@ async fn test_fragment_on_interface() {
     .unwrap();
 
     assert_eq!(errors.len(), 0);
-
     assert_eq!(
         res,
         graphql_value!({
             "characters": [
                 {"__typename": "Human", "id": 0, "name": "human-32"},
-                {"__typename": "Droid", "id": 1, "name": "R2-D2"}
-            ]
+                {"__typename": "Droid", "id": 1, "name": "R2-D2"},
+            ],
         }),
     );
 
@@ -117,14 +116,13 @@ async fn test_fragment_on_interface() {
     .unwrap();
 
     assert_eq!(errors.len(), 0);
-
     assert_eq!(
         res,
         graphql_value!({
             "characters": [
                 {"__typename": "Human", "id": 0, "name": "human-32"},
-                {"__typename": "Droid", "id": 1, "name": "R2-D2"}
-            ]
+                {"__typename": "Droid", "id": 1, "name": "R2-D2"},
+            ],
         }),
     );
 }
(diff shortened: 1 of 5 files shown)


juniper/src/types/async_await.rs
```diff
@@ -306,7 +306,11 @@ where
                 );
 
                 let concrete_type_name = instance.concrete_type_name(sub_exec.context(), info);
-                if fragment.type_condition.item == concrete_type_name {
+                let type_name = instance.type_name(info);
+
+                if fragment.type_condition.item == concrete_type_name
+                    || Some(fragment.type_condition.item) == type_name
+                {
```
tyranron, line 312: @ArsileLuci you've considered async resolving only. We need to consider sync resolving too.

integration_tests/juniper_tests/src/issue_922.rs
```diff
@@ -0,0 +1,121 @@
...
+        for obj in values {
+            if let juniper::Value::Object(obj) = obj {
+                assert!(obj.contains_field("id"), "id field should be present");
+                assert!(obj.contains_field("name"), "name field should be present");
+            } else {
+                assert!(false, "List should contain value");
+            }
+        }
+    } else {
+        assert!(false, "List should be returned")
```
tyranron, line 118: That's somewhat ackward assertions. We can do better by using `graphql_value!` macro. See integration tests for interfaces, for example.

integration_tests/juniper_tests/src/issue_922.rs
```diff
@@ -0,0 +1,121 @@
+use juniper::*;
+
+struct Query;
+
+#[juniper::graphql_object]
+impl Query {
+    fn characters(executor: &Executor) -> Vec<CharacterValue> {
+        executor.look_ahead();
```
tyranron, line 8: @ArsileLuci why do we need `executor` here? Does it influence test somehow? Remove it, if it's unnecessary.

integration_tests/juniper_tests/src/issue_922.rs
```diff
@@ -0,0 +1,121 @@
...
+                name
+            }
+            ... on Droid {
+                id
+                name
+            }
+        }
+    "#;
+
+    let (res, errors) = juniper::execute(
```
tyranron, line 89: @ArsileLuci consider to cover both `execute` (async) and `execute_sync` in tests.

tyranron (changes requested):
Thanks! But some improvements are required.

tyranron (approved):
@ArsileLuci GJ, thanks! 🍻

PR #926, opened 2021-05-07
Fix "Switch to HashMap for the internal representation of object fields (#872)"

xDarksome:
#872 has introduced broken `Object::partial_eq` [implementation](https://github.com/graphql-rust/juniper/pull/872/files#diff-1ce35005f42b1b24ff8638fe5093453b57e3be4164d7d7c417de17de134452c5R12) therefore rendering some tests useless.
 Also, `Object` field merging functionality was completely removed in that PR for some reason, which should have resulted in failing tests but didn't. 

#914 was caused by this and #915 partially resolved the issue by implementing the same thing in the different place.

This PR rollbacks the `Object` implementation to its original state, switches the underlying type to `HashMap` and (hopefully) breaks nothing else.


dyedgreen:
Do you think it makes sense to add a test for the partial eq? (I saw that it’s derived now; but having a test like this might prevent a similar regression in a future refactor; especially seeing the trait being implemented correctly is so important for a lot of the other existing tests)

tyranron:
@dyedgreen don't think so. I think the test has sense only if the code written manually. So, it should've been added when manual implementation was added.

tyranron (approved):
@xDarksome thaks! 🍻

PR #932, opened 2021-05-23
Update bson crate to 2.0

grantperry:
bson is currently 2.0.0-beta, but the changes I've made are at least a step closer to the bson 2.0 stable release.

I've updated the related tests as well. just running the tests on `juniper` works perfectly, but running ci-flow spits out a fair bit.


LegNeato (changes requested):
Test failures look legit?

juniper/src/integrations/bson.rs
```diff
@@ -36,13 +36,13 @@ where
     S: ScalarValue,
 {
     fn resolve(&self) -> Value {
-        Value::scalar((*self).to_rfc3339())
+        Value::scalar(DateTime::from(*self).to_rfc3339())
     }
```
patrickfreed, line 40: These compilation failures can be avoided by adding the `chrono-0_4` feature flag to the `bson` dependency (this was added in 2.0).

Issue #877, opened 2021-02-14
Issues using graphql_scalar proc macro inside a macro


urkle:
I have a construct where I need to generate several scalars that all behave similarly so I created a macro to wrap my code. This worked in Juniper 0.14.2 via the gaphql_scalar!, however switching it to the #[graphql_scalar] is not working.

```rust
#[macro_export]
macro_rules! id_type {
    ( $name:ident, $value:ident ) => {
        pub type $name = ID<$value>;

        juniper::graphql_scalar!($name where Scalar = <S> {
            // Define how to convert your custom scalar into a primitive type.
            resolve(&self) -> Value {
                juniper::Value::scalar(self.to_base64())
            }

            // Define how to parse a primitive type into your custom scalar.
            from_input_value(v: &InputValue) -> Option<$name> {
                v.as_scalar_value::<String>()
                    .and_then(|s| $name::from_base64(s).ok())
            }

            // Define how to parse a string value.
            from_str<'a>(value: ScalarToken<'a>) -> juniper::ParseScalarResult<'a, S> {
                if let juniper::parser::ScalarToken::String(value) = value {
                    Ok(S::from(value.to_owned()))
                } else {
                    Err(juniper::parser::ParseError::UnexpectedToken(juniper::parser::Token::Scalar(value)))
                }
            }
        });
    }
}
```

New version (that fails)
```rust
#[macro_export]
macro_rules! id_type {
    ( $name:ident, $value:ident ) => {
        pub type $name = ID<$value>;

        #[juniper::graphql_scalar]
        impl<S>  juniper::GraphQLScalar for $name
            where
                S: juniper::ScalarValue
        {
            // Define how to convert your custom scalar into a primitive type.
            fn resolve(&self) -> juniper::Value {
                juniper::Value::scalar(self.to_base64())
            }

            // Define how to parse a primitive type into your custom scalar.
            fn from_input_value(value: &juniper::InputValue) -> Option<$name> {
                value.as_scalar_value()
                    .and_then(|v| v.as_str())
                    .and_then(|s| $name::from_base64(s).ok())
            }

            // Define how to parse a string value.
            fn from_str<'a>(value: juniper::ScalarToken<'a>) -> juniper::ParseScalarResult<'a, S> {
                <String as juniper::ParseScalarValue<S>>::from_str(value)
            }
        }
    }
}
```

Error that occurs
```
error: GraphQL scalar unable to find target for implementation target for `GraphQLScalar`
[...]

PR #933, opened 2021-05-23
Fix using graphql_scalar proc macro inside a macro

ihor-rud:
Fix https://github.com/graphql-rust/juniper/issues/877


LegNeato:
Thank you! Sorry for the slow response.

LegNeato:
This is now released on crates.io

PR #935, opened 2021-05-24
Don't require `()` TypeInfo for schema language methods

sujayakar:
The `as_schema_language` and `as_parser_document` methods are currently only implemented when `QueryT::TypeInfo = ()` (and the same for `MutationT` and `SubscriptionT`). This requirement is superfluous, so we can just move these methods to the other `impl` block that doesn't have these constraints on `TypeInfo`.


sujayakar:
Hmm, it looks like the failing actions are also broken on `master`. Let me know if I should take another look.

LegNeato:
Thanks for the pr! 🍻 . Sorry it took so long to merge.

LegNeato:
Yep, I plan to do it tonight!

LegNeato:
This is now released on crates.io

PR #946, opened 2021-06-11
Fixes panic when spreading untyped union fragment.

### LINKED ISSUE #945: Panic when spreading untyped union fragment
**Describe the bug**
Updated my project from `0.15.3` to `0.15.5` and started observing the following sort of panic when spreading an untyped union fragment. Note that the specified concrete type is actually the union type.
`panicked at 'Concrete type <UnionType> is not handled by instance resolvers on GraphQL union <UnionType>`

**To Reproduce**
Example schema:

```graphql
union Character = Human | Droid
type Droid {
  id: ID!
  sensorColor: String!
}
type Human {
  id: ID!
  eyeColor: String!
}

type Query {
  mainCharacter: Character!
}
```

with the following query:

```graphql
query Query {
  mainCharacter {
    ...CharacterFragment
  }
}

fragment CharacterFragment on Character {
  __typename
  ... on Human {
    id
    eyeColor
  }
  ... on Droid {
    id
    sensorColor
  }
}
```

**Expected behavior**
No panic. And some result like:
```json
{
  "mainCharacter": { "__typename": "Droid", "id": 1, "sensorColor": "red" }
}
```

**Additional context**
I have an incoming PR that seems to fix this issue while still passing the tests for #922. Looks like it should be as simple as using the `concrete_type_name` when invoking `instance.resolve_into_type_async`.

imintz:
closes #945 

See the included test for an example repro of this panic. I haven't spent enough time to fully grok the flow of the code here, so please let me know if I'm pulling the wrong lever and accidentally fixing the issue.

### DIFF
--- a/integration_tests/juniper_tests/src/issue_945.rs
+++ b/integration_tests/juniper_tests/src/issue_945.rs
@@ -1,0 +1,95 @@
+use juniper::*;
+
+struct Query;
+
+#[graphql_object]
+impl Query {
+    fn artoo() -> Character {
+        Character::Droid(Droid {
+            id: 1,
+            name: "R2-D2".to_owned(),
+            sensor_color: "red".to_owned(),
+        })
+    }
+}
+
+#[derive(GraphQLUnion)]
+enum Character {
+    Droid(Droid),
+    #[allow(dead_code)]
+    Human(Human),
+}
+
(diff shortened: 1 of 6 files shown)


LegNeato:
Thanks! 🍻

LegNeato:

PR #952, opened 2021-06-29
Actix 4 support

engylemure:
At the moment it's a work in progress since version 4 is still in development at the beta stage.


LegNeato:
Any word on when they will cut a release? I can't ship to crates.io with git dependencies and I would like to update with the latest buts here. So I'd prefer to land changes focusing on the latest beta or update this and land once the git deps are on crates.io

fourbytes:
Just a heads up that the latest betas of `actix-cors` and `actix-identity` support `actix-web` beta.8. It would be great if we could get a pre-release pushed to crates.io with these.

LegNeato:

Issue #947, opened 2021-06-17
Add support for Rocket 0.5


Axel-2:
**Is your feature request related to a problem? Please describe.**
Hello, since one week, Rocket 0.5 was released and Juniper-Rocket does not work with this new version. 

**Describe the solution you'd like**
Add support for Rocket 0.5

LegNeato:
We have support in-repo, we just need to make `juniper_rocket_async` be released as `juniper_rocket`

PR #955, opened 2021-07-06
Make juniper_rocket_async replace juniper_rocket

LegNeato:
Now that rocket v0.5 is async and in rc, merge the two juniper projects.

Fixes https://github.com/graphql-rust/juniper/issues/947.


LegNeato:

PR #966, opened 2021-07-19
Allow using Rust arrays as GraphQL lists (#918)

### LINKED ISSUE #918: Does not GraphQLObject support [u8; N] slice?
I have a struct with field of [u8; 5], but seems juniper does not support this type of data:

```rust
#[derive(GraphQLObject)]
pub struct Test {
    demo: [u8; 5]
}
```

the trait `GraphQLValue<__S>` is not implemented for `[u8; 9]`

tyranron:
Fixes #918

Allows usage of Rust arrays as GraphQL lists.


tyranron:
@LegNeato the main challenge imposed by this PR is that for input values Rust array validation is stricter than for vectors and slices, because we have to ensure that input GraphQL list length matches the one of the declared Rust array.

Unfortunately, current `juniper` architecture keeps input validation and user defined types far appart. Everything we declare implmenting `GraphQLType`, `GraphQLValue` (and others) comes into play after the input query was validated, and the current validation mechanism has no knowledge about expected list sizes, which results in panicking here and there when Rust arays are used as inputs.

Idealy, to overcome this, it would be nice to have an ability to participate in validation for custom defined types (like visitor pattern, for example), but that goes far beyond theme of this PR and requires quite a massive refactoring of ast/validation layer.

For now, I've just tweaked the ast/meta/validation layers with an optional `expected_size` information for lists. Which is `None` for vectors and slices, but drives in for arrays.

Additionally, gradual Rust array initialization requires some additional `unsafe` squats in the current stable Rust. I've tried my best to provide enough `// SAFETY:` comments to explain what is happening and why.

LegNeato:
I had a change for this but looks like you beat me to the punch! I'll take a look soon, but at quick glance I didn't see Input Coercion , but perhaps I missed it:

https://spec.graphql.org/June2018/#sec-Type-System.List

juniper/src/types/containers.rs
```diff
@@ -257,6 +256,155 @@ where
...
+                        //         exactly the number of added `items`.
+                        unsafe {
+                            ptr::drop_in_place(elem.as_mut_ptr());
+                        }
+                    }
+                    None
+                }
+            }
+            ref other => {
+                other.convert().and_then(|e: T| {
```
tyranron, line 373: @LegNeato here is the input coercion in `FromInputValue` implementation. It's litterally the same as for `Vec`s, but additionally does a length check.

juniper/src/types/utilities.rs
```diff
@@ -24,11 +24,25 @@ where
...
+                    }
+                }
+                items
+                    .iter()
+                    .all(|i| is_valid_literal_value(schema, inner, &i.item))
+            }
+            ref v => {
+                if let Some(expected) = expected_size {
+                    if expected != 1 {
+                        return false;
```
tyranron, line 41: @LegNeato and here is an additional validation for input coersion, when array is used.

integration_tests/juniper_tests/src/array.rs
```diff
@@ -0,0 +1,257 @@
...
+            .unwrap_err()
+            .to_string()
+            .contains(r#"Invalid value for argument "input", expected type "[Boolean!]!""#));
+    }
+
+    #[tokio::test]
+    async fn can_coerce_from_raw_value_if_single() {
+        let query = r#"
+            query Query {
+                first(input: true)
```
tyranron, line 211: @LegNeato we also do have some tests checking this works as expected.

tyranron:
@LegNeato additionally, I've reimplented `unsafe` squats a bit in `FromInputValue` implementation. The previous one was not panic safe.

LegNeato (approved):
I am not qualified to review the unsafe code but everything else looks good to me! Thank you so much, solid work

LukasKalbertodt:
If this change could be released, that'd be great :)

PR #968, opened 2021-07-21
Provide access for integration GraphQLRequest wrappers to its inner types (#930)

### LINKED ISSUE #930: Consider making inner GraphQLRequest public or implement Display
Hello team,

We're using `juniper_rocket_async` and it all works great, but I can't find a way of accessing the inner graphql request object. I would like to log it, either as just a string or as json.

 
Specifically, it would be convenient if this inner type was public https://github.com/graphql-rust/juniper/blob/9c9639e8f0da0337880759c6fde469f863f74841/juniper_rocket_async/src/lib.rs#L64

However, that maybe is an API change you're not very interested in. Any ideas for a better solution? I would be happy to provide a PR.

Best,
Knut

tyranron:
Fixes #930

For all `juniper_*` integrations makes `juniper::GraphQLBatchRequest` accessible from newtype wrappers.

### DIFF
--- a/juniper_rocket/CHANGELOG.md
+++ b/juniper_rocket/CHANGELOG.md
@@ -1,6 +1,7 @@
 # master
 
 - Compatibility with the latest `juniper`.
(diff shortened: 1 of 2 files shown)


tyranron:
Seems that only `juniper_rocket` has that.

Issue #880, opened 2021-02-18
Argument default values and descriptions not available in introspection


mrtnzlml:
Hello, me again! 😅 

**Describe the bug**

Argument default values and descriptions are not available in introspection (and via `as_schema_language`). I followed this syntax: https://graphql-rust.github.io/juniper/master/types/objects/complex_fields.html?highlight=arguments#customizing-arguments

**To Reproduce**

1. create a field with arguments like so:

```rust
#[juniper::graphql_object]
impl Product {
    #[juniper::graphql(arguments(first(default = 1, description = "TEST")))]
    fn images(&self, first: i32) -> Vec<Image> {
        // …
    }
    // …
}
```

2. introspect the schema and try to get the default value and description (it's not there)

**Expected behavior**

The default value and description should be available via introspection.

**Additional context**

Here is how the introspection looks like in my case (notice the missing arg description and default value):

```json5
{
  name: "images",
  description: "A list of images for this product, meant to be displayable to the customer.",
  args: [
    {
      name: "first",
      description: null,
      type: {
        kind: "NON_NULL",
        name: null,
        ofType: {
          kind: "SCALAR",
          name: "Int",
          ofType: null,
        },
      },
      defaultValue: null,
    },
  ],
  type: {
    kind: "NON_NULL",
    name: null,
    ofType: {
      kind: "LIST",
      name: null,
      ofType: {
        kind: "NON_NULL",
        name: null,
        ofType: {
          kind: "OBJECT",
          name: "Image",
          ofType: null,
        },
      },
    },
  },
  isDeprecated: false,
  deprecationReason: null,
}
```

tyranron:
@mrtnzlml the attribute on a field should be without a prefix: `#[graphql(...)]`, not  `#[juniper::graphql(...)]`.

mrtnzlml:
@tyranron I just tried to write it like this and it doesn't seem to make any difference:

```rust
#[juniper::graphql_object]
impl Product {
    #[graphql(arguments(first(default = 1, description = "TEST")))]
    fn images(&self, first: i32) -> Vec<Image> {
        // …
    }
    // …
}
```

This also makes me think: should it throw some error when I use `juniper::graphql`? Newbie question.

tyranron:
@mrtnzlml it should work, there is a lot of open code proving it works, including integrations tests in this repository. Maybe you miss some step.

> should it throw some error when I use juniper::graphql?

[...]

Issue #875, opened 2021-02-13
Async function on graphql_object with lifetime


TilBlechschmidt:
**Describe the bug**
When defining a struct, giving it a lifetime and then writing a `impl` block with the `graphql_object` macro containing a async function which accesses anything from `self`, the compiler starts complaining that `Sync` from somewhere deep in the `graphql_object` macro is less specific than `Sync` (wherever that comes from).

```rust
error[E0308]: mismatched types
   --> src/main.rs:5:1
    |
54  | #[graphql_object(context = Context)]
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    | |
    | one type is more general than the other
    | in this macro invocation
    |
   ::: /[...]/.cargo/git/checkouts/juniper-29ae65ca2e7a9997/a9f6464/juniper_codegen/src/lib.rs:467:1
    |
467 | pub fn graphql_object(args: TokenStream, input: TokenStream) -> TokenStream {
    | --------------------------------------------------------------------------- in this expansion of `#[graphql_object]`
    |
    = note: expected type `Sync`
               found type `Sync`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0308`.
```

**To Reproduce**
Take the lifetime example from the `graphql_object` documentation and make the function async. Bit of a silly example as it does not have to be async but it proves the point (same thing happens when you actually use more complex async tasks).
Here is the bare-bones code snippet to reproduce the error:
```rust
struct WithLifetime<'a> {
    value: &'a str,
}

#[juniper::graphql_object]
impl<'a> WithLifetime<'a> {
    async fn value(&self) -> &str {
        self.value
    }
}
```

**Expected behavior**
I'd expect it to work just the same with async as long as the reference to the object is retained in the place where the function is called 🤷

tyranron:
I think this happens due to `#[graphql_object]` macro desugars its method to closures:
```rust
#[graphql_object]
impl<'a> WithLifetime<'a> {
    async fn value(&self) -> &str {
        self.value
    }
}

// desugars in resolver code to something like this:
let res = executor.resolve_field(|| self.value);
```

And when lifetimes are involved, Rust usually generates types for closures using HRTBs: `for<'a>`. And that's the point where "less specific" error comes from.

[...]

Issue #827, opened 2020-12-14
having a variable named 'args' in a function leads to weird error messages


beevans:
I had a variable I'd named "args" in a function.  It lead to a whole bunch of compile errors on an unrelated variable when the juniper::graphql_object macro was unpacked.

It would be nice to have an error message that detects reserved words in an argument list.

LegNeato:
We should make these hermetic or mangle the names to make it work.

LegNeato:
So you happen to have a quick repro and the error message?

beevans:
I had a function whose header was basically this:

#[juniper::graphql_object(Context = Context)]
impl TestMutation {
    async fn run_task_test(
        context: &Context,
        jobname: String,
        taskname: String,
        fsname: String,
        args: String,
        fidlist: Vec<String>,
    ) -> juniper::FieldResult<bool> {
    func(fidlist);
};
....

The errors were:

error[E0277]: the type `str` cannot be indexed by `std::vec::Vec<std::string::String>`
 --> example.rs:57:1
  |
57 | #[juniper::graphql_object(Context = Context)]
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ string indices are ranges of `usize`
  |
  = help: the trait `std::slice::SliceIndex<str>` is not implemented for `std::vec::Vec<std::string::String>`
  = note: this error originates in an attribute macro (in Nightly builds, run with -Z macro-backtrace for more info)
error[E0308]: mismatched types
 --> example.rs:57:1
  |
57 | #[juniper::graphql_object(Context = Context)]
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected struct `std::vec::Vec`, found `&str`
  |
  = note: expected struct `std::vec::Vec<std::string::String>`
      found reference `&‘static str`
  = note: this error originates in an attribute macro (in Nightly builds, run with -Z macro-backtrace for more info)
error[E0277]: the trait bound `&_: juniper::FromInputValue` is not satisfied
  --> example.rs:57:1
  |
57 | #[juniper::graphql_object(Context = Context)]
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `juniper::FromInputValue` is not implemented for `&_`
  | 
[...]

Issue #814, opened 2020-12-03
graphql_interface and graphql_object does not mix well


skerkour:
**Describe the bug**

A `struct` using `#[graphql_object]` (for complexe fields resolvers) and `#[graphql_interface]` does not generate a valid schema (`as_schema_language`)

**To Reproduce**
```rust
use juniper::{graphql_interface, graphql_object, graphql_interface};

#[graphql_interface(for = User)]
pub trait Namespace {
    fn id(&self) -> Option<uuid::Uuid>;
    fn created_at(&self) -> chrono::DateTime<chrono::Utc>;
    fn name(&self) -> String;
}


#[derive(Debug, Clone)]
pub struct User {
    id: uuid::Uuid,
    created_at: chrono::DateTime<chrono::Utc>,
    name: String,
    other: String,
}

#[graphql_object]
#[graphql(impl = NamespaceValue)]
impl User {
    pub fn username() -> String {
        panic!("not implemented"); // TODO
    }
}

#[graphql_interface]
impl Namespace for User {
    fn id(&self) -> Option<uuid::Uuid> {
        Some(self.id)
    }

    fn created_at(&self) -> chrono::DateTime<chrono::Utc> {
        self.created_at
    }

    fn name(&self) -> String {
        self.name.clone()
    }
}
```
generates

```graphql
interface Namespace {
  id: Uuid
  createdAt: DateTimeUtc!
  name: String!
}


type User {
  username: String!
}

"DateTime"
scalar DateTimeUtc
"Uuid"
scalar Uuid
```

**Expected behavior**
The expected generated schema is as follow
```graphql
type User implements Namespace {
  id: Uuid!
  createdAt: DateTimeUtc!
  name: String!
  username: String!
}
```

**Additional context**
branch: `master`
rev: 4ffd276a5b33ae4423f295675f977ed6cc7fb88a

skerkour:
Also,
When I try as follow:
```rust
#[graphql_object(
    impl = NamespaceValue,
)]
impl User {
    pub fn username() -> String {
        panic!("not implemented"); // TODO
    }
}

#[graphql_interface]
impl Namespace for User {
    fn id(&self) -> Option<uuid::Uuid> {
        Some(self.id)
    }

    fn created_at(&self) -> chrono::DateTime<chrono::Utc> {
        self.created_at
    }

    fn name(&self) -> String {
        self.name.clone()
    }
}

```

it generates the following code:
```graphql
type User implements Namespace {
  username: String!
[...]

PR #971, opened 2021-07-22
Rework codegen for GraphQL objects

### LINKED ISSUE #525: Allow argument documentation

I like to provide clear documentation to my end users about everything in the schema, and this includes arguments.

As of Rust 1.39, attributes are allowed on parameters. rust-lang/rfcs#2565.

I would like to see these used to provide documentation for arguments.

```rust
#[juniper::object(Context = Context)]
impl QueryRoot {
    #[doc = "Ability to document the endpoint already exists"]
    fn some_endpoint(
        context: &Context,
        #[doc = "Ability to document arguments would be awesome"]
        id: ID,
        #[doc = "Only include items in one of these states. Defaults to all states."]
        states: Option<Vec<StateInput>>
    ) -> SomeObject {
        todo!()
    }
}
```

We don't need to use `#[doc = ""]`. We could also use a custom attribute.

**Additional context**

Documenting arguments is common. For example, on the [GitHub v4 API](https://developer.github.com/v4/query/), everything considered a "Connection" has its arguments clearly documented.

#334 seems to suggest that this works. However, I can't seem to get it to work. Possibly it doesn't work in `#[juniper::object]` macros?

tyranron:
Part of #421  
Superseeds #441   
Fixes #525, #697, #814, #827, #875, #880

This PR migrates `#[derive(GraphQLObject)]`, `#[graphql_object]` and `#[graphql_subscription]` macro implementations to the same machinery used by `#[graphql_interface]`.

Additionally, it changes codegen in the way, that computed fields are not desugared into async blocks, but rather are preserved as methods and called directly.

## Checklist
- [ ] Implemented
- [ ] Old code is cleaned
- [ ] Covered with integration tests
- [ ] Covered with codegen failure tests
- [ ] Documantation is updated
- [ ] Changelog entry added

### DIFF
--- a/juniper_codegen/src/lib.rs
+++ b/juniper_codegen/src/lib.rs
@@ -143,6 +143,7 @@ pub fn derive_input_object(input: TokenStream) -> TokenStream {
     }
 }
 
+// TODO
 #[proc_macro_error]
 #[proc_macro_derive(GraphQLObject, attributes(graphql))]
 pub fn derive_object(input: TokenStream) -> TokenStream {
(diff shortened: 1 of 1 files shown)


tyranron:
@LegNeato this one is too huge, so I merge it without requesting your review. It's quite a breaking change, so, please, don't release `master` yet. There will be subsequent PRs with more polishing, reworking and fixing before releasing `0.16`.

LegNeato:
Sounds good, thank you for the great work!

Issue #953, opened 2021-07-04
Performance Tracing Extension


mathroc:
**Is your feature request related to a problem? Please describe.**

I’d like to be able to get performance feedback from my queries in case I want to identify which resolver is causing slow queries

**Describe the solution you'd like**

Apollo has defined a response format for a `tracing` extension that’s already implemented by multiple GraphQL servers and at least by apollo engine and graphql playground

**Describe alternatives you've considered**

telemetry can be useful too (#423, #713) when it’s complete enough, but requires additional tooling and is not available easily to the API users

**Additional context**

https://github.com/apollographql/apollo-tracing

LegNeato:
Would love this behind a feature flag as it is non-standard.

PR #720, opened 2020-07-25
Add tracing support

mihai-dinculescu:
This is a stab at #713.

PR #972, opened 2021-07-23
Add tracing support for GraphQL objects, interfaces and subscriptions

ArsileLuci:
This is upgraded version of #720 
And part of #953 

This PR adds tracing support for GraphQL objects, interfaces and subscriptions.

- Add `trace` argument to `#[graphql_object]`, `#[graphql_interface]` attributes and top-level `#[graphql]` for derived GraphQL objects.
- Add `#[tracing]` attribute on GraphQL fields, that can be used to tune how each field is traced
- Impl tests for code related to tracing

- [ ] Ensure proper formatting
- [ ] Run all tests
- [ ] Update the CHANGELOG


ArsileLuci:
cc @tyranron

juniper/src/lib.rs
```diff
@@ -275,30 +301,43 @@ where
...
         }
     }
 
-    executor::execute_validated_query_async(&document, operation, root_node, variables, context)
-        .await
+    let f = executor::execute_validated_query_async(
+        &document, operation, root_node, variables, context,
+    );
+
+    __juniper_instrument_trace!(f, "execute_validated_query").await
```
tyranron, line 340: @ArsileLuci seems that shoud be:
```rust
__juniper_instrument_trace!(f, "execute_validated_query_async").await
```

juniper/src/util.rs
```diff
@@ -51,3 +51,22 @@ fn test_to_camel_case() {
     assert_eq!(&to_camel_case("a")[..], "a");
     assert_eq!(&to_camel_case("")[..], "");
 }
+
+#[cfg(feature = "tracing-futures")]
```
tyranron, line 55: @ArsileLuci why `"tracing-futures"`? It will complicate looking for feature-gated stuff. It's better to gate everything under the single `"tracing"` feature, unless these two are going to be used separately, which is not our case.

juniper/src/util.rs
```diff
@@ -51,3 +51,22 @@ fn test_to_camel_case() {
...
+#[doc(hidden)]
+pub mod tracing {
+    use tracing::Span;
+    use tracing_futures::{Instrument, Instrumented};
+
+    /// Required to pass sanity tests when `Instrument` already imported.
+    pub trait InstrumentInternal {
+        fn __instrument(self, span: Span) -> Instrumented<Self>
```
tyranron, line 63: @ArsileLuci still, seems to be quite redundant.

juniper_codegen/src/util/tracing.rs
```diff
@@ -0,0 +1,352 @@
...
+        );
+    )
+}
+
+// Returns code to start tracing of async future
+pub fn async_tokens(ty: &impl TracedType, field: &impl TracedField) -> TokenStream {
+    if !is_traced(ty, field) {
+        return quote!();
+    }
+    quote!(.__instrument(_tracing_span))
```
tyranron, line 287: @ArsileLuci yup, as expected. Use universal trait method syntax call, and you won't need that custom trait. Like that:
```rust
let f = ::juniper::tracing::Instrument(f, _tracing_span);
```

juniper_codegen/src/util/mod.rs
```diff
@@ -1291,7 +1366,7 @@ impl GraphQLTypeDefiniton {
                                         Err(e) => Err(ex.new_error(e)),
                                     }
                                 }
-                            });
+                            }) #trace_async;
```
tyranron, line 1369: and here it will be like
```rust
});
#trace_async
```

juniper_codegen/src/util/mod.rs
```diff
@@ -4,6 +4,9 @@ pub mod duplicate;
 pub mod parse_impl;
 pub mod span_container;
 
+#[cfg(feature = "tracing")]
+pub mod tracing;
```
tyranron, line 8: @ArsileLuci it's better to move it from `util` module to crate root and have a straight-forward `tracing` module.

juniper_codegen/src/util/mod.rs
```diff
@@ -680,6 +706,9 @@ pub struct GraphQLTypeDefinitionField {
     pub is_async: bool,
     pub default: Option<TokenStream>,
     pub span: Span,
+
+    #[cfg(feature = "tracing")]
+    pub tracing: Option<tracing::Attr>,
```
tyranron, line 711: @ArsileLuci also, as `#[tracing]` represent a separate attribute, it woud be nice to not embed it here, but rather parse separately and use directly.

juniper/src/tests/fixtures/tracing/schema.rs
```diff
@@ -0,0 +1,551 @@
...
+    #[tracing(complex, no_trace)]
+    fn no_trace_complex(&self) -> i32 {
+        4
+    }
+}
+
+/// Derived GraphQL object marked with `trace = "complex"`.
+#[derive(GraphQLObject)]
+#[graphql(trace = "complex")]
+pub struct DerivedComplex {
```
tyranron, line 472: @ArsileLuci I think we can use here `#[tracing(complex)]` too.

Also, let's check whether we can do that for `#[graphql_object]` macro, by looking up this attribute on a definition, parsing it and removing it. If not, it would be better to keep the similar to other cases format:
```rust
#[graphql_object(tracing(complex))]
```

This should also simplify the parsing, as we would have similar patterns everywhere.

juniper_codegen/src/util/tracing.rs
```diff
@@ -0,0 +1,352 @@
...
+    token,
+};
+
+#[derive(Debug, Default)]
+pub struct Attr {
+    /// Optional span rename, if `None` method name should be used instead.
+    name: Option<syn::LitStr>,
+
+    /// Overwritten `level` of span generated, if `None` `Level::INFO` should be used.
+    level: Option<syn::LitStr>,
```
tyranron, line 18: @ArsileLuci also, why parse all the attribute arguments by ourselves? Cannot we piggyback everything to upstream (`tracing` crate)?

tyranron (changes requested):
@ArsileLuci thanks, quite raw and rough at the moment, needs more design polishing and some reconsiderations.

juniper_codegen/src/util/tracing.rs
ArsileLuci: @tyranron This becomes rather complicated once we introduce streams, to trace subscriptions, and in my opinion it' much easier to slap extension trait where it needs and keep it as simple as possible.

juniper_codegen/src/util/tracing.rs
ArsileLuci: @tyranron [`span!(...)`](https://docs.rs/tracing/0.1.26/tracing/macro.span.html) macro is pretty strict on argument order, and we need to cleanup arguments specific for `juniper`, so a little of extra work here will allow to forget about argument order.

juniper_codegen/src/util/mod.rs
ArsileLuci: @tyranron 

> parse separately and use directly.

You mean instead of expanding it within `#[graphql_object]`/`#[graphql_interface]` etc, we could just expand it separately for each field?

tyranron:
@ArsileLuci #971 is merged.

docs/book/content/tracing/index.md
```diff
@@ -0,0 +1,296 @@
...
+        // Some query
+# unimplemented!()
+    }
+}
+```
+
+In example above both `filter` and `count` of `products` field will be recorded
+in produced [`Span`]. All fields will be recorded using their `fmt::Debug`
+implementation, if your field doesn't implement `fmt::Debug` you'll get compile
+time error. In this case ypu should either implement `fmt::Debug` or skip it
```
mathroc, line 219:
```suggestion
time error. In this case you should either implement `fmt::Debug` or skip it
```

ArsileLuci: @mathroc Thanks for Your help, I'll fix this with next commit.

ArsileLuci:
@tyranron I want to discuss one feature in `tracing` that is currently risky to implement in Juniper. This is so called [`Empty` fields](https://docs.rs/tracing/0.1.26/tracing/field/struct.Empty.html), due to how we resolve fields we cannot use it in actual resolver (`fn` that's called within `resolve_field` and `resolve_field_async` to get value), we're limited to context of `resolve_field` or `resolve_field_async`.

It could be implemented by dirty modification of `fn`s signature, but it's a little bit unfair, I guess.

mathroc:
Hi!

Once this lands, I'd like to work on https://github.com/graphql-rust/juniper/issues/953 and hope to be able to leverage this. My idea is to create a tracing collector and capture relevant span data to later be able to output the expected JSON. I'm wondering if there is enough already or if some tracing could be added (now or later):

```
    "tracing": {
      "version": 1,
      "startTime": <>, → that would be the "enter execute"
      "endTime": <>, → "close execute"
      "duration": <>, → the duration of the "execute" span
      "parsing": { → I don't think this is present at the moment
        "startOffset": <>, 
        "duration": <>,
      },
      "validation": {  -> there's "validate_document" & "validate_input_values" but I don't know if a duration can be extracted from simple spans. Should they be wrapped by a "start/close validation" span ?
        "startOffset": <>,
        "duration": <>,
      },
      "execution": {
        "resolvers": [ → They are traced. I would need to check if, 1: they can be identified easily, 2: everything below can be extracted from the span or could be added later
          {
            "path": [<>, ...],
            "parentType": <>,
            "fieldName": <>,
            "returnType": <>,
            "startOffset": <>,
            "duration": <>,
          },
          ...
        ]
      }
```

Do you think something like that would be possible?

LegNeato:
What outstanding issues are we waiting on here?

ArsileLuci:
Hi @mathroc 

About this one: 
> "validation": {  -> there's "validate_document" & "validate_input_values" but I don't know if a duration can be extracted from simple spans. Should they be wrapped by a "start/close validation"

Everything related to duration/time should be implemented on `Subscriber`(https://docs.rs/tracing/0.1.29/tracing/subscriber/trait.Subscriber.html#) side.

> 1: they can be identified easily, 2: everything below can be extracted from the span or could be added later

1) It's not so simple, probably the only way to identify whether this is field resolver or not is to add custom marker field in every resolver, something like `juniper-resolver=true`.

2) Let's walk through every field and check what we can do:
- path: This will involve some work but all resolver spans follow one naming schema `<TypeName>.<resolverName>`. Combining it with custom marker field, mentioned above, you can walkthrough every single span and build resolver chain.
- parentType: If I understand it right this is `<TypeName>` in span name so there should be no problems with it.
- fieldName: Same to `parentType` but this is `<resolverName>`
- returnType: This should be implemented, probably using late binding fields
- startOffset and duration: also should be implemented manually

The biggest disadvantage I see so far is making a whole lot of custom fields. In process of working on this PR I've tested this with stack `tracing -> tracing_opentelemetry -> opentelemetry_jaeger` and all custom fields were sent to Jaeger. This resulted in Jaeger denying nearly half of my traces because "they are too large".

mathroc:
thx @ArsileLuci for the feedback. Would you recommend doing something completely separate instead?

ArsileLuci:
@mathroc 

> Would you recommend doing something completely separate instead?

If you need this exact feature, it could be implemented under feature flag either disabling or enabling it. The issue with Jaeger, could be simply solved, by just disabling this extra fields. Just explicitly mention it in the docs, and this should be enough.

@LegNeato 

> What outstanding issues are we waiting on here?

I don't think there's any, except this branch being a bit old (and forgotten).

tyranron:
@LegNeato this branch requires some serious review. We (me, @ArsileLuci and @ilslv) have plans to revive it after merging #975

After this PR we will land some subsequent ones, hopefully to fully close the question about tracing in `juniper`.

PR #981, opened 2021-09-18
Mutable async context example

esseswann:


docs/book/content/types/objects/using_contexts.md
```diff
@@ -87,4 +86,55 @@ impl User {
...
+If you are using async runtime like `tokio` for mutable references you will need to use a corresponding async version of RwLock:
+```rust
+# extern crate juniper;
+# use std::collections::HashMap;
+# use juniper::graphql_object;
+# use tokio::sync::RwLock;
+#
+// This struct represents our context.
+struct Database {
+    requested_count: HashMap<i32, i32>,
```
kpp, line 99: Why HashMap?

esseswann: It was cloned from previous example, what do you think is more appropriate here?

kpp: I was mislead by `.entry(0)`.

tyranron (changes requested):
@esseswann thanks 🍻

But CI errors seem to be legit. Please, update the provided example so it passes the Book tests.

LegNeato:
Thank you!

PR #987, opened 2021-10-15
Allow post-validation `FromInputValue` errors

tyranron:
Required for #975

## Synopsis

At the moment, `FromInputValue` fallibility is considered during input operation validation only. Also, the real error is erased, due to `Option<T>` usage, thus argument absence is unobviously mixed with convertion errors over the code base.

Being applicable to most cases, still there are some situations where the error may occur only during resolution after passing the validation (from #975 it's deserializing `json::Value` into `Json<T: Deserialize>`). At the moment, we're ending up with runtime panics in such situation, which is not good.

## Solution

Make `FromInputValue` returning `Result` with the actual error preservation. Require this error be `IntoFieldError` convertible, so it may naturally bubble-up during operation execution.

## Checklist

- [x] tests renewed
- [x] documentation updated
- [x] CHANGELOG entry described


tyranron:
cc @ilslv 

Before continuing this, make sure `master` is built OK on latest nightly, and update it (if required) in a separate PR.

ilslv:
@tyranron PR is ready for review

juniper_codegen/src/graphql_union/mod.rs
```diff
@@ -694,9 +700,16 @@ impl VariantDefinition {
...
-                return #resolving_code;
+            match <#ty as ::juniper::GraphQLType<#scalar>>::name(info) {
+                Some(name) => {
+                    if type_name == name {
+                        let fut = ::juniper::futures::future::ready({ #expr });
+                        return #resolving_code;
+                    }
+                }
+                None => return Box::pin(::juniper::futures::future::ready(
+                    Err(::juniper::FieldError::from("This GraphQLType has no name"))
```
tyranron, line 711: @ilslv to DRYing the generated code a bit, let's move such things to helper functions inside `juniper` crate, and reuse them here.

tyranron (commented):
@ilslv I've refactored base machinery in `juniper`:
- more errors propagation;
- simplifying code with using `String` in implementations rather than `FieldError<S>`.

Codegen needs some updates, though, described below.

ilslv:
@tyranron this PR is ready for review

juniper_codegen/src/common/field/arg.rs
```diff
@@ -351,14 +351,69 @@ impl OnMethod {
...
+                                            let e = ::juniper::IntoFieldError::into_field_error(e);
+                                            ::juniper::FieldError::new(
+                                                format!(#err_text, e.message()),
+                                                e.extensions().clone(),
+                                            )
+                                        })
+                                },
+                                Ok,
+                            )
+                        })?
```
tyranron, line 372: @ilslv  I don't follow the logic here.

It was:
1. `args.get::<#ty>(#name)` returns `Option<T>`.
2. If it's `None`, then `::juniper::FromInputValue::<#scalar>::from_implicit_null` returns `Option<T>`.
3. If it's `None` we panic.

Now it is:
1.  `args.get::<#ty>(#name)` returns `Result<Option<T>, FieldError>`.
2. If it's `Ok`, we check the inner `opt`.
3. If it `Some` we map it to `Some(Ok(v))` and so have `Result<Option<Result<T>, FieldError>, FieldError>`.

Why?

It seems the more natural way is:
1. Unpack the `Result` early: `args.get::<#ty>(#name)?`
2. `.transpose()` the inner error in `Option` and return it with `?`.

tyranron: Oh... understood. `.map_or_else` doesn't map an inner `Option` value, but rather maps the whole `Option` into something.

`.ok_or_else` will do better here.

juniper_codegen/src/common/field/arg.rs
```diff
@@ -351,14 +351,69 @@ impl OnMethod {
...
+                        .and_then(|opt| {
+                            opt.map_or_else(
+                                || {
+                                    <#ty as ::juniper::FromInputValue<#scalar>>::
+                                        from_implicit_null()
+                                        .map_err(|e| {
+                                            let e = ::juniper::IntoFieldError::into_field_error(e);
+                                            ::juniper::FieldError::new(
+                                                format!(#err_text, e.message()),
+                                                e.extensions().clone(),
```
tyranron: This mess is better be replaced with `.map_message` method on `FieldError`.

juniper_codegen/src/common/field/arg.rs
tyranron: Meh... `.ok_or_else()` cannot map absense to `Result::Ok`.

tyranron (commented):
@ilslv LGTM

PR #990, opened 2021-11-03
Fix broken links in README.md

oliverlevay:


tyranron (changes requested):
@oliverlevay thanks for bringing attention to this, but removing link to latest stable version Book is not a solution, because we do intend to have them both.

The links will be fixed with the next release.

PR #996, opened 2021-11-23
Macro for `InputValue` creation

### LINKED ISSUE #503: Support experssions when using `juniper::graphql_value`
Say I need to convert a Custom Error into `FieldError` with extensions, I will do:

```rust
    #[derive(thiserror::Error, Debug, Deserialize, Serialize)]
    #[serde(tag = "type", content = "data")]
    enum SomeError {
        #[error("SomeError Item: arg1: {arg1}, arg2: {arg2}")]
        Item { arg1: String, arg2: i32 },
    }

    impl IntoFieldError for SomeError {
        fn into_field_error(self) -> juniper::FieldError {
            juniper::FieldError::new(
                "Get Some Error",
                juniper::graphql_value!({
                    "type": "SomeError",
                    "arg1": &self.arg1,
                    "arg2": &self.arg2,
                }),
            )
        }
    }
```

However, this will not compile, maybe because the macro does not support varients...

One walkround is:

```rust
        juniper::FieldError::new(
           "Get Some Error",
            juniper::Value::object(
                vec![
                    ("type","SomeError"),
                    ("arg1", juniper::Value::scalar(&self.arg1)),
                    ("arg2", juniper::Value::scalar(&self.arg2)),
                ]
                .into_iter()
                .collect(),
            ),
        )
```

But of course, it's not as clear as the former one...

tyranron:
Resolves #503

## Synopsis

We have a `graphql_value!()` macro which allows `Value`s creation in a convenient way.

We don't have similar for `InputValue`s and `Variables`, however. Which makes test cases unergonomic, harder to write and read.

## Solution

Introduce `graphql_input_value!()` and `graphql_vars!()` macros for that purpose.

## Additionally

While `graphql_value!()` exists, it has some inconveniences at the moment:
- It accepts `None` for `null` values. It would be better to follow `serde_json::json!()` macro here and accept `null`s too instead of `None`.
- It doesn't play well with accepting expressions. In some cases it just fails.

To cover those needs, it is worth to look at how [`serde_json::json!()` is implemented](https://docs.rs/serde_json/1.0.71/src/serde_json/macros.rs.html#53-280) and repeat the simplified version for our needs, both for `graphql_input_value!()` and `graphql_value!()` macros.


tyranron:
@ilslv it would be better also to move macros definitions into their own modules.

ilslv:
@tyranron `graphql_input_value!` should be able to encode `InputValue::Variable` and `InputValue::Enum`. GraphQL uses `$` prefix to indicate variable: [example]. But rust declarative macros can't match `$`, so we could possibly use something like `#` or `@`, but it will be hard to remember and it doesn't solve `InputValue::Enum`. So using `let` and `enum` prefixes makes most sense to me:

```rust
graphql_input_value!({
    "key": 123,
    "var": let variable,
    "enum": enum ENUM,
    "arr": [123, let another, enum OTHER]
});
```




[example]: https://spec.graphql.org/June2018/#example-704b8

tyranron:
@ilslv it's seems that for enums there should be just an `ident` usage, if we want to [comply with GraphQL syntax](https://spec.graphql.org/October2021/#EnumValue).

So, we have the following options here:
1. Either use the variant you've proposed, which is unintuitive and compliant with GraphQL syntax, but it's `macro_rules` so good IDE assist.
2. Or implement via proc macro and support syntax for variables as `$var` and enum values as `UPPERCASED_IDENT`, but then loose IDE assist.
3. Or throw support for variables and enums and don't bother at all.
4. Or throw support for variables only while `stringify!(ident)` should work for enum values, so we still may use `macro_rules` for that, and preserve IDE assist.

Considering that this feature is primarily for internal usage, what do you prefer?

ilslv:
@tyranron great idea with `ident`, completely missed that. 
I think we should stick with declarative macros, because of IDE support and much simpler implementation. It may be less robust, but I don't think it has to be, considering usage. 
I lean towards `stringify!(ident)` for enums and `#var` for variables. It gives us most amount of flexibility and if this macro will become more popular than we expect, we'll be able to rewrite it into proc macro with `$var` and least amount of breaking changes.

tyranron:
@ilslv ok, let's do this way then. 

> `#var` for variables

Maybe `@var` then? `#` is usually associated with comments or meta.

ilslv:
## FCM

```
Implement `graphql_input_value!` and `graphql_vars!` macros (#996, #503)

- add `From` impls to `InputValue`, that mirror `Value` impls to provide better support for `Option` handling
- support expressions in `graphql_value!` macro
- use `null` in addition to `None` to create `Value::Null` in `graphql_value!` macro to mirror `serde_json::json!`
```

@tyranron this PR is ready for review. I've left support for creating `Value::Null` with `None` because this feature is part of support for better handling of `Option`s:
```rust
#[test]
fn option_support() {
    let val = Some(42);
    assert_eq!(graphql_value!(None), V::Null);
    assert_eq!(graphql_value!(null), V::Null);
    assert_eq!(graphql_value!(Some(42)), V::scalar(42));
    assert_eq!(graphql_value!(val), V::scalar(42));
}
```

tyranron (commented):
@ilslv despite the described nitpickts, you've done a brilliant work with macros implementation. Thanks!

PR #1002, opened 2021-12-08
Update book links to fix 404

talhaguy:
A few of the book links were 404'ing and this PR updates them. Please let me know if the updates I have made are not intended.


tyranron (approved):
@talhaguy thanks! 🍻

PR #1006, opened 2021-12-15
feat: Add support for the time crate

scott-wilson:
This adds support for the `time` crate due to concerns with safety in the `chrono` crate (see https://github.com/rustsec/advisory-db/pull/1082).

So, I added support for time, since it looks like that might become the "preferred" date/time crate in the future.

PR #1010, opened 2021-12-23
Update `chrono` scalars according to the `graphql-scalars.dev`

ilslv:
Revealed from https://github.com/graphql-rust/juniper/pull/1006#pullrequestreview-834160552

## Synopsis

For now scalars from `time` and `chrono` crates are different.

## Solution

Update `chrono` scalars according to the [`graphql-scalars.dev`](https://graphql-scalars.dev)

## Checklist

- Created PR:
    - [x] In [draft mode][l:1]
    - [x] Has assignee
    - [ ] and approved
- Before merge:
    - [x] Milestone is set
    - [x] PR's name and description are correct and up-to-date
    - [x] `Draft: ` prefix is removed
    - [x] All temporary labels are removed


ilslv:
## FCM

```
Update `chrono` scalars according to the `graphql-scalars.dev` (#1010)

- remove `scalar-naivetime` feature
```

juniper/src/integrations/chrono.rs
```diff
@@ -78,54 +146,82 @@ where
...
+                   See also [`chrono::DateTime`][2]`<`[`FixedOffset`][3]`>` \
+                   for details.\
+                   \n\n\
+                   [0]: https://datatracker.ietf.org/doc/html/rfc3339#section-5.6\n\
+                   [1]: https://graphql-scalars.dev/docs/scalars/date-time\n\
+                   [2]: https://docs.rs/chrono/*/chrono/struct.DateTime.html\n\",
+                   [3]: https://docs.rs/chrono/*/chrono/offset/struct.FixedOffset.html",
+    specified_by_url = "https://graphql-scalars.dev/docs/scalars/date-time"
+)]
+impl<S> GraphQLScalar for DateTime<FixedOffset>
```
tyranron, line 212: What we do really want here is
```rust
impl<S: ScalarValue, Tz: TimeZone> GraphQLScalar for DateTime<Tz>
```
but `#[graphql_scalar]` macro doesn't allows us to make generic impls.

So, let's first rewrite `#[graphql_scalar]` so we can do that.

Another problem arosen here is that `chrono` is very bad at relations between `TimeZone`s. We have no relations like (`From`) between `Utc`, `FixedOffset` and `Local`, and the only thing how we can create `Tz: TimeZone` is `TimeZone::from_offset()`, where `Offset` is another generic parameter.

So we should write up some `FromUtcOffset` (or similar) machinery here, to wire them up, so, after parsing `DateTime<FixedOffset>` we can convert it into arbitrary `DateTime<Tz>`.

docs/book/content/types/scalars.md
```diff
@@ -35,7 +35,7 @@ Juniper has built-in support for a few additional types from common third party
 crates. They are enabled via features that are on by default.
 
 * uuid::Uuid
-* chrono::DateTime
+* chrono::{DateTime, NaiveDate, NaiveDateTime, NaiveTime}
```
tyranron, line 38: I think we should update `chrono-tz` integration as well inside this PR.

tyranron (changes requested):

juniper/src/integrations/chrono.rs
ilslv: @tyranron 

> So we should write up some `FromUtcOffset` (or similar) machinery here, to wire them up, so, after parsing `DateTime<FixedOffset>` we can convert it into arbitrary `DateTime<Tz>`.

All in all the main problem is that to convert into arbitrary `DateTime<Tz>` you need an instance of a type, not just type itself. So even with custom machinery we can support only `FixedOffset` and `Utc`. [`chrono-tz`][1] on the other hand, uses big enum instead of separate types to represent all timezones, so our users will have to implement needed timezones manually.

I propose different approach: we allow only `DateTime<Utc>` and `DateTime<FixedOffsets>` as inputs, while any `DateTime<Tz>` as output. So usage will look something like that:

```rust
#[graphql_object]
impl QueryRoot {
    fn transform(input: DateTime<Utc>) -> DateTime<chrono_tz::Tz> {
        input.with_timezone(&chrono_tz::CET)
    }
}
```

The only downside is that we won't be able to use macros to implement scalar and fallback to manual implementation. This happens because with don't have a mechanism to enforce different trait bounds on different resolvers.




[1]: https://docs.rs/chrono-tz/0.6.1/chrono_tz/index.html

tyranron (approved):

PR #1013, opened 2022-01-01
Update versions referenced in server integration docs

DigitalZebra:
Simple update to latest versions of `juniper` and `juniper_$SERVER` versions referenced in the server integration docs, as current versions are stale.


tyranron (approved):
@DigitalZebra thanks! 🍻

PR #1014, opened 2022-01-10
Redesign `#[graphql_scalar]` macro

ilslv:
Part of #1010

## Synopsis

Main goal of this PR is to support generic scalars.

## Solution

In addition to this PR tries to unify `#[graphql_scalar]` and `#[derive(GraphQLScalar)]` macros. This is done by introducing ability to use custom resolvers: 

```rust
#[derive(GraphQLScalar)]
#[graphql(to_output_with = Self::to_output)]
struct Scalar(i32);

impl Scalar {
    fn to_output(&self) -> Value {
        Value::scalar(self.0 + 1)
    }
}
```

But we can't remove `#[graphql_scalar]` macro entirely because of uncovered use-case of derive macro: types from other crate

```rust
#[graphql_scalar(
    with = object_id,
    parse_token = String,
    scalar = LocalScalarValue,
)]
type ObjectId = bson::oid::ObjectId;

mod object_id {
    use super::*;

    pub(super) type Error = String;

    pub(super) fn to_output(v: &ObjectId) -> Value<LocalScalarValue> {
        Value::scalar(v.to_hex())
    }

    pub(super) fn from_input(v: &InputValue<LocalScalarValue>) -> Result<ObjectId, Error> {
        v.as_string_value()
            .ok_or_else(|| format!("Expected `String`, found: {}", v))
            .and_then(|s| {
                ObjectId::parse_str(s).map_err(|e| format!("Failed to parse `ObjectId`: {}", e))
            })
    }
}
```

## Checklist

- Created PR:
    - [x] In [draft mode][l:1]
    - [x] Has assignee
    - [ ] and approved
- Before merge:
    - [ ] Milestone is set
    - [ ] PR's name and description are correct and up-to-date
    - [ ] `Draft: ` prefix is removed
    - [ ] All temporary labels are removed


ilslv:
@tyranron currently scalar definitions looks like:

```rust
#[graphql_scalar(name = "DateTimeFixedOffset")]
impl<S> GraphQLScalar for DateTime<FixedOffset> 
where 
    S: ScalarValue
{
    fn resolve(&self) -> Value {
        Value::scalar(self.to_rfc3339())
    }

    fn from_input_value(v: &InputValue) -> Result<DateTime<FixedOffset>, String> {
        v.as_string_value()
            .ok_or_else(|| format!("Expected `String`, found: {}", v))
            .and_then(|s| {
                DateTime::parse_from_rfc3339(s)
                    .map_err(|e| format!("Failed to parse `DateTimeFixedOffset`: {}", e))
            })
    }

    fn from_str<'a>(value: ScalarToken<'a>) -> ParseScalarResult<'a, S> {
        if let ScalarToken::String(value) = value {
            Ok(S::from(value.to_owned()))
        } else {
            Err(ParseError::UnexpectedToken(Token::Scalar(value)))
        }
    }
}
```

Problems I see with following design and would like to fix:
1. The only generic argument allowed is single `ScalarValue`
2. Proc macro searches for methods with names `from_input_value` and `from_str`, extracts impl blocks and method argument name, then re-expands them with argument type replaced `::juniper::InputValue` or `::juniper::ScalarValue`. That allows you to not import `InputValue` and `ScalarValue`, but may cause problems. For example lifetime on `from_str` has to be exact `'a`.

How I see the solution:

1. Actually declare GraphQLScalar trait
```rust
trait GraphQLScalar<S = DefaultScalarValue> {
    type Error;

    fn resolve(&self) -> Value<S>;

    fn from_input_value(v: &InputValue<S>) -> Result<Self, Self::Error>;

    fn from_str<'a>(value: ScalarToken<'a>) -> ParseScalarResult<'a, S>;
}
```
This would work much better with IDE autocompletion, because now I have no idea what fields are expected and what happens under the hood, which leads just to a lot of copy-paste action. The only downside I see is that you'll have to explicitly specify `Error` type, but we already will be releasing breaking changes with #987 and I even would argue this will make transition more obvious. 

Final solution will look like:
```rust
#[graphql_scalar(name = "DateTimeFixedOffset")]
impl<S> GraphQLScalar<S> for DateTime<FixedOffset> 
where 
    S: ScalarValue
{
    type Error = String;

    fn resolve(&self) -> Value<S> {
        Value::scalar(self.to_rfc3339())
    }

    fn from_input_value(v: &InputValue<S>) -> Result<Self, Self::Error> {
        v.as_string_value()
            .ok_or_else(|| format!("Expected `String`, found: {}", v))
            .and_then(|s| {
                DateTime::parse_from_rfc3339(s)
                    .map_err(|e| format!("Failed to parse `DateTimeFixedOffset`: {}", e))
            })
    }

    fn from_str<'a>(value: ScalarToken<'a>) -> ParseScalarResult<'a, S> {
        if let ScalarToken::String(value) = value {
            Ok(S::from(value.to_owned()))
        } else {
            Err(ParseError::UnexpectedToken(Token::Scalar(value)))
        }
    }
}
```

The only potential problem is that some users may forget to use `#[graphql_scalar]` attribute, so `GraphQLScalar` may be a child of a marker trait like `DeriveGraphqlScalarAttribute` to get somewhat understandable error in that case.

tyranron:
@ilslv I think we do want here stay closer to what `#[graphql_object]` does. No need to imitate a trait. Just go with a regular `impl` block. Another point that the current `#[graphql_scalar]` macro is too boilerplatish.

We may even stray from this design further, by uniting `#[derive(GraphQLScalarValue)]` and `#[graphql_scalar]` macros, but the problem is that we cannot place `#[derive(GraphQLScalar)]` on a foreign type. So the simplified `impl` block with optional methods seems to be a reasonable solution here:

How about something like this:
```rust
#[graphql_scalar(name = "DateTimeFixedOffset")]
impl DateTime<FixedOffset> { 
    // We specify `ScalarValue` as type parameter here, or just use
    // parametrized `Value<MyCustomScalarValue>` in return type.
    fn to_output<S: ScalarValue>(&self) -> Value<S> {
        Value::scalar(self.to_rfc3339())
    }

    // As these are just separate methods, we may have parsing abstracted over
    // `S: ScalarValue`, while resolving only available for `MyCustomScalarValue`.
    // This allows more flexibility in constraints, if someone need them, as follows
    // the granularity of the original traits.
    fn from_input<S: ScalarValue>(v: &InputValue<S>) -> Result<Self, String> {
        v.as_string_value()
            .ok_or_else(|| format!("Expected `String`, found: {}", v))
            .and_then(|s| {
                DateTime::parse_from_rfc3339(s)
                    .map_err(|e| format!("Failed to parse `DateTimeFixedOffset`: {}", e))
            })
    }

    // We may also thing about defaults for this method, or out-of-the-box implementations
    // specified via attribute, as this is very commonly repeated pattern.
    // Very-very rarely the library user needs custom logic here.
    fn parse_token<'a, S: ScalarValue>(value: ScalarToken<'a>) -> ParseScalarResult<'a, S> {
        if let ScalarToken::String(s) = value {
            Ok(S::from(s.to_owned()))
        } else {
            Err(ParseError::UnexpectedToken(Token::Scalar(value)))
        }
    }
}
```

```rust
#[graphql_scalar(name = "DateTimeFixedOffset")]
#[graphql_scalar(parse_token(String, Int))] // optional
impl DateTime<FixedOffset> { 
    fn to_output<S: ScalarValue>(&self) -> Value<S> {
        Value::scalar(self.to_rfc3339())
    }

    fn from_input(v: &InputValue<DefaultScalarValue>) -> Result<Self, String> {
        v.as_string_value()
            .ok_or_else(|| format!("Expected `String`, found: {}", v))
            .and_then(|s| {
                DateTime::parse_from_rfc3339(s)
                    .map_err(|e| format!("Failed to parse `DateTimeFixedOffset`: {}", e))
            })
    }
}
```

ilslv:
@tyranron maybe it's just me, but I found frustrating every time I need to find some other scalar definition to copy meaningful methods and then autocomplete is broken because of proc macro expansion. That's why I would like to see `GraphQLScalar` as a trait. Also error messages would be much more understandable. 

> ```rust
>     // We specify `ScalarValue` as type parameter here, or just use
>     // parametrized `Value<MyCustomScalarValue>` in return type.
>     fn to_output<S: ScalarValue>(&self) -> Value<S> {
>         Value::scalar(self.to_rfc3339())
>     }
> ```

We can move `ScalarValue` to methods instead of type parameter

>```rust
>    // As these are just separate methods, we may have parsing abstracted over
>    // `S: ScalarValue`, while resolving only available for `MyCustomScalarValue`.
>    // This allows more flexibility in constraints, if someone need them, as follows
>    // the granularity of the original traits.
>    fn from_input<S: ScalarValue>(v: &InputValue<S>) -> Result<Self, String> {
>        v.as_string_value()
>            .ok_or_else(|| format!("Expected `String`, found: {}", v))
>            .and_then(|s| {
>                DateTime::parse_from_rfc3339(s)
>                    .map_err(|e| format!("Failed to parse `DateTimeFixedOffset`: {}", e))
>            })
>    }
>```

I don't see an obvious benefit of this and only as unnecessary complexity. You'll need to know exact trait this would resolve to (which method name is different) to use this feature.  

>```rust
>    // We may also thing about defaults for this method, or out-of-the-box implementations
>    // specified via attribute, as this is very commonly repeated pattern.
>    // Very-very rarely the library user needs custom logic here.
>    fn parse_token<'a, S: ScalarValue>(value: ScalarToken<'a>) -> ParseScalarResult<'a, S> {
>        if let ScalarToken::String(s) = value {
>            Ok(S::from(s.to_owned()))
>        } else {
>            Err(ParseError::UnexpectedToken(Token::Scalar(value)))
>        }
>    }
>```

I like this idea and this could be achieved with `GraphQLScalar` trait. But also we could add methods like `parse_string/int/bool_tokens()` to make it more idiomatic.

tyranron:
@ilslv 

> maybe it's just me, but I found frustrating every time I need to find some other scalar definition to copy meaningful methods and then autocomplete is broken because of proc macro expansion.

Well... two things:
1. You don't need autocompletion on these methods, because you won't use them directly in the user code. They're rather consumed and used by the inner `juniper` machinery.
2. Even if you do want use this methods, you shouldn't have any problems with autocompoletion, as they're will be just regular methods... oh, right, we cannot declare regular methods on foreign types >_<

ilslv:
@tyranron 

> 1. You don't need autocompletion on these methods, because you won't use them directly in the user code. They're rather consumed and used by the inner juniper machinery.

With trait approach we can simply leave them inside user code, so they can simply reuse them without digging into what exactly happen (I guess this use-case is very rare)

> 2. Even if you do want use this methods, you shouldn't have any problems with autocompoletion, as they're will be just regular methods...

I never can remember exact names and arguments of those methods, but with trait it's enough to remember single name and autocompletion will help me. Also, once we start reassembling those methods into other traits, autocompletion freaks out. 

> oh, right, we cannot declare regular methods on foreign types >_<

Didn't see this coming too.

tyranron:
@ilslv what if:

```rust
#[graphql_scalar(name = "DateTimeFixedOffset")]
#[graphql_scalar(parse_token(String, Int))]
#[graphql_scalar(resolve_with = my_func)]
#[graphql_scalar(from_input_with = my_func2)]
type DateTimeFixedOffset = DateTime<FixedOffset>;

fn my_func<S: ScalarValue>(&dt: DateTimeFixedOffset) -> Value<S> {
    Value::scalar(dt.to_rfc3339())
}

fn my_func2<S: ScalarValue>(v: &InputValue<S>) -> Result<DateTimeFixedOffset, String> {
    v.as_string_value()
       .ok_or_else(|| format!("Expected `String`, found: {}", v))
       .and_then(|s| {
           DateTime::parse_from_rfc3339(s)
               .map_err(|e| format!("Failed to parse `DateTimeFixedOffset`: {}", e))
       })
}
```

This will work well with both aucompletion and foreign types, preserving the desired granularity.

Also this paves the way to unite `#[derive(GraphQLScalar)]` and `#[graphql_scalar]` macros.

tyranron:
```rust
#[derive(GraphQLScalar)]
#[graphql(name = "DateTimeFixedOffset")]
#[graphql(parse_token(String, Int))] // optional
#[graphql(resolve_with = Self::my_func)] // optional
#[graphql(from_input_with = Self::my_func2)] // optional
struct DateTimeFixedOffset(DateTime<FixedOffset>);

impl DateTimeFixedOffset {
    fn my_func<S: ScalarValue>(&self) -> Value<S> {
        Value::scalar(self.0.to_rfc3339())
    }

    fn my_func2<S: ScalarValue>(v: &InputValue<S>) -> Result<Self, String> {
        v.as_string_value()
           .ok_or_else(|| format!("Expected `String`, found: {}", v))
           .and_then(|s| {
               DateTime::parse_from_rfc3339(s)
                   .map_err(|e| format!("Failed to parse `DateTimeFixedOffset`: {}", e))
           })
           .map(Self)
    }
}
```

ilslv:
@tyranron 

>```rust
>#[derive(GraphQLScalar)]
>#[graphql(parse_token(String, Int))] // optional
>#[graphql(resolve_with = Self::my_func)] // optional
>#[graphql(from_input_with = Self::my_func2)] // optional
>struct DateTimeFixedOffset(DateTime<FixedOffset>);
>
>impl DateTimeFixedOffset {
>    fn my_func<S: ScalarValue>(&self) -> Value<S> {
>        Value::scalar(self.0.to_rfc3339())
>    }
>
>    fn my_func2<S: ScalarValue>(v: &InputValue<S>) -> Result<Self, String> {
>        v.as_string_value()
>           .ok_or_else(|| format!("Expected `String`, found: {}", v))
>           .and_then(|s| {
>               DateTime::parse_from_rfc3339(s)
>                   .map_err(|e| format!("Failed to parse `DateTimeFixedOffset`: {}", e))
>           })
>           .map(Self)
>    }
>}
>```

This approach will require wrapping everything into a newtype and specify generic bounds on struct itself, rather then on a trait impl. But most importantly, I don't know a good way to treat generic scalars with maybe some optional bounds.

```rust
#[derive(GraphQLScalar)]
#[graphql(scalar = S)] // Does this mean generic scalar or struct `S`?
#[graphql(parse_token(String, Int))] // optional
#[graphql(resolve_with = Self::my_func)] // optional
#[graphql(from_input_with = Self::my_func2)] // optional
struct DateTime<Tz: TimeZone>(chrono::DateTime<Tz>);
//             ^^^^^^^^^^^^^^ Bound will have to be here.

impl<Tz: TimeZone> DateTime<Tz> {
    fn my_func<S: ScalarValue>(&self) -> Value<S> {
        Value::scalar(self.0.to_rfc3339())
    }

    fn my_func2<S: ScalarValue>(v: &InputValue<S>) -> Result<Self, String> {
        v.as_string_value()
           .ok_or_else(|| format!("Expected `String`, found: {}", v))
           .and_then(|s| {
               DateTime::parse_from_rfc3339(s)
                   .map_err(|e| format!("Failed to parse `DateTimeFixedOffset`: {}", e))
           })
           .map(Self)
    }
}
```

tyranron:
@ilslv 

> This approach will require wrapping everything into a newtype.

No, the example was written as not "instead of", but rather as "in addition to", for powering exisiting `#[derive(GraphQLScalarValue)]` macro and show that they can share the implementation.

> specify generic bounds on struct itself, rather then on a trait impl.

I think this should be OK, regarding the fact how scalars are used. But we always may introduce another attribute argument for specifying exoic bounds.

> But most importantly, I don't know a good way to treat generic scalars with maybe some optional bounds.

> ```rust
> #[graphql(scalar = S)] // Does this mean generic scalar or struct `S`?
> ```

We have this solved in `#[graphql_object]` and old `#[graphql_inteface]` as far as I recall.

```rust
#[graphql(scalar = S)] // `S` is a type
#[graphql(scalar = S: ScalarValue)] // `S` is a type parameter
```

tyranron:
With some trait-based introspection built on type-level we may event try to infer scalar type (or type parameter) from functions, as they have stable signatures.

ilslv:
@tyranron 

> We have this solved in #[graphql_object] and old #[graphql_inteface] as far as I recall.
> ```rust
> #[graphql(scalar = S)] // `S` is a type
> #[graphql(scalar = S: ScalarValue)] // `S` is a type parameter
> ```

Actually this is a bit more complex, because to construct `scalar::Type` we additionally search inside generics, but all in all this should work. 

https://github.com/graphql-rust/juniper/blob/c866e091a40c69ee0af1ff8a798723852e5d70ad/juniper_codegen/src/common/scalar.rs#L152-L173

> With some trait-based introspection built on type-level we may event try to infer scalar type (or type parameter) from functions, as they have stable signatures.

I've tried everything I could imagine, but the main problem is that so far i can tell there is no way to get exact type of the `fn some_fn(...) -> ...` to call associated type on it.

tyranron:
@ilslv 

> I've tried everything I could imagine, but the main problem is that so far i can tell there is no way to get exact type of the `fn some_fn(...) -> ...` to call associated type on it.

Yeah, you're right. That's the fundamental limitation of havong no `typeof`, so we cannot get the type of some value. Even TAITs won't help here afaik.

ilslv:
## FCM

```
Redesign `#[graphql_scalar]` macro: (#1014, #1000)
- support generic scalars with introduction of actual `GraphQLScalar` trait
- add `Error` associated type to the `GraphQLScalar` trait
- rename `resolve` method to `to_output`
- rename `from_input_value` method to `from_input`
- rename `from_str` method to `parse_token`
```

tyranron (changes requested):
Discussed. First, we're going to introduce a `#[derive(GraphQLScalar)]` macro being capable of the same as the current `#[graphql_scalar]`. And only then we make a `#[graphql_scalar]` a shim over `#[derive(GraphQLScalar)]` to be used on type aliases.

ilslv:
@tyranron the problem with that approach, is [generic bounds on type aliases are planned to be promoted into an error](https://github.com/rust-lang/rust/issues/49441). This isn't a dealbreaker, as we can remove original type alias, as it's an attribute macro, but the question is should we enforce users to write potentially invalid rust code and make IDE yell at them?

tyranron:
@ilslv why do we need trait bounds at type alias definition at all?

ilslv:
@tyranron in case an original struct lacks them and we need to enforce them. Imagine, that `chrone::DateTime<Tz>` wouldn't have `: TimeZone` bound on a struct level. In that case this is especially critical, as this approach is intended to be used with foreign types.

tyranron:
@ilslv so, we need them on generated `impl`s, not the type alias itself.

Then let's allow them as far as Rust allows them, but also let's provide additional attribute argument allowing to specify any desired additional bounds like:
```rust
#[graphql_scalar(where(Tz: chrono::TimeZone + Debug,  chrono::DateTime<Tz>: fmt::Debug))]
type DateTime<Tz> = chrono::DateTime<Tz>;
```

juniper/src/executor_tests/introspection/mod.rs
```diff
@@ -22,20 +21,22 @@ enum Sample {
 
 struct Scalar(i32);
 
-#[graphql_scalar(name = "SampleScalar")]
-impl<S: ScalarValue> GraphQLScalar for Scalar {
-    fn resolve(&self) -> Value {
-        Value::scalar(self.0)
+#[graphql_scalar(with = sample_scalar, parse_token = i32)]
+type SampleScalar = Scalar;
```
tyranron, line 25: Let's allow only `parse_token(i32)` argument form. `parse_token = i32` is kinda confusing.

juniper/src/executor_tests/variables.rs
```diff
@@ -5,30 +5,38 @@ use crate::{
...
 };
 
 #[derive(Debug)]
 struct TestComplexScalar;
 
-#[graphql_scalar]
-impl<S: ScalarValue> GraphQLScalar for TestComplexScalar {
-    fn resolve(&self) -> Value {
+#[graphql_scalar(
+    name = "TestComplexScalar",
```
tyranron, line 17: Let's makr such places with `// TODO` comment to not forget restore it to `derive` variant in the next PR.

tyranron (changes requested):
@ilslv anything else seems well enough 👍

The part with `from_input_error = String` still quite bothers me, but I don't see how to remove it. The `with = module` pattern is quite neat. 👍

ilslv:
@tyranron I don't know, why I didn't thought of this earlier, but we can expand into something like

```rust
impl<S: ScalarValue> FromInputValue<S> for Scalar {
    type Error = FieldError<S>;

    fn from_input_value(v: &InputValue<S>) -> {
        forwarded(v).map_err(IntoFieldError::<S>::into_field_error)
    }
}
```

Just transform provided error into `FieldError` by the way.

juniper/src/executor_tests/variables.rs
```diff
@@ -5,30 +5,35 @@ use crate::{
...
+
+mod test_complex_scalar {
+    use super::*;
+
+    pub(super) fn to_output<S: ScalarValue>(_: &ComplexScalar) -> Value<S> {
         graphql_value!("SerializedValue")
     }
 
-    fn from_input_value(v: &InputValue) -> Result<TestComplexScalar, String> {
+    pub(super) fn from_input<S: ScalarValue>(v: &InputValue<S>) -> Result<ComplexScalar, String> {
```
tyranron, line 31: @ilslv with necessity of `type Error` being removed it's much more pleasant now 👍

Regarding this pattern, I wonder whether we can push it a little bit further in terms of ergonomics and have this:
```rust
#[graphql_scalar(
    name = "TestComplexScalar",
    with = Self, // or `with = ComplexScalar`, so it acts like module
    parse_token(String),
)]
type ComplexScalar = TestComplexScalar;

impl ComplexScalar {
    fn to_output<S: ScalarValue>(&self) -> Value<S> {
        graphql_value!("SerializedValue")
    }

    fn from_input<S: ScalarValue>(v: &InputValue<S>) -> Result<Self, String> {
        v.as_string_value()
            .filter(|s| *s == "SerializedValue")
            .map(|_| Self)
            .ok_or_else(|| format!(r#"Expected "SerializedValue" string, found: {}"#, v))
    }
}
```

We may date even to not require specifying `with =` attribute argument, implying this case as a default one. And leave `with` arguments only for overriding things.

See how nice it is:
```rust
#[graphql_scalar(
    name = "TestComplexScalar",
    parse_token(String),
)]
type ComplexScalar = TestComplexScalar;

impl TestComplexScalar {
    fn to_output<S: ScalarValue>(&self) -> Value<S> {
        graphql_value!("SerializedValue")
    }

    fn from_input<S: ScalarValue>(v: &InputValue<S>) -> Result<Self, String> {
        v.as_string_value()
            .filter(|s| *s == "SerializedValue")
            .map(|_| Self)
            .ok_or_else(|| format!(r#"Expected "SerializedValue" string, found: {}"#, v))
    }
}
```

Seems that nothing forbids this, while having it finaly feels "quite right" as reduces boilerplate amap.

tyranron (changes requested):

juniper/src/executor_tests/variables.rs
ilslv: @tyranron `with = Self` is a great idea and it even works on current branch, but we should definitely add tests for it.

I'm not sure about not specifying `with =`. 

First of all this makes errors very hard to interpret, as the is no way to check whether method exists with pretty error. Especially here I think it's important, as this is a most basic form of a macro, where we should guide users with well-formed error messages on how to achieve their goal, rather then yell at them with hard to understand compiler error.

Also let's remember the use-case for `#[graphql_scalar]` macro: implement scalar on foreign types. If you can impl on a type, why not use `#[derive(GraphQLScalar)]`? It now has all the capabilities of `#[graphql_scalar]`.
Also this syntax will be unique for the `#[graphql_scalar]`, as with `#[derive(GraphQLScalar)]` it will introduce ambiguity:

```rust
// Should this use field or Self?
#[derive(GraphQLScalar)]
struct ComplexScalar {
    field: String,
}

impl ComplexScalar {
    fn to_output<S: ScalarValue>(&self) -> Value<S> {
        graphql_value!("SerializedValue")
    }

    fn from_input<S: ScalarValue>(v: &InputValue<S>) -> Result<Self, String> {
        v.as_string_value()
            .filter(|s| *s == "SerializedValue")
            .map(|_| Self)
            .ok_or_else(|| format!(r#"Expected "SerializedValue" string, found: {}"#, v))
    }
}

// And what to use now?
#[derive(GraphQLScalar)]
struct EvenMoreComplexScalar {
    field: String,
    another_field: i32,
}

impl EvenMoreComplexScalar {
    fn to_output<S: ScalarValue>(&self) -> Value<S> {
        graphql_value!("SerializedValue")
    }

    fn from_input<S: ScalarValue>(v: &InputValue<S>) -> Result<Self, String> {
        v.as_string_value()
            .filter(|s| *s == "SerializedValue")
            .map(|_| Self)
            .ok_or_else(|| format!(r#"Expected "SerializedValue" string, found: {}"#, v))
    }
}
```

I think this is the case where we should be explicit rather then implicit and preserve unified syntax across `#[graphql_scalar]` and `#[derive(GraphQLScalar)]`. Realistically `#[graphql_scalar]` should be the last resort, because of it's goal and usage of it requiring local `ScalarValue` implementation.

tyranron: @ilslv there is no misconception here. You're looking onto `#[derive(GraphQLScalar)]` and `#[graphql_scalar]` macros as to something different. I dare to treat them _as the same thing_, so, eventually, there will be no difference if we write:
```rust
#[derive(GraphQLScalar)]
struct MyScalar;
```
or
```rust
#[graphql_scalar]
struct MyScalar;
```
The semantics will be preserved, and both macro should act like a singe one.

So, keeping in mind the described above...
> ```rust
> // Should this use field or Self?
> #[derive(GraphQLScalar)]
> struct ComplexScalar {
>     field: String,
> }
> ```
It should use `Self`.

> ```rust
> // And what to use now?
> #[derive(GraphQLScalar)]
> struct EvenMoreComplexScalar {
>     field: String,
>     another_field: i32,
> }
```

Again, `Self`. 

If we'd like to provide some sort of delefation to the inner scalar here, we should do that with additional attributes like that:
```rust
#[derive(GraphQLScalar)]
#[graphql(transparent)] // delegates everything to the inner field
struct ComplexScalar {  // only single-field structs are allowed
    field: String,
}
```
or even:
```rust
#[derive(GraphQLScalar)]
struct EvenMoreComplexScalar {
    #[graphql(delegate)] // delegates everything to this inner field
    field: String,
    another_field: i32,
}
```
But the default macro semantics shouldn't change depending on what it's applied to.

> First of all this makes errors very hard to interpret, as the is no way to check whether method exists with pretty error. Especially here I think it's important, as this is a most basic form of a macro, where we should guide users with well-formed error messages on how to achieve their goal, rather then yell at them with hard to understand compiler error.

Correct me if I'm wrong, but it seems that the default compiler error here won't be that bad, as it only will say about "No method `to_output` found for type `MyScalarValue`" or something similar. Actually, could you provide an example of such error?

> Also let's remember the use-case for #[graphql_scalar] macro: implement scalar on foreign types. 

And this point seems quite right here, as foreign types will unlikely have those methods defined. So I propose making `with = ` (or similar) arguments mandatory only when the macro is placed _onto a type alias_, and implying `with = Self` automatically for structs (despite what macro is used exactly).

ilslv: @tyranron 

> Actually, could you provide an example of such error?

```rust
struct CustomCounter(i32);

#[graphql_scalar]
type Counter = CustomCounter;

impl Counter {
    // missing impl
    // fn to_output<S: ScalarValue>(v: &Counter) -> Value<S> {
    //     Value::scalar(v.0)
    // }

    fn from_input<S: ScalarValue>(v: &InputValue<S>) -> Result<Counter, String> {
        v.as_int_value()
            .ok_or_else(|| format!("Expected `String`, found: {}", v))
            .map(CustomCounter)
    }

    fn parse_token<S: ScalarValue>(value: ScalarToken<'_>) -> ParseScalarResult<'_, S> {
        <i32 as ParseScalarValue<S>>::from_str(value)
    }
}
```

```
error[E0599]: no function or associated item named `to_output` found for struct `with_self::CustomCounter` in the current scope
   --> integration_tests/juniper_tests/src/codegen/impl_scalar.rs:424:5
    |
422 |     struct CustomCounter(i32);
    |     -------------------------- function or associated item `to_output` not found for this
423 | 
424 |     #[graphql_scalar]
    |     ^^^^^^^^^^^^^^^ function or associated item not found in `with_self::CustomCounter`
    |
    = note: this error originates in the attribute macro `graphql_scalar` (in Nightly builds, run with -Z macro-backtrace for more info)
```
> ```rust
> #[derive(GraphQLScalar)]
> struct EvenMoreComplexScalar {
>     #[graphql(delegate)] // delegates everything to this inner field
>     field: String,
>     another_field: i32,
> }
> ```

Ok, with introduction of `#[graphql(delegate)]`/`#[graphql(transparent)]` things are much clearer now. I do like this approach more, as usually just newtyping a struct isn't enough to call it a new scalar, you need to check some additional invariants.

> So I propose making `with =` (or similar) arguments mandatory only when the macro is placed onto a type alias, and implying `with = Self` automatically for structs (despite what macro is used exactly).

I do quite like this approach 👍

tyranron: @ilslv 

> Should this PR also implement `#[graphql_scalar]` on structs or only type aliases?

I'd opt for _yes_. And not only structs, but enums/unions too. Any type supported by `proc_macro_derive`.

> Should we tread this PR like "implementing `#[graphql_scalar]` entirely" or rather `implementing `#[graphql_scalar]` for type aliases` and leave `#[graphql_scalar]` and `#[derive(GraphQLScalar)]` on structs and enums for #1017?

The implementation doesn't differ much for type aliases and structs. Though we may leave `transparent` implementation part for #1017, and impl here only basics which are similar for both type aliases and structs.

PR #1021, opened 2022-01-27
Forward port

LegNeato:
Pulling the changes from the `security-0.15.8` branch into master

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -123,7 +123,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: install
-          args: --version=0.11.2 cargo-release
+          args: --version=0.19.4 cargo-release
 
       - name: Setup git
         run: |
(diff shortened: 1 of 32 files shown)


tyranron (commented):
@LegNeato maybe it's better to not duplicate the changes of `security-0.15.8`, but rather merge them "as is" resolving the necessary conflicts, and preserving original commits.

This way we won't have a forked histrory in the repo, and may preserve the appropriate tags on oifignal commits in `master`.

If you don't mind, I may roll the changes.

LegNeato:
Hm, I thought that is what I did!

tyranron:
@LegNeato I have really hard times with `cargo-make` and `cargo-release` every time I use them for this project. There are just to many caveats with them I hit into.

I'll disable release automation in `master` branch temporarily, and will rework CI and releasing in a separate PR soon, trying to reduce the role of these tools as much as possible.

PR #1023, opened 2022-02-10
Fixed docs/book build (missing #)

rimutaka:
mdbook refused to build/serve the full contents of the book due to the missing title. It should not be required and may be a recent regression in their code, but the book builds fine after I pit that # at the top.


rimutaka:
I filed a bug report with MDBOOK at https://github.com/rust-lang/mdBook/issues/1743. This PR won't be needed if that is resolved.

PR #1032, opened 2022-03-01
Add `arbitrary` support

LegNeato:
This adds optional support for https://github.com/rust-fuzz/arbitrary/ behind
the `arbitrary` feature flag (off by default).

This will be used in a future pull request to enable fuzzing using `cargo-fuzz`

### DIFF
--- a/juniper/Cargo.toml
+++ b/juniper/Cargo.toml
@@ -26,6 +26,7 @@ default = [
     "url",
     "uuid",
 ]
+arbitrary = ["dep:arbitrary", "smartstring/arbitrary"]
 expose-test-schema = ["anyhow", "serde_json"]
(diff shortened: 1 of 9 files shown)


LegNeato:
Note that I haven't added this to CI yet.

LegNeato:
Note that I had to manually implement rather than use the `Arbitrary` derive in many places due to lifetimes (see https://github.com/rust-fuzz/arbitrary/issues/104)

LegNeato:
This isn't 100% yet because the arbitrary `Name` implementation on `Schema` causes an error of `IncorrectFormat` all the time. That being said, this is a good start and allows everything to compile. I'm working on the custom arbitrary implementations that more closely match the expectations of the underlying data model.

PR #1054, opened 2022-04-09
Incorrect handling of types matched as idents

Cerber-Ursi:
The issue was originally discovered on [URLO](https://users.rust-lang.org/t/evaluation-macro-rules-arguments-before-procedural-macro/73681). Minimal example:
```rust
macro_rules! impl_id {
    ($typename:ident) => {
        #[juniper::graphql_object]
        impl $typename { // ERROR: could not determine ident for the `impl` type
            fn id(&self) -> i32 {
                42
            }
        }
    };
}

struct Unit;
impl_id!(Unit);
```
The reason for this error is the fact that, after being matched as `ident`, the identifier is not a bare ident anymore, it's a [`syn::TypeGroup`](https://docs.rs/syn/latest/syn/struct.TypeGroup.html), which is not handled correctly.

I've added the example above as test and modified the `TypeExt::unparenthesized` accordingly. It's entirely possible, however, that there are other places where either explicit match for `TypeGroup` or the call to `unparenthesized` should be added; I'm just not sure how to generate realistic test cases similar to the one above.

---

Side note: I wasn't able to run tests for this workspace, until I added features `chrono` and `chrono-tz` to the invocation. It's likely the consequence of #1010, and I'm not sure whether this is intentional (or desirable).


tyranron (approved):
@Cerber-Ursi thanks! 🍻

Very valuable addition. Didn't know about that nuance.

PR #1060, opened 2022-05-04
bigdecimal and rust_decimal optional support

Niedzwiedzw:


tyranron (approved):
@Niedzwiedzw thanks! 🍻

Issue #780, opened 2020-10-08
`GraphQLObject` field attribute like serde's `serialize_with`


DusterTheFirst:
**Is your feature request related to a problem? Please describe.**
When creating graphql objects using the `GraphQLObject` custom derive which have types that do not implement [`GraphQLType`](https://docs.rs/juniper/0.14.2/juniper/trait.GraphQLType.html) you end up being forced to use the `#[juniper::graphql_object]` proc macro and redefine methods to expose each and every field manually. It would be useful to have a `#[transform = "fn_name"]` field attribute that would allow users to transform the type into another ie. u64 to i32. 

This improvement would probably be a better solution to #553 if #769 gets implemented

DusterTheFirst:
Another solution could be to enable a way to make a type that implements `Display` implement `GraphQLType`

PR #1072, opened 2022-05-31
Rework core traits

tyranron:
Requires #1028, #1047, #1052  
Required for #975  
Related to #1091  
Eliminates #1097

This PR represents an attempt to make a fundamental rework of Juniper resolving machinery (`GraphQLType`, `GraphQLValue`, etc).

## Motivation

At the moment, the core traits like `GraphQLType` and `GraphQLValue` impose the following problems to codebase:
- Polution with trait bounds. Due to trait definitions providing default methods implementations, they are obligated to have additional bounds in their definitions, which inevitably polutes all the dependent code. This is also more compicated due to a single trait having multiple responsibilities, so in contexts where we require only some single responsibility, we still need to place all the bounds for other responsibilities too.
- No polymorphism for `Context` and `TypeInfo` types, as they're defined as associative types. So, once some leaf type in schema uses a different type for `Context`, we're fucked, and should either make the whole schema work with a single `Context` type, or doesn't compile at all. The same applies to `TypeInfo`, which, in result, makes impossible to combine static and dynamic GraphQL schemas at all.

## Solution

This PR tries to remodel the core traits machinery in the following ways:
- Make traits definitions granular: a separate trait for a separate need. This allows every piece of the codebase to work exactly with the requirements it really needs.
- Lift bounds from traits definitions. Keeping bounds only on `impl` blocks prevents redudant bounds pollution of the code base.
- More clever bounds on `impl` blocks. Require only the minimum to make `impl` work. For the `transparent` behavior the bounds should bubble-up structurally (structure implements trait if all the required fields implement trait).
- Make `TypeInfo` and `Context` as generic type parameters for all possible implementations, and require `Borrow<ConcreteContextType>` for the implementation needs. Not a single implemetation should put a concrete type into these type parameters, otherwise the whole polymorphism will be broken (as we have now with associative types). This allows us to fully remove the `context =` arguments in macros, and to use different context types in different methods of a single type. As for type info, it allows to combine dynamic and static schemas naturally.
- Impose a lifetime in `FromInputValue` trait, like `serde::Deserialize<'de>` does. This allows to pass references as arguments (and so parse unsized types) to fields in the similar degree as supported by `serde` deserialization.
- Decouple the responsibility of omitting orphan rules from the `ScalarValue` type parameter into a separate type parameter. Remove `ScalarValue` type parameter in places where it's not really required.

## Example

Meet a shiny new Juniper:
```rust
struct Human(u32);

#[graphql::object]
impl Human {
    fn id(&self) -> u32 {
        self.0
    }

   // Yes, we now have some degree of borrowing in arguments.
   fn has_name(name: &str) -> bool { 
       name == "John"
   }
   
   fn avatar(&self, context: &FileStorage) -> Option<String> {
       Some(context.lookup_avatar(self.0)?.to_base64_str())
   }

   fn profile(&self, context: &dyn UserRepository) -> Profile {
       context.get_profile(self.0).expect("profile should always exist")
   }
}

struct Context {
    repo: Box<dyn UserRepository>,
    storage: FileStorage,
}

impl Borrow<dyn UserRepository> for Context {
    fn borrow(&self) -> &dyn UserRepository {
        &*self.repo
    }
}

impl Borrow<FileStorage> for Context {
    fn borrow(&self) -> &FileStorage {
        &self.storage
    }
}

let context = Context::init();
juniper::execute(query, None, &schema, &graphql_vars! {}, &context).await;
// Code won't compile if the type, passed as context, doesn't implement all
// the `Borrow`s required by schema fields.
// For a single context type it works naturally due to
// existing `impl<T> Borrow<T> for T` blanket impl.
```

## Further investigation

The solution for the following _additional capabilities_ is not figured out yet and requires further investigation:
- [x] Polymorphism over `ScalarValue`.  At the moment the concrete `ScalarValue` type in a leaf type implementation polutes all the schema the same way the `Context`/`TypeInfo` does. But we cannot just `Borrow<CustomScalar>` from generic like we're doing for `Context`. Still it would be nice to figure out the way for combining types using different `ScalarValue`s in their implementations.
- [x] Separate type parameter for omitting orphan rules. At the moment we hack via using a custom `ScalarValue` type in implementations for this purpose. But it's not the responsibility of the `ScalarValue` notion, and using `ScalarValue` for that makes to put it into places where it's not required at all (like `reflect::BaseType` trait). Having a separate mechanic for omitting orphan rules would be much more intuitive. However, to work this polymorphically, we need to resolve the same implications imposed by previous question.

The main problem with polymorphism in both questions is that we cannot express `forall` quantifier for types in Rust:
```rust
trait Foo<T> {}
impl<A, B, T> Foo<A> for Box<T> where T: Foo<B> {}

// error[E0207]: the type parameter `B` is not constrained by the impl trait, self type, or predicates
//  --> src/lib.rs:2:9
//   |
// 2 | impl<A, B, T> Foo<A> for Box<T> where T: Foo<B> {}
//   |         ^ unconstrained type parameter
```

__Solved:__ See https://github.com/graphql-rust/juniper/pull/1072#issuecomment-1147604603

## Progress

This is a huge work, as requires an accurate re-implementation of the whole Juniper resolving and type registration machinery.
- [x] Pointer types
    - [x] `&T` (ref)
    - [x] `&mut T` (mut ref)
    - [x] `Box<T>`
    - [x] `Rc<T>`
    - [x] `Arc<T>`
    - [x] `Cow<'_, T>`
- [x] Nullable types
    - [x] `Option<T>`
    - [x] `Nullable<T>`
- [ ] List types
    - [x] `Vec<T>`
    - [x] `[T]` (slice)
    - [x] `[T; N]` (array)
    - [ ] `HashSet` (requires #1070)
    - [ ] `BTreeSet`
- [x] `str`
- [x] `#[graphql::Scalar]` macro
- [x] `#[graphql::Enum]` macro (requires #1047)
- [x] `#[graphql::InputObject]` macro (requires #1052)
- [ ] `#[graphql::Object]` macro
- [ ] `#[graphql::Union]` macro
- [ ] `#[graphql::Interface]` macro (requires #1028)
- [ ] `#[graphql::Subscription]` macro

### DIFF
--- a/juniper/src/schema/schema.rs
+++ b/juniper/src/schema/schema.rs
@@ -141,7 +141,7 @@ impl<'a, S: ScalarValue + 'a> SchemaType<'a, S> {
         self.description.as_deref()
     }
 
-    fn types(&self) -> Vec<TypeType<S>> {
+    fn types(&self) -> Vec<TypeType<'_, S>> {
         self.type_list()
             .into_iter()
             .filter(|t| {
@@ -156,21 +156,21 @@ impl<'a, S: ScalarValue + 'a> SchemaType<'a, S> {
     }
 
     #[graphql(name = "queryType")]
-    fn query_type_(&self) -> TypeType<S> {
+    fn query_type_(&self) -> TypeType<'_, S> {
         self.query_type()
     }
 
     #[graphql(name = "mutationType")]
-    fn mutation_type_(&self) -> Option<TypeType<S>> {
+    fn mutation_type_(&self) -> Option<TypeType<'_, S>> {
         self.mutation_type()
     }
 
     #[graphql(name = "subscriptionType")]
-    fn subscription_type_(&self) -> Option<TypeType<S>> {
+    fn subscription_type_(&self) -> Option<TypeType<'_, S>> {
         self.subscription_type()
     }
 
-    fn directives(&self) -> Vec<&DirectiveType<S>> {
+    fn directives(&self) -> Vec<&DirectiveType<'_, S>> {
         self.directive_list()
     }
 }
@@ -214,7 +214,7 @@ impl<'a, S: ScalarValue + 'a> TypeType<'a, S> {
     fn fields(
         &self,
         #[graphql(default = false)] include_deprecated: Option<bool>,
-    ) -> Option<Vec<&Field<S>>> {
+    ) -> Option<Vec<&Field<'_, S>>> {
         match self {
             TypeType::Concrete(&MetaType::Interface(InterfaceMeta { ref fields, .. }))
             | TypeType::Concrete(&MetaType::Object(ObjectMeta { ref fields, .. })) => Some(
@@ -231,14 +231,14 @@ impl<'a, S: ScalarValue + 'a> TypeType<'a, S> {
         }
     }
 
-    fn of_type(&self) -> Option<&TypeType<S>> {
+    fn of_type(&self) -> Option<&TypeType<'_, S>> {
         match self {
             TypeType::Concrete(_) => None,
             TypeType::List(l, _) | TypeType::NonNull(l) => Some(&**l),
         }
     }
 
-    fn input_fields(&self) -> Option<&[Argument<S>]> {
+    fn input_fields(&self) -> Option<&[Argument<'_, S>]> {
         match self {
             TypeType::Concrete(&MetaType::InputObject(InputObjectMeta {
                 ref input_fields,
@@ -343,7 +343,7 @@ impl<'a, S: ScalarValue + 'a> Field<'a, S> {
         self.description.as_deref()
     }
 
-    fn args(&self) -> Vec<&Argument<S>> {
+    fn args(&self) -> Vec<&Argument<'_, S>> {
         self.arguments
             .as_ref()
             .map_or_else(Vec::new, |v| v.iter().collect())
@@ -434,7 +434,7 @@ impl<'a, S: ScalarValue + 'a> DirectiveType<'a, S> {
         self.is_repeatable
     }
 
-    fn args(&self) -> &[Argument<S>] {
+    fn args(&self) -> &[Argument<'_, S>] {
         &self.arguments
     }
 
--- a/juniper_codegen/src/common/parse/mod.rs
+++ b/juniper_codegen/src/common/parse/mod.rs
@@ -124,9 +124,16 @@ pub(crate) trait TypeExt {
     /// inferring.
     fn lifetimes_anonymized(&mut self);
 
+    /// Anonymizes all the lifetime parameters of this [`syn::Type`] (except
+    /// the `'static` ones), making it suitable for using in contexts with
+    /// inferring, and returns it as a new type, leaving the original one
+    /// unchanged.
+    fn to_anonymized_lifetimes(&self) -> syn::Type;
+
     /// Lifts all the lifetimes of this [`syn::Type`] (even elided, but except
     /// `'static`) to a `for<>` quantifier, preserving the type speciality as
-    /// much as possible.
+    /// much as possible, and returns it as a new type, leaving the original one
+    /// unchanged.
     fn to_hrtb_lifetimes(&self) -> (Option<syn::BoundLifetimes>, syn::Type);
 
     /// Returns the topmost [`syn::Ident`] of this [`syn::TypePath`], if any.
@@ -231,6 +238,12 @@ impl TypeExt for syn::Type {
                 lt.ident = syn::Ident::new("_", Span::call_site());
             }
         });
+    }
+
+    fn to_anonymized_lifetimes(&self) -> syn::Type {
+        let mut ty = self.clone();
+        ty.lifetimes_anonymized();
+        ty
     }
 
     fn to_hrtb_lifetimes(&self) -> (Option<syn::BoundLifetimes>, syn::Type) {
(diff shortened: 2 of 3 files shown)


tyranron:
Regarding "Further investigation" section...

> Polymorphism over `ScalarValue`.

Well, this is simple. Once we decouple the orphan rules omitting responsibility from `ScalarValue` type param, it worths nothing to act with the `ScalarValue` the same way as with `Context`/`TypeInfo`: keep it as generic everywhere and require `Into<MyCustomScalarValue>` in concrete `impl` blocks. This way the requirement will bubble-up naturally up to the place where root node is `execute`ed, without messing up any intermediate types.

> Separate type parameter for omitting orphan rules.

> The main problem with polymorphism in both questions is that we cannot express `forall` quantifier for types in Rust:
> ```rust
> trait Foo<T> {}
> impl<A, B, T> Foo<A> for Box<T> where T: Foo<B> {}
>
> // error[E0207]: the type parameter `B` is not constrained by the impl trait, self type, or predicates
> //  --> src/lib.rs:2:9
> //   |
> // 2 | impl<A, B, T> Foo<A> for Box<T> where T: Foo<B> {}
> //   |         ^ unconstrained type parameter
> ```

This one is trickier as hits something quite fundamental regarding the Rust type system:
1. To omit orphan rules we need to specify _our_ type in a type parameter. Specifying it as an associated type doesn't allow to omit orphan rules, as doesn't give a dimension for parametrization. Associated type - is a concrete type, not a parameter.
2. To be able to write the `impl` above, we should have `B` as a associated type parameter, so it be a concrete type, not a "Rust, please, guess it out of nowhere" type parameter.

We cannot have 1 and 2 at the same time, as the requirements are clear contradiction.

To have them both, we need to introduce some transparent wrapper type holding our type parameter `B`:
```rust
struct Owned<X, By>(X, PhatomData<By>);
```
Now our `B` is constrained and we have the desired "polymorphism":
```rust
impl<A, B, T> Foo<A> for Owned<Box<T>, B> where T: Foo<B> {}
```
Hooray? Not yet! The problem has shifted to the place where we should or should not wrap the type. Unfortunately, we cannot do that seamlessly.
```rust
trait IntoOwned {
    type Owned;
}

struct Mine;
impl juniper::IntoOwned for u8 {  // oops! orphan rules again
    type Owned = juniper::Owned<Self, Mine>;
}
```
Without seamless conversion, we cannot use this implementation just like that:
```rust
#[derive(graphql::Object)]
struct Human {
    id: u8,
}
```
This shows very well our fundamental limitation of unconstrained type parameter, described above. Who should guess which implementation should be used here? Different crates may provide different `impl`s for `u8`. Well, so we may conclude that the user code _must specify the desired implementation_ to use.

Ok, then:
```rust
#[derive(graphql::Object)]
struct Human {
    id: Owned<u8, Mine>,
}
```
Hmm, not quite ergonomic. We don't want to mess with wrappers in the user code.
```rust
#[derive(graphql::Object)]
struct Human {
    #[graphql(owner = Mine)]
    id: u8,  // and we do wrapping under-the-hood in the macro expansion only
}
```
Naming still requires bikeshedding to be better, though.

Conclusions?

With this small ergonomic hit (in a form of additional attribute argument in places where foreign types and our types are connected) we solve the problem of "polymorphism" for a type parameter for omitting orphan rules. This way a custom type won't pollute all the schema types above it, just a field which refers it directly.

And now we're able to use `ScalarValue` only where it's really needed.

ilslv:
@tyranron this sounds a lot like [`#[serde(with = "module")]`][1] attribute and looks like a big ergonomic improvement 👍

From what i can tell, there is no way we can exactly replicate [`#[serde(with = "module")]`][1] by providing `mod` with a couple of functions (maybe until we can use `fn` pointers inside `const` generics?), but I think we can require users also add a type alias for a local struct:

```rust
#[derive(graphql::Object)]
struct Human {
    #[graphql(with = "custom")]
    id: u8,  // and we do wrapping under-the-hood in the macro expansion only
}

mod custom {
    pub(super) enum Custom {}

    // functions ...
}
```




[1]: https://serde.rs/field-attrs.html#with

tyranron:
@ilslv they look similar, but the logic seems different there. `with` is more [like "use intermediate transformation"](https://github.com/graphql-rust/juniper/issues/780), while `owner` (renamed to `behavior`  in sources) is like "use the concrete implementation for this type".

> (maybe until we can use `fn` pointers inside `const` generics?)

Nope, as `const` values are values, not types, so don't interfere with orphan rules.

tyranron:
_Nota bene_

At the moment we have a problem:
```
error[E0637]: `&` without an explicit lifetime name cannot be used here
   --> juniper/src/schema/schema.rs:140:37
    |
140 |     fn description(&self) -> Option<&str> {
    |                                     ^ explicit lifetime name needed here
```

Meaning that `Option<&str>: ::juniper::resolve::Resolve` bound at `impl` level is problematic.

Seems to be resolvable by quantifying all the type lifetimes into HTRB:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=1f8b36c78e61a771540b7a3324f9dd8b

LegNeato:
Do GATs change any of the design thinking here?

tyranron:
@LegNeato at the moment I don't see any need of them here, but it may appear eventually to solve some problems.

elementary-particle:
Are we going to complete this rework eventually? This is a lot of work indeed, we should not leave them around.

PR #1080, opened 2022-07-06
Fix input value coercion with defaults (#1073)

### LINKED ISSUE #1073: Input value defaulting behaviour is wrong
**Describe the bug**
The current behaviour for default values does not align with the GraphQL spec, or the reference implementation:

1. For a non-null field with a default value, it should be possible to override the value with a null value:
```graphql
# Schema
type Query {
  test(arg: Int = 100): Int
}
# Query
query {
  test(arg: null)
}
```
Currently Juniper takes the default value in this case.

2. A similar issue applies for variables:
```graphql
query ($arg: Int) {
  test(arg: $arg)
}
// or even with a variable default
query ($arg: Int = 10) {
  test(arg: $arg)
}
```
Passing an explicit `null` as the value for `$arg` should pass through to the `test` resolver.

3. When declaring a non-optional field as having a default value, Juniper currently downgrades the type to non-null when adding the default. This incorrectly implies that an explicit null is allowable.

E.g.:
```rust
#[graphql_object]
impl Query {
    fn non_null_default(#[graphql(default = 100)] arg: i32) -> Option<i32> {
        Some(arg)
    }
}
```
Results in an exposed schema that is similar to this:
```graphql
type Query {
  nonNullDefault(arg: Int = 100): Int
}
```
instead of:
```graphql
type Query {
  nonNullDefault(arg: Int! = 100): Int
}
```

## The GraphQL Spec

Under [Input Coercion](https://spec.graphql.org/draft/#sec-Input-Objects.Input-Coercion) section of the spec, it is specified that the default value should only apply if no value is provided:

> If no value is provided for a defined input object field and that field definition provides a default value, the default value should be used.

And additionally that an explicit `null` is considered semantically different to having not provided a value:

> If the value null was provided for an input object field, and the field’s type is not a non-null type, an entry in the coerced unordered map is given the value null. In other words, there is a semantic difference between the explicitly provided value null versus having not provided a value.

And similarly for variables, the same is true:

> If a variable is provided for an input object field, the runtime value of that variable must be used. If the runtime value is null and the field type is non-null, a field error must be raised. If no runtime value is provided, the variable definition’s default value should be used. If the variable definition does not provide a default value, the input object field definition’s default value should be used.

## Comparison to the reference implementation
I have created a repo here with a suite of tests against both Juniper and graphql-js that can be used for comparison: https://github.com/meiamsome/juniper-default-value

The results of the comparision tests are as follows:
graphql-js:
```
> jest index.test.js

 PASS  ./index.test.js
  ✓ default value (3 ms)
  ✓ literal number (1 ms)
  ✓ literal null
  ✓ variable field default value
  ✓ variable default value
  ✓ variable default value explicit null
  ✓ variable number
  ✓ variable null
  ✓ schema for default non null is non null (1 ms)
```

juniper:
```
running 9 tests
test test::variable_number ... ok
test test::variable_default_value ... ok
test test::default_value ... ok
test test::variable_field_default_value ... ok
test test::literal_null ... FAILED
test test::variable_null ... FAILED
test test::literal_number ... ok
test test::variable_default_value_explicit_null ... FAILED
test test::schema_for_default_non_null_is_non_null ... FAILED

failures:

---- test::literal_null stdout ----
thread 'test::literal_null' panicked at 'assertion failed: `(left == right)`
  left: `Some(100)`,
 right: `None`', src/lib.rs:82:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- test::variable_default_value_explicit_null stdout ----
thread 'test::variable_default_value_explicit_null' panicked at 'assertion failed: `(left == right)`
  left: `Some(100)`,
 right: `None`', src/lib.rs:146:9

---- test::variable_null stdout ----
thread 'test::variable_null' panicked at 'assertion failed: `(left == right)`
  left: `Some(100)`,
 right: `None`', src/lib.rs:190:9

---- test::schema_for_default_non_null_is_non_null stdout ----
thread 'test::schema_for_default_non_null_is_non_null' panicked at 'assertion failed: `(left == right)`
  left: `Object(Object { key_value_list: {"__schema": Object(Object { key_value_list: {"queryType": Object(Object { key_value_list: {"fields": List([Object(Object { key_value_list: {"name": Scalar(String("test")), "args": List([Object(Object { key_value_list: {"name": Scalar(String("arg")), "type": Object(Object { key_value_list: {"name": Scalar(String("Int")), "kind": Scalar(String("SCALAR")), "ofType": Null} })} })])} }), Object(Object { key_value_list: {"name": Scalar(String("nonNullDefault")), "args": List([Object(Object { key_value_list: {"name": Scalar(String("arg")), "type": Object(Object { key_value_list: {"name": Scalar(String("Int")), "kind": Scalar(String("SCALAR")), "ofType": Null} })} })])} })])} })} })} })`,
 right: `Object(Object { key_value_list: {"__schema": Object(Object { key_value_list: {"queryType": Object(Object { key_value_list: {"fields": List([Object(Object { key_value_list: {"name": Scalar(String("test")), "args": List([Object(Object { key_value_list: {"name": Scalar(String("arg")), "type": Object(Object { key_value_list: {"name": Scalar(String("Int")), "kind": Scalar(String("SCALAR")), "ofType": Null} })} })])} }), Object(Object { key_value_list: {"name": Scalar(String("nonNullDefault")), "args": List([Object(Object { key_value_list: {"name": Scalar(String("arg")), "type": Object(Object { key_value_list: {"name": Null, "kind": Scalar(String("NON_NULL")), "ofType": Object(Object { key_value_list: {"name": Scalar(String("Int")), "kind": Scalar(String("SCALAR"))} })} })} })])} })])} })} })} })`', src/lib.rs:301:9


failures:
    test::literal_null
    test::schema_for_default_non_null_is_non_null
    test::variable_default_value_explicit_null
    test::variable_null

test result: FAILED. 5 passed; 4 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

ilslv:
Resolves #1073

## Solution

Don't additionally wrap field's type into an `Option` and don't replace missing `Variables` with `InputValue::Null` too soon. 

## Checklist

- Created PR:
    - [x] In [draft mode][l:1]
    - [x] Has assignee
    - [x] and approved
- Before merge:
    - [x] Milestone is set
    - [x] PR's name and description are correct and up-to-date
    - [x] `Draft: ` prefix is removed
    - [x] All temporary labels are removed

### DIFF
--- a/book/src/types/scalars.md
+++ b/book/src/types/scalars.md
@@ -205,10 +205,7 @@ where
         .ok_or_else(|| format!("Expected `String` or `Int`, found: {}", v))
 }
 
-fn parse_token<S>(value: ScalarToken<'_>) -> ParseScalarResult<'_, S> 
-where
-    S: ScalarValue
-{
+fn parse_token<S: ScalarValue>(value: ScalarToken<'_>) -> ParseScalarResult<S> {
     <String as ParseScalarValue<S>>::from_str(value)
         .or_else(|_| <i32 as ParseScalarValue<S>>::from_str(value))
 }
@@ -258,7 +255,7 @@ impl StringOrInt {
             .ok_or_else(|| format!("Expected `String` or `Int`, found: {}", v))
     }
   
-    fn parse_token<S>(value: ScalarToken<'_>) -> ParseScalarResult<'_, S>
+    fn parse_token<S>(value: ScalarToken<'_>) -> ParseScalarResult<S>
     where
         S: ScalarValue,
     {
@@ -308,7 +305,7 @@ mod string_or_int {
             .ok_or_else(|| format!("Expected `String` or `Int`, found: {}", v))
     }
   
-    pub(super) fn parse_token<S>(value: ScalarToken<'_>) -> ParseScalarResult<'_, S>
+    pub(super) fn parse_token<S>(value: ScalarToken<'_>) -> ParseScalarResult<S>
     where
         S: ScalarValue,
     {
--- a/juniper/CHANGELOG.md
+++ b/juniper/CHANGELOG.md
@@ -71,7 +71,7 @@ All user visible changes to `juniper` crate will be documented in this file. Thi
 - Unsupported expressions in `graphql_value!` macro. ([#996], [#503])
 - Incorrect GraphQL list coercion rules: `null` cannot be coerced to an `[Int!]!` or `[Int]!`. ([#1004])
 - All procedural macros expansion inside `macro_rules!`. ([#1054], [#1051])
-- Input value coercion with defaults. ([#1073], [#1078])
+- Input value coercion with defaults. ([#1073], [#1080])
 
 [#503]: /../../issues/503
 [#750]: /../../issues/750
@@ -102,7 +102,7 @@ All user visible changes to `juniper` crate will be documented in this file. Thi
 [#1057]: /../../pull/1057
 [#1060]: /../../pull/1060
 [#1073]: /../../issues/1073
-[#1078]: /../../pull/1078
+[#1080]: /../../pull/1080
 [ba1ed85b]: /../../commit/ba1ed85b3c3dd77fbae7baf6bc4e693321a94083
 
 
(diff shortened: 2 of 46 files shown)


ilslv:
## FCM

```
Fix input value coercion with defaults (#1073, #1080)
```

tyranron (approved):
@ilslv what a nice job 👍

Thanks! ♥️

PR #1084, opened 2022-07-18
Quality of life on stable channel

btielen:
While working on an integration for Axum, I experienced some inconveniences:

1. In the root directory `cargo test` fails on rust-stable because
    - `juniper_codegen_tests` fails on stable
    - `juniper_book_tests` fails sometimes, I think because the life cycle of the build script in `book/tests/build.rs` is unpredictable.
3. `cargo fmt` fails on rust stable, while `CONTRIBUTING.md` states that formatting has to be done on stable (like `cargo clippy`)
4. `cargo test -p juniper_integration_tests` takes long to build and even crashed my computer a few times for some reason. See also this [todo](https://github.com/graphql-rust/juniper/blob/ef7a7e82296b9217baf551fa7fb718e4d472bb7f/.github/workflows/ci.yml#L201)

## Changes
This PR fixes these inconveniences by:

- Only running the `juniper_codegen_tests` on nightly
- Removing `juniper_book_tests` and using `mdbook test` instead
- Moving back to formatting on stable
- Splitting and running `juniper_integration_tests` in smaller units

## Conclusion

- Running `cargo test` from the root directory should now pass all tests on stable, beta and nightly
- Building and testing time for `juniper_integration_tests` reduced from 3m56 to 2m04 on my computer
- Allows testing of `juniper_integration_tests` on windows in the CI-pipeline
- `cargo clippy` and `cargo fmt` should now be both run on stable


btielen:
So `mdbook test` seems to fail on Windows, I was just able to confirm this on a local machine. We get this error:

`LINK : fatal error LNK1181: cannot open input file 'windows.lib'`

By removing (the unnecessary) `extern crate tokio` from line 85 `book/src/advanced/subscriptions.md` I was able to cut down one error. The other error on line 98 of `book/src/types/objects/using_contexts.md` also uses tokio. Maybe the feature `sync` for tokio was not activated?

Futher investigastion needed.

tyranron (changes requested):
@btielen thanks! Overall seems to be a nice life improvement, especially with book tests!

As for formatting on stable, I'd rather keep formatting on nightly. Stable formatting lacks some good features, we're intending to use. Just use `make fmt` for formatting. `CONTRIBUTING.md` just hasn't been updated for a while 😅

btielen:
So it looks like the `mdbook test` failing on windows is caused by a bug in rustdoc. I issued a a [bug report](https://github.com/rust-lang/rust/issues/99466) at the rust repo. I was wondering how a crate like Tokio is testing their documentation book. It turns out they are also using rustdoc, but their tests also completely fail on Windows. The doc tests are also not being run on windows on the CI-pipeline. For now I ignored the test that was causing the problems.

I made the requested changes. 

@ilslv The `juniper_codegen_tests` now fail, although I didn't make any changes to them. Is it caused by a new version of rust nightly? Would it be possible to run the tests only on stable for more stability?

ilslv:
@btielen 

> Is it caused by a new version of rust nightly?

Almost certainly yes.

> Would it be possible to run the tests only on stable for more stability?

I'm not sure about that. We use `nightly`, because there we emit unstable [`proc_macro` diagnostics](https://github.com/rust-lang/rust/issues/54140) using [`proc_macro_error` crate](https://docs.rs/proc-macro-error/latest/proc_macro_error/). You can run tests with env variable `TRYBUILD=overwrite` and it should automatically overwrite failing tests.

Issue #1085, opened 2022-07-20
Error when explicit null provided for nullable list input parameter


ajawalker:
**Description**

When an input parameter type is a nullable list, providing an explicit value of `null` as the argument for that parameter generates an error.

**To Reproduce**

In `Cargo.toml`: `juniper = "0.15.9"`

Query object:

```
pub struct Query;

#[graphql_object]
impl Query {
    fn test(input: Option<Vec<i32>>) -> Option<Vec<i32>> {
        input
    }
}
```
This query:
```
query {
  test(input:null)
}
```
Returns this error:
```
{
  "errors": [
    {
      "message": "Invalid value for argument \"input\", expected type \"[Int!]\"",
      "locations": [
        {
          "line": 2,
          "column": 14
        }
      ]
    }
  ]
}
```
Providing an explicit value or providing no value (an implicit null) both produce the expected result. Also, an explicit `null` value works for other nullable types, just not nullable list types.

**Expected behavior**

Providing an explicit `null` argument for a nullable list parameter should not produce an error, but should behave exactly the same as an implicit `null` for `Option<Vec<_>>` types, or become the `ImplicitNull` variant for `Nullable<Vec<_>>` types.

tyranron:
[Released in 0.15.10](https://github.com/graphql-rust/juniper/releases/tag/juniper-v0.15.10).

PR #1086, opened 2022-07-21
Accept explicit null argument for nullable list parameter

ajawalker:
Addresses https://github.com/graphql-rust/juniper/issues/1085

Explicit `null` values are not being accepted for nullable list parameters. This change causes explicit `null`s (as well as variables) to be accepted. Also adds tests for `null` values being provided for both nullable and non-nullable lists.


tyranron (approved):
@ajawalker thanks! 🍻

PR #1099, opened 2022-08-29
Remove indirect time 0.1.44 dependency by chrono

lclc:
By default chrono uses time 0.1.44, which has a vulnerability:
https://rustsec.org/advisories/RUSTSEC-2020-0071

chrono 0.4.20 does not depend on the vulnerable parts of the time 0.1.x
versions, see https://github.com/chronotope/chrono/releases/tag/v0.4.20

This change is to satisfy simple dependency scanners.

### DIFF
--- a/juniper/Cargo.toml
+++ b/juniper/Cargo.toml
@@ -40,7 +40,7 @@ anyhow = { version = "1.0.32", default-features = false, optional = true }
 async-trait = "0.1.39"
 bigdecimal = { version = "0.3", optional = true }
 bson = { version = "2.4", features = ["chrono-0_4"], optional = true }
-chrono = { version = "0.4", features = ["alloc"], default-features = false, optional = true }
+chrono = { version = "0.4.20", features = ["alloc"], default-features = false, optional = true }
 chrono-tz = { version = "0.6", default-features = false, optional = true }
 fnv = "1.0.3"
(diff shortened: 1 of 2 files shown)


lclc:
Actually doesn't help much because of the https://github.com/iron/logger dependency which depends on time = "0.1.42" (https://github.com/iron/logger/blob/master/Cargo.toml#L13) and hasn't been updated for 2.5 years.

tyranron:
@lclc `logger` is used in `examples/` only, so may be thrown away, or replaced with some custom `log` or `tracing` stuff.

But the `time 0.1.44` is still [required by `hyper 0.10`](https://github.com/hyperium/hyper/blob/v0.10.16/Cargo.toml#L29), which, in turn, is [required by `iron 0.6`](https://github.com/iron/iron/blob/0.6.1/Cargo.toml#L33), and thus neither one is `optional = true`, we cannot eliminate it in any way.

lclc:
Indeed. I've been trying to get rid of the time 0.1.44 dependency in our project, but chrono and time 0.1.44 is a dependency of so many dependencies (of dependencies, of dependencies, ..) that it's pretty much impossible.

Anyway, the next release of chrono (0.5) plans to remove the time dependency completely.

That would leave [iron](https://github.com/iron/iron/) with a direct dependency on time. Unfortunately, iron seems to be unmaintained.

tyranron (approved):

PR #1101, opened 2022-09-12
Change body type of `juniper-hyper` functions from `Body` to `String`

### LINKED ISSUE #1096: Return `Response<String>` instead of `Response<Body>` in hyper integration
Sometimes it's useful to inspect the response returned by `juniper_hyper::graphql`. For example, in case of an error I would like to debug-log the whole JSON response. This is currently very tricky. 

The function returns `Response<Body>` and the only way to read that `Body` is via the `Stream` trait (or the helper functions in `hyper::body` which also use that trait). But all those methods require `Body` by value or `&mut Body`. And in general that makes sense: `Body` can be many things and is not necessarily a byte buffer. 

The best I could come up with to log the body is:

```rust
let (parts, body) = response.into_parts();
let body = hyper::body::to_bytes(body).await
    .expect("could not read API response body (this should never happen)");
debug!("Response body of failed API request: {}", String::from_utf8_lossy(&body));
response = hyper::Response::from_parts(parts, body.into());
```

I need to completely unpack the response to then assemble it again. And the `await` and `expect` are both not nice here.

From a quick glance at the `juniper_hyper` code, it seems like the `Body` is always created via `From<String>`. But `String` implements the trait `hyper::HttpBody` directly! So my suggestion is that all functions return `Response<String>`. With that, my logging would be as simple as:

```
debug!("Response body of failed API request: {}", response.body());
```

LukasKalbertodt:
All responses were created from strings anyway (via `Body::from`), but `String` implements the `Body` trait directly. With this change, it is easy to inspect the response body, which is not true for `Body`.

Fixes #1096

### DIFF
--- a/juniper_hyper/CHANGELOG.md
+++ b/juniper_hyper/CHANGELOG.md
@@ -11,6 +11,7 @@ All user visible changes to `juniper_hyper` crate will be documented in this fil
 ### BC Breaks
 
 - Switched to 0.16 version of [`juniper` crate].
+- Changed the return type of all functions from `Response<Body>` to `Response<String>`
 
 
 
--- a/juniper_hyper/CHANGELOG.md
+++ b/juniper_hyper/CHANGELOG.md
@@ -11,7 +11,10 @@ All user visible changes to `juniper_hyper` crate will be documented in this fil
 ### BC Breaks
 
 - Switched to 0.16 version of [`juniper` crate].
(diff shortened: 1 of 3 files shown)


LukasKalbertodt:
The single CI failure is not caused by the changes and needs to be restarted.

tyranron (approved):
@LukasKalbertodt thanks!

PR #1136, opened 2023-01-05
make_graphql_filter now takes Arc schema; fix #1135

### LINKED ISSUE #1135: Why doesn't juniper_warp::make_graphql_filter take schema: Arc<RootNode<...>>?
I notice that `juniper_hyper::graphql` takes the schema/rootnode argument as type `Arc<RootNode<...>>` but `juniper_warp::make_graphql_filter` only takes the schema as `RootNode<...>`.

This seems to make `juniper_warp` difficult to use in practice in the scenario of library integrations (e.g., a juniper-based GraphQL plugin for tauri).

In the example [here](https://github.com/graphql-rust/juniper/blob/6aa63854372bf50d7134491495568593e968c7c6/juniper_warp/examples/warp_server.rs#L13) this issue is side-stepped completely by creating a function `schema()` which returns a fresh `RootNode` instance.

In a library integration, however, you do not have access to the components necessary to construct a new `RootNode`.

It seems that `juniper_warp` should take the argument as an `Arc<RootNode<...>>` like `juniper_hyper` in order to avoid this problem.

Is there a reason why it is not defined this way?

silvanshade:
Fixes #1135

### DIFF
--- a/book/src/servers/index.md
+++ b/book/src/servers/index.md
@@ -12,6 +12,6 @@ third-party integration crates that will get you there.
     - [Iron](iron.md)
     - [Hyper](hyper.md)
 - [Third Party Integrations](third-party.md)
-    - [Actix-Web](https://github.com/actix/examples/tree/master/juniper)
+    - [Actix-Web](https://github.com/actix/examples/tree/master/graphql/juniper)
     - [Finchers](https://github.com/finchers-rs/finchers-juniper)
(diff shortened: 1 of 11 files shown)


tyranron (approved):
@silvanshade thanks! And sorry for replying late on this.

I've tinkered your solution a bit, so both `Arc`ed and raw `schema` may be specified, making the change non-breaking and API more ergonomic.

PR #1145, opened 2023-02-21
Expose applies_for for look ahead

### LINKED ISSUE #1138: Ability to get applies_for using the look_ahead functionality
Currently there doesn't seem to be a public way to get the applies_for for children, e.g there is no way to get the fact that we want primaryFunction only for Droid.
```
query Hero {
    hero {
        name
        ... on Droid {
            primaryFunction
        }
        ... on Human {
            height
        }
    }
}",
```
The way to access the look_ahead functionality seems to be calling look_ahead() on an Executor. But this returns a LookAheadSelection while applies_for lives at the ChildSelection level.

Am I missing something? Is there another way to access this data?

daviddanielarch-nextroll:
Resolves #1138

This PR exposes the applies_for in the look ahead functionality by moving the current applies_for from the ChildSelection struct into the inner field of the LookAheadSelection.


daviddanielarch-nextroll:
The CI step that is failing seems to fail on master as well https://github.com/graphql-rust/juniper/actions/runs/4241804294/jobs/7372654238

juniper/src/executor/look_ahead.rs
```diff
@@ -100,7 +100,6 @@ where
 #[derive(Debug, Clone, PartialEq)]
 pub struct ChildSelection<'a, S: 'a> {
     pub(super) inner: LookAheadSelection<'a, S>,
-    pub(super) applies_for: Applies<'a>,
```
tyranron, line 103: This change makes `ChildSelection` completely redundant. I looked at the code and haven't seen using it outside this module. I guess it could be stripped out completely.

juniper/src/executor/look_ahead.rs
```diff
@@ -367,6 +374,9 @@ pub trait LookAheadMethods<'sel, S> {
 
     /// Get an iterator over the children for the current selection
     fn children(&self) -> Vec<&Self>;
+
+    /// Get the parent type in case there is any for this selection
+    fn applies_for(&self) -> Applies;
```
tyranron, line 379: Should we really expose `Applies` struct here? Wouldn't it be better to return `Option<&str>` instead, leaving `Applies` an inner implementation detail?

tyranron (changes requested):
@daviddanielarch-nextroll would like to hear your opinion.

juniper/src/executor/look_ahead.rs
daviddanielarch-nextroll: You are right. I'll remove it.

juniper/src/executor/look_ahead.rs
daviddanielarch-nextroll: Yeah, makes more sense.

PR #1180, opened 2023-08-02
Make types public

### LINKED ISSUE #1117: Fully Dynamic Schemas
First, I would like to say thank you for this great project.

I am looking to create a framework where the bulk of the code is written and rust and delegates to a language like python for business logic. Given these constraints, the schema of the API is inherently dynamic and thus is not known at compile time. It would be optimal if somehow juniper could support this (admittedly niche) use case. 

If juniper provided a lower level API that let you register types to a `Registry` dynamically and then propagate that up to an executable schema (side stepping traits like `GraphQLType` - which are not instance based), would allow for this to work.  

I looked to see if this is possible today given the API's exposed by juniper. While the registry type is exposed and does seem to be the main point of interaction, this means you lose much of the query validation and other great work done by the juniper create. 

I considered instead depending on [graphql-parser](https://github.com/graphql-rust/graphql-parser) and doing the validation and execution myself. But again, I lose a lot of what juniper has to offer. As an alternative to my idea above, Is there room for some crate between the ergonomics that juniper provides and the raw graphql parsing done in `graphql-parser`? 

**Additional context**
This issue has been brought up in several issues in the past. Namely issues #7 and #106 but haven't really been obviously resolved.

tusharmath:
Types can be used to generate a RootNode dynamically. This is a quick fix for #1117.
if this looks good, I can create a few more PRs that make some of the internal types public, allows developers to create RootNode and Schema using techniques other than Macros.


LegNeato (changes requested):
Can you add an integration test to make sure we don't break this in the future? Can you also add to the docs that these public types are experimental and not to be relied on, linking back to the issue? Thanks!

tyranron:
@LegNeato I'll look at this next week.

juniper/src/types/name.rs
```diff
@@ -6,7 +6,7 @@ use std::{
 };
 
 #[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub struct Name(String);
```
tyranron, line 9: @tusharmath I don't like this change, because it breaks encapsulation for the `Name` type here and allow freely bypass its invariants. It's far better here to provide a meaningful constructor-function, which checks the invariants.

juniper/src/lib.rs
```diff
@@ -33,8 +33,8 @@ mod ast;
 pub mod executor;
 mod introspection;
 pub mod parser;
-pub(crate) mod schema;
-mod types;
```
tyranron, line 37: @tusharmath here types are mostly re-exported to the crate's root. If something is missing, I think it's better to add that to the re-exports, rather than making these modules public.

tyranron (changes requested):
@tusharmath the proposed changes are quite controversial. Overall, I'm not a fan of them.

Could you show a code that requires these changes to achieve what you want? Having it, it would be better to come up with the solution that satisfies both of us.

LegNeato:
I'm going to close this out.  Feel free to comment if you want it reopened to continue discussion!

PR #1195, opened 2023-10-02
Use fully qualified path to types in generated code

### LINKED ISSUE #1194: GraphQLInputObject macro fails if crate has a local type alias named "Result"
**Describe the bug**
In a crate with this type alias:
```rust
type Result<T> = std::result::Result<T, diesel::result::Error>;
```

Deriving `GraphQLInputObject` for structs that include optional values will fail.

**To Reproduce**

Given the above type alias, and the following struct

```rust
#[derive(juniper::GraphQLInputObject)]
#[graphql(name = "UserInformation")]
pub struct Update {
    pub email: Option<String>,
    pub username: Option<String>,
}
```

The following error is returned:

```
error[E0107]: type alias takes 1 generic argument but 2 generic arguments were supplied
  --> models/src/user.rs:29:51
   |
29 | #[derive(Serialize, Deserialize, Debug, Validate, juniper::GraphQLInputObject)]
   |                                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |                                                   |
   |                                                   expected 1 generic argument
   |                                                   help: remove this generic argument
   |
note: type alias defined here, with 1 generic parameter: `T`
  --> models/src/lib.rs:52:6
   |
52 | type Result<T> = std::result::Result<T, diesel::result::Error>;
   |      ^^^^^^ -
   = note: this error originates in the derive macro `juniper::GraphQLInputObject` (in Nightly builds, run with -Z macro-backtrace for more info)
```

**Expected behavior**

Successful compilation, as happens when the type alias is renamed to anything else, for example, there are no issues with the following:

```rust
type DatabaseResult<T> = std::result::Result<T, diesel::result::Error>;
```

**Additional context**

```toml
juniper = { git = "https://github.com/graphql-rust/juniper.git", rev ="984973658217f73d63608283d18723e11ce65161", default-features = false, features = ["schema-language", "url", "chrono"] }
```

tpoliaw:
If user code has a type alias (or custom type) called `Result` or `Option` etc,
this would be used instead of the intended standard library types causing
surprising compile errors from generated code.

Using the fully qualified path allows the generated code to be isolated
from user types.

Fixes #1194

### DIFF
--- a/juniper_codegen/src/common/deprecation.rs
+++ b/juniper_codegen/src/common/deprecation.rs
@@ -102,12 +102,12 @@ impl Directive {
 
 impl ToTokens for Directive {
     fn to_tokens(&self, into: &mut TokenStream) {
-        let reason = self
-            .reason
-            .as_ref()
-            .map_or_else(|| quote! { None }, |text| quote! { Some(#text) });
+        let reason = self.reason.as_ref().map_or_else(
+            || quote! { ::core::option::Option::None },
+            |text| quote! { ::core::option::Option::Some(#text) },
+        );
         quote! {
-            .deprecated(::std::option::Option::#reason)
+            .deprecated(#reason)
         }
         .to_tokens(into);
     }
--- a/juniper_codegen/src/common/field/arg.rs
+++ b/juniper_codegen/src/common/field/arg.rs
@@ -335,13 +335,13 @@ impl OnMethod {
                                 ::juniper::IntoFieldError::<#scalar>::into_field_error(e)
                                     .map_message(|m| format!(#err_text, m))
                             })
-                    }, Ok))
+                    }, ::core::result::Result::Ok))
                 };
                 if for_async {
                     quote! {
                         match #arg {
-                            Ok(v) => v,
-                            Err(e) => return Box::pin(async { Err(e) }),
+                            ::core::result::Result::Ok(v) => v,
+                            ::core::result::Result::Err(e) => return ::std::boxed::Box::pin(async { ::core::result::Result::Err(e) }),
                         }
                     }
                 } else {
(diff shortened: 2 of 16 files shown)


tpoliaw:
Looking at this again there are many other cases that could fail in the same way (`Some`, `None`, `Send`, `Sync` etc). I will try and go through and add the paths for the other types but for traits/types that are available in both `std` and `core`, is there a preference for one over the other?

juniper_codegen/src/common/deprecation.rs
```diff
@@ -105,9 +105,9 @@ impl ToTokens for Directive {
         let reason = self
             .reason
             .as_ref()
-            .map_or_else(|| quote! { None }, |text| quote! { Some(#text) });
+            .map_or_else(|| quote! { ::std::option::Option::None }, |text| quote! { ::std::option::Option::Some(#text) });
         quote! {
-            .deprecated(::std::option::Option::#reason)
+            .deprecated(#reason)
```
tpoliaw, line 110: This worked before but I changed it to make it more clear that the types are correctly qualified. I spent quite a while trying to track down an `expected identifier, found ::` error caused by the doubly qualified path (`::std::option::Option::std::option::Option::Some(...)` after only changing the other line.

Happy to revert if preferred.

LegNeato:
> Looking at this again there are many other cases that could fail in the same way (`Some`, `None`, `Send`, `Sync` etc). I will try and go through and add the paths for the other types but for traits/types that are available in both `std` and `core`, is there a preference for one over the other?

The preference would be core ove std where possible. Last I looked we still build on no_std.

tpoliaw:
> Last I looked we still build on no_std

Does that include using the derived code? There are several instances of things like `String` and `Box` floating around that presumably require std.

I will update the changes in this PR to use core though. Should existing references to `::std::*` in derives be updated at the same time?

LegNeato:
Sorry, I was mistaken...I got my wires crossed with wasm

LegNeato:
Awesome, thanks!

PR #1191, opened 2023-09-14
Improve WebSocket integration

tyranron:
Follows and extends #1158  
Related to #1022, #1186

## Synopsis

In #1158 was added [`graphql-transport-ws` GraphQL over WebSocket Protocol](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md) integration. However, new `juniper_graphql_transport_ws` crate wasn't wired correctly into CI, and clear documentation "who is who" is missing, leading to issues like #1186

## Solution

Fill up the gaps around documentation, tooling and examples regarding new crates.

## Additionally

Provide out-of-the box resolvers for automatic negotiation of the used WebSocket protocol in integration crates.

PR #1196, opened 2023-10-16
Merge `juniper_graphql_transport_ws` and `juniper_graphql_ws` crates (#1022)

### LINKED ISSUE #1186: Subscription argumennt Error using juniper_graphql_transport_ws
I have
```rust
async fn subscriptions(
    req: HttpRequest,
    stream: web::Payload,
    schema: web::Data<Schema>,
) -> Result<HttpResponse, Error> {
    let context = Context::new()
    let schema = schema.into_inner();
    let config = ConnectionConfig::new(context);
    let config = config.with_keep_alive_interval(Duration::from_secs(15));

    subscriptions_handler(req, stream, schema, config).await
}
```
but `subscriptions_handler(req, stream, schema, config).await`  the `config` argument gives error

```
expected a `FnOnce<(HashMap<std::string::String, InputValue>,)>` closure, found `ConnectionConfig<context::Context>`
the trait `FnOnce<(HashMap<std::string::String, InputValue>,)>` is not implemented for `ConnectionConfig<context::Context>`
the trait `juniper_graphql_ws::Init<S, CtxT>` is implemented for `juniper_graphql_ws::ConnectionConfig<CtxT>`
required for `ConnectionConfig<context::Context>` to implement `juniper_graphql_ws::Init<DefaultScalarValue, context::Context>`
```

tyranron:
Part of #1022  
Resolves #1186
Follows #1158, #1191

## Motivation

Despite the raised concern for the initial implementation https://github.com/graphql-rust/juniper/pull/1158#issuecomment-1502016225:
> A common use-case is for servers to support both simultaneously and negotiate the correct protocol to use on each connection, as discussed in https://github.com/seanmonstar/warp/issues/760. To support that use-case, we can't force the users into one or the other at compile-time.
>
> I also think the changes, while relatively minor, are pervasive enough that it makes switching between the protocols via conditionals very messy and hard to follow.

Integration of both protocols into a single crate showed, that [fully separate crates (and therefore, types) introduce integration problems](https://github.com/graphql-rust/juniper/issues/1186), which are resolved by reusing [same types in both crates](https://github.com/graphql-rust/juniper/commit/984973658217f73d63608283d18723e11ce65161#diff-8eac07fbae2b725f66195a37e9f095c3462aec1b74b0fd92623fc441e3a5810eR32).

Following this, as we aim to support both protocols in our HTTP integration crates, both `juniper_graphql_transport_ws` and `juniper_graphql_ws` crates are always used and depend on each other, which makes a little sense to support them a separate crates.

## Solution

Instead, it's enough to keep only one `juniper_graphql_ws` crate, where support both protocols via _additive_ Cargo features. This will simplify maintenance and usage costs, while allowing to reuse similar types naturally, giving the desired level of polymorphism for supporting both protocols at the same time.

## Checklist

- [x] Tests are updated
- [x] Documentation is updated
- [x] CI is updated
- [x] CHANGELOGs are updated


tyranron:
@ccbrown since you're the original author, would like to see any comments on this before merging.

juniper_graphql_ws/Cargo.toml
```diff
@@ -3,17 +3,28 @@ name = "juniper_graphql_ws"
...
+keywords = ["apollo", "graphql-transport-ws", "graphql-ws", "subscription", "websocket"]
 exclude = ["/release.toml"]
 
+[package.metadata.docs.rs]
+all-features = true
+rustdoc-args = ["--cfg", "docsrs"]
+
+[features]
+graphql-transport-ws = []
+graphql-ws = []
```
ccbrown, line 26: Why gate these behind features at all? The modules for these are small and don't pull in additional dependencies. Unless I'm missing something, as a user I'm not sure why I would ever want to explicitly enable or disable these.

ccbrown (commented):
This looks good to me! I just have one non-blocking question.

juniper_graphql_ws/Cargo.toml
tyranron: @ccbrown I was thinking that some people may want to opt-in for the concrete protocol version and don't consider another one at all, thus avoid increasing compilation times and auto-completion pollution in any way.

PR #1200, opened 2023-10-28
Avoid an unecessary clone in `LookAheadMethods::children`

soerenmeier:
Depending on how the trait will be used in the future this might not be a good idear.
Another possibility would be to return `impl Iterator<Item=&Self>` this is already possible in nightly and was stablized two weeks ago https://github.com/rust-lang/rust/pull/115822


tyranron (approved):
@soerenmeier makes sense. Thank you!

PR #1208, opened 2023-11-09
Abstract (start, end) from Spanning struct into new struct Span

audunhalland:
This makes it easier to treat the source region (the Span) separately from the spanned item in Spanning.

### DIFF
--- a/juniper/src/parser/utils.rs
+++ b/juniper/src/parser/utils.rs
@@ -50,19 +50,19 @@ impl Span {
     }
 }
 
-/// Data structure used to wrap items with start and end markers in the input source
-#[derive(Debug, Clone, PartialEq, Eq, Hash, Copy)]
+/// Data structure used to wrap items into a [`Span`].
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
 pub struct Spanning<T> {
-    /// The wrapped item
+    /// Wrapped item.
     pub item: T,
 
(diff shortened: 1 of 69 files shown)


tyranron (approved):
@audunhalland thanks for the effort!

I've bikeshedded your solution a bit, around accepting/returning references for `Copy` type in new APIs, and readability/understandability where `Spanning::start()`/`Spanning::end()` methods are used.

juniper/src/parser/utils.rs
```diff
@@ -8,99 +8,129 @@ pub struct SourcePosition {
...
+        Self {
+            start: SourcePosition::new_origin(),
+            end: SourcePosition::new_origin(),
         }
     }
+}
+
+/// Data structure used to wrap items into a [`Span`].
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+pub struct Spanning<T> {
```
audunhalland, line 55: In the planned follow-up PR/rebase, which adds spanning to arguments and LookAheads, this struct gets a second generic argument `<Sp = Span>`, so that the span can be a reference in some contexts (one-word copy instead of six words).

I was a bit unsure whether to include that change in this PR instead, because I got the idea after this refactor was done.

tyranron: @audunhalland it would be OK to have in the follow-up PR.

PR #1209, opened 2023-11-09
Spanned arguments

audunhalland:
Continued from #1206

This change improves diagnostics when performing manual/domain-specific post-validation of arguments using LookAheadArguments. In the case of an error, the produced error message may be able to refer to the exact position in the problematic GraphQL arguments.

On an unrelated note, I noticed `LookAheadArgument` does a deep/greedy transform of the `InputValue` to perform variable substitution. A potential improvement to this is to only do this lazily for each expansion step, i.e. the LookAheadArgument privately keeps a reference to its `InputValue` and the variable table and somehow exposes a LookAhead interface on top of this.

### DIFF
--- a/juniper/src/parser/utils.rs
+++ b/juniper/src/parser/utils.rs
@@ -52,15 +52,15 @@ impl Span {
 
 /// Data structure used to wrap items into a [`Span`].
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
-pub struct Spanning<T> {
+pub struct Spanning<T, Sp = Span> {
     /// Wrapped item.
     pub item: T,
 
     /// [`Span`] of the wrapped item.
-    pub span: Span,
+    pub span: Sp,
 }
 
-impl<T> Spanning<T> {
+impl<T> Spanning<T, Span> {
     #[doc(hidden)]
     pub fn new(span: Span, item: T) -> Self {
         Self { item, span }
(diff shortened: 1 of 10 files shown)


juniper/src/parser/utils.rs
```diff
@@ -52,15 +52,15 @@ impl Span {
 
 /// Data structure used to wrap items into a [`Span`].
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
-pub struct Spanning<T> {
+pub struct Spanning<T, Sp = Span> {
```
audunhalland, line 55: The added generic argument probably influences how this is documented, but not sure.

The point of the generic is to avoid making a separate type like `BorrowedSpanning`, but I'm up for improvement suggestions.

The ability to borrow a Span can lead to slightly more performant code in various places. Currently only in `look_ahead`, other future suggestions are e.g. the input validation rules.

juniper/src/schema/meta.rs
```diff
@@ -202,7 +202,7 @@ pub struct Argument<'a, S> {
     #[doc(hidden)]
     pub arg_type: Type<'a>,
     #[doc(hidden)]
-    pub default_value: Option<InputValue<S>>,
+    pub default_value: Option<Spanning<InputValue<S>>>,
```
tyranron: @audunhalland I'm not sure whether this should be `Spanning`, because, judging by code, it always has `Spanning::unlocated()` value. Keeping `Spanning` as a part of type here, may give assumptions that the field is actually `Span`ned here properly, in a way.

Better remove it, and do the `Spanning::unlocated()` wrapping explicitly in the places where this is required.

tyranron (approved):
@audunhalland thanks!

PR #1210, opened 2023-11-09
Remove `iron` integration

tyranron:
This PR removes `juniper_iron` crate from workspace, and thus, support for the [`iron` crate] for future `juniper` versions.

## Motivation

[`iron` crate] seems to become unmaintained and unsupported for a while, confirmed by its authors:
- https://github.com/iron/iron/issues/643#issuecomment-1282016240
- https://github.com/iron/iron/issues/635#issuecomment-797584494

It also depends on quite old versions, introducing regular problems to CI.

## Support criteria

Once [`iron` crate] becomes actively maintained again and supports latest versions of its dependencies, it would be easy to restore `juniper_iron` crate back into the workspace and maintain it.

[`iron` crate]: https://docs.rs/iron


tyranron:
@LegNeato before merging this, I would like to hear any comments of yours on this, if you have ones to share.

LegNeato:
I think this is the right call! I doubt anyone is using iron anymore, and if they are they can use old versions or add a pr and sign up to support it themselves.

PR #1212, opened 2023-11-12
Lazy look_ahead

audunhalland:
I've noticed that when using the look-ahead feature, this always involves a full traverse/transform of the input document under that node. This can be a problem when the input documents are large and complex.

This is an attempt to rewrite the `look_ahead` API so that its invocation becomes basically free. Instead of doing a forced deep transformation up front, the required work has been moved into the methods explicitly called by the API user.

A single level of `LookAheadChildren` gets computed when the user calls `LookAheadSelection::children()`.
The arguments/values to a field are also lazily computed, and variable substitution is now done purely on demand.

The PR is in a draft state and the old `look_ahead` API is still intact for the time being, I'm interested in general feedback first before finishing it off.

### Summary of API breakages:
* `trait LookAheadMethods` removed. In my opinion there's no need for this and it always requires an explicit extra import.
* `struct ConcreteLookAheadSelection` removed
* `struct LookAheadArguments` removed, replaced by an opaque `impl Iterator`
* `struct LookAheadSelection` now has some inherent methods at least partially compatible with `LookAheadMethods`:
  * `field_name`, `has_arguments`: Compatible signature
  * `arguments` now returns `impl Iterator<Item = LookAheadArgument>` instead of `&[LookAheadArgument]`.
  * `argument` now returns `Option<LookAheadArgument>` instead of `Option<&LookAheadArgument>`.
  * `children` now returns `LookAheadChildren` instead of `Vec<&Self>`. `LookAheadChildren` implements `IntoIterator`.
  * `field_original_name` / `field_alias`: New in `0.16`
* Some methods moved into new struct `LookAheadChildren`:
  * `LookAheadMethods::child_names`
  * `LookAheadMethods::has_children` moved to `LookAheadChildren::is_empty`
  * `LookAheadMethods::select_child` moved to `LookAheadChildren::select`
* `LookAheadSelection::for_explicit_type` moved to `LookAheadSelection::children_for_explicit_type`, which returns `LookAheadChildren`.
* `LookAheadValue::List(_)` contains a new struct `LookAheadList` with `IntoIterator` + `iter` method.
* `LookAheadValue::Object(_)` contains a new struct `LookAheadObject` with `IntoIterator` + `iter` method.


juniper/src/executor/look_ahead_lazy.rs
```diff
@@ -0,0 +1,1555 @@
...
+/// A JSON-like list that can be used as an argument in the query execution.
+#[derive(Clone, Copy, Debug)]
+pub struct LookAheadList<'a, S> {
+    input_list: &'a [Spanning<InputValue<S>>],
+    vars: Option<&'a Variables<S>>,
+}
+
+impl<'a, S: ScalarValue> LookAheadList<'a, S> {
+    /// Returns an iterator over the list's elements.
+    pub fn iter(&self) -> impl Iterator<Item = BorrowedSpanning<'a, LookAheadValue<'a, S>>> + '_ {
```
audunhalland, line 73: These wrapper types should ideally have `IntoIterator` implementations I suppose, but I don't think the added boilerplate justifies it currently. I think it can maybe be added later when/if Rust gets the ability to write shorter IntoIterator implementations. Especially I think exposing a new public type for every iterator is a little painful.

juniper/src/executor/look_ahead_lazy.rs
```diff
@@ -0,0 +1,1555 @@
...
+    }
+
+    /// Returns the (potentially aliased) name of the field, represented by the current selection.
+    pub fn field_name(&self) -> &'a str {
+        self.field_alias()
+            .unwrap_or_else(|| self.field_original_name())
+    }
+
+    /// Returns the top level arguments from the current selection.
+    pub fn arguments(&self) -> Option<LookAheadArguments<'a, S>> {
```
audunhalland, line 310: This now encodes whether arguments are present or not, but I'm not sure whether this is necessary. The point was to meaningfully replace `fn LookAheadMethods::has_arguments`.

juniper/src/executor/look_ahead_lazy.rs
```diff
@@ -0,0 +1,1555 @@
...
+        }
+    }
+
+    /// Returns the children from the current selection.
+    pub fn children(&self) -> LookAheadChildren<'a, S> {
+        self.build_children(Applies::All)
+    }
+
+    /// Returns the children from the current selection that only applies to a specific type.
+    pub fn children_for_explicit_type(&self, type_name: &str) -> LookAheadChildren<'a, S> {
```
audunhalland, line 331: This is the replacement for the `ConcreteLookAheadSelection` struct. I didn't fully understand why that exists as a separate type, this simplifies things IMO.

The largest difference between the old and new code is that the type filter is not "persistent" anymore. But I think filtering fields for a specific type is most meaningful at a specific document level anyway.

juniper/src/executor/mod.rs
```diff
@@ -743,6 +746,55 @@ where
             })
     }
 
+    /// Construct a lookahead selection for the current selection.
+    ///
+    /// This allows seeing the whole selection and perform operations
+    /// affecting the children.
+    pub fn look_ahead_lazy(&'a self) -> look_ahead_lazy::LookAheadSelection<'a, S> {
```
audunhalland, line 753: The replacement for `Executor::look_ahead`. 

As far as I can see, there are no tests for `Executor::look_ahead`. So I don't know whether the `FragmentSpread` fallback branch at the bottom actually works or not. The first branch where it finds the field works, I have tested this manually.

juniper/src/executor/look_ahead_lazy.rs
```diff
@@ -0,0 +1,1555 @@
...
+/// The arguments passed into a query.
+#[derive(Copy, Clone, Debug)]
+pub struct LookAheadArguments<'a, S> {
+    arguments: &'a Arguments<'a, S>,
+    vars: &'a Variables<S>,
+}
+
+impl<'a, S> LookAheadArguments<'a, S> {
+    /// Returns an iterator over the arguments.
+    pub fn iter(&self) -> impl Iterator<Item = LookAheadArgument<'a, S>> + '_ {
```
audunhalland, line 183: It might be odd that `LookAheadChildren` implements `IntoIterator` but `LookAheadArguents` does not.

The reason is that the former owns its data, but the latter uses a slice with a `.map` operation, and we'd need a new type for the `type IntoIter`.

juniper/src/executor/look_ahead_lazy.rs
audunhalland: This could change into ` -> impl Iterator<Item = LookAheadArgument>`. But then does `has_arguments` need to be added back? One could achieve the same with ortogonal features  `look_ahead.arguments().next().is_some()`.

audunhalland:
I know the change looks large, I made it so because I essentially see it as a rewrite. I would like to get some indication on whether you would accept a change along these lines. If you are positive, I'll finish up the PR, overwrite the existing implementation and update the integration tests. I've finished this locally.

audunhalland:
@tyranron Thanks! If you instead prefer a PR closer to the finished product, I can fix that before that time.

juniper/src/executor/mod.rs
tyranron: @audunhalland 

> edit: I found the tests, there are a lot! But testing just executor.look_ahead(); after it turns lazy is arguably a bit weaker. I suggest at least adding a level of .children(); to the integration tests.

Will be happy to see this 👍

juniper/src/executor/look_ahead_lazy.rs
tyranron: @audunhalland there are some inconsistencies, though:
- `LookAheadArguments::iter()` returns `Iterator<Item = LookAheadArgument<'a, S>>` over owned values.
- `LookAheadChildren::iter()` returns `Iterator<Item = &LookAheadSelection<'a, S>>` over references.
- `<LookAheadChildren as IntoIterator>::into_iter()` returns `Iterator<Item = LookAheadSelection<'a, S>>` over owned values.

I suggest that `T::iter()` inherent method and `<&T as IntoIterator>::into_iter()` should always return `Iterator<Item = &T>` over references.

For owned values `Iterator<Item = T>` there should be either `<T as IntoIterator>::into_iter()` impl or  `T::into_iter()` inherent method.

juniper/src/executor/look_ahead_lazy.rs
```diff
@@ -0,0 +1,1561 @@
...
+        ),
+    > {
+        let vars = self.vars;
+        self.input_object.iter().map(move |(key, val)| {
+            (
+                Spanning {
+                    span: &key.span,
+                    item: key.item.as_str(),
+                },
+                LookAheadValue::from_input_value(&val.item, &val.span, vars),
```
tyranron, line 120: @audunhalland so, this is the place where the actual laziness happens. Instead of traversing and converting the whole AST on the root value, we do it in the actual iteration.
However, it seems that it happens each time we do this iteration. So, if we repeatedly work in the same method on the same AST node multiple times, we do parse/convert AST nodes multiple times too.

I wonder whether it can introduce any performance issues to the level we should implement memoization out-of-the-box?

tyranron (changes requested):
@audunhalland the overall redesign looks interesting, and I'm much in favor to lazify things. Moreover, I'd like having it for the upcoming 0.16 release, so its benefirs could be [described in the Book](https://github.com/graphql-rust/juniper/pull/1230).

However, some API decisions should be thought, discussed and bikeshedded a little bit more, as I don't have strong opinion about them yet.

juniper/src/executor/look_ahead_lazy.rs
```diff
@@ -0,0 +1,1561 @@
...
+    Scalar(&'a S),
+    Enum(&'a str),
+    List(LookAheadList<'a, S>),
+    Object(LookAheadObject<'a, S>),
+}
+
+impl<'a, S: ScalarValue + 'a> LookAheadValue<'a, S> {
+    fn from_input_value(
+        input_value: &'a InputValue<S>,
+        span: &'a Span,
```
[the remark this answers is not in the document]
audunhalland: Mainly because `input_value` and `span` do not always come from the same source (because of `InputValue::Variable`, where this is re-entrant). Changing the argument to `BorrowedSpanning` will look less elegant at some callsites. Just keep them as two separate parameters. Also this is not an public function so i guess this isn't a big issue.

juniper/src/executor/look_ahead_lazy.rs
audunhalland: It's not so easy to return iterators over references when the iterated item is lazily computed. We're iterating over things that _wrap_ references. `LookAheadArguments` has `arguments: &Arguments` internally, but this isn't exposed directly. Instead each argument gets turned into a `LookAheadArgument` lazily as we iterate over them. If we wanted to iterate over `&LookAheadArgument` references, we'd have to preallocate a collection of those, which to me just seems wasteful without any particular gain.

juniper/src/executor/look_ahead_lazy.rs
audunhalland: `fn children()` is an eager computation that collects into a vector. That's why I made this alternative API `children_for_explicit_type`: It also collects into a vector, only with a stricter filter.

`children().for_explicit_type("Person")` would allocate a vector and then filter that into a new vector again, so I would consider this API less "Rusty" in the sense that it doesn't encourage the cheapest possible way to perform some computation.

juniper/src/executor/look_ahead_lazy.rs
audunhalland: I agree, I'll reintroduce `has_arguments` and do `arguments() -> LookAheadArguments`.

juniper/src/executor/look_ahead_lazy.rs
```diff
@@ -0,0 +1,1561 @@
...
+    }
+
+    /// Returns the child selection for the specified field.
+    ///
+    /// If a child has an alias, it will only match if the alias matches the specified `name`.
+    pub fn select(&self, name: &str) -> Option<LookAheadSelection<'a, S>> {
+        self.children
+            .iter()
+            .find(|child| child.field_name() == name)
+            .cloned()
```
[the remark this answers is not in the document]
audunhalland: I don't see how, unless `SelectionSource` and `LookAheadSelection` have _manual_ implementations of `Clone` and `Copy` (instead of `#[derive]`). The problem with the derive is with the generic parameter `S: ScalarValue`; Copy gets conditionally implemented given `S: Copy`, which is false.

juniper/src/executor/look_ahead_lazy.rs
audunhalland: I interpret this mapping as very cheap. There's no allocation involved; it involves an enum `match` and the most expensive thing that can happen is one HashMap lookup in case you're looking at a Variable. The rest is just plain reference copying.

It's a completely shallow mapping. Sub-iteration of the contents of lists and objects is something the API user opts in to.

juniper/src/executor/look_ahead_lazy.rs
audunhalland: Done.

juniper/src/executor/mod.rs
audunhalland: I will fix all the tests after (if!) we reach agreement on an API and the old look_ahead has been removed.

juniper/src/executor/look_ahead_lazy.rs
tyranron: @audunhalland 

> It's not so easy to return iterators over references when the iterated item is lazily computed.

I do understand this concern. My comment is about naming only. If we cannot return iterators over references, then let's just name the methods `into_iter()` or impl the `IntoIterator` trait instead.

juniper/src/executor/look_ahead_lazy.rs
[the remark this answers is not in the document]
tyranron: @audunhalland yup, `std` derives are awfully dumb when it comes to generics, so providing a manual `Copy` impl is totally OK here.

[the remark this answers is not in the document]
tyranron: Fixed in https://github.com/graphql-rust/juniper/pull/1212/commits/149e61f3b52226121df03f43f3cd377315368b17.

juniper/src/executor/look_ahead_lazy.rs
tyranron: @audunhalland thinking about it a little bit more, I agree. Maybe we should just leave a notice in documentation about this, so the library user can memoize the result into a temporary binding, if he does a lot of repeated stuff and needs to.

juniper/src/executor/look_ahead_lazy.rs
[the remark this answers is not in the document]
tyranron: @audunhalland well, the only place where they come not from the same source is [mapping an ` `InputValue::Variable`](https://github.com/graphql-rust/juniper/blob/149e61f3b52226121df03f43f3cd377315368b17/juniper/src/executor/look_ahead_lazy.rs#L49-L52), which is somewhat a special case regarding that we expand the variable. I think `BorrowedSpanning` cover the majority of cases better, and makes it much clearer that the provided `Span` is related to the `input_value`.

juniper/src/executor/look_ahead_lazy.rs
tyranron: @audunhalland 

> edit: That said, even `fn children()` could _theorectically_ return an iterator of lazily computed children without any vector allocation, but since there is no generator/coroutine in Rust yet I considered the implementation too complex.

Looked into details, and agree that it would be too complex for the sake of decoupled methods.

Let's go with `children_for_explicit_type()` then, and leave a `// TODO:` comment in hope for generators stabilization.

audunhalland: Thinking about it again, `fn children()` would always require allocation in Rust, since it involves potentially unbounded recursion (visiting FragmentSpreads that could refer to other FragmentSpreads, is that a valid model in GraphQL)?

The way it's done how involves recursion using the stack, and pushing the results to the end of a vector. A Rust generator would be stackless, turning things around: To support recursion it would need to use the heap for _that_, while yielding the actual items would be lazy. We'd be forced to use heap allocation in either case.

juniper/src/executor/look_ahead_lazy.rs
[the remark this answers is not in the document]
audunhalland: Do you think the parameter should be a `BorrowedSpanning` or a `&BorrowedSpanning`? Sometimes we have only a `&BorrowedSpanning<InputValue>` that we want to map to a LookAheadValue. This is another reason it makes more sense to me to accept two individual borrowed parameters rather than one struct. The struct approach seems more shoe-horned.

juniper/src/executor/look_ahead_lazy.rs
audunhalland: Regarding `LookAheadChildren::iter(&self)` and that it yields borrowed `&LookAheadSelection` (this being inconsistent with some of the other `iter(&self)` iterators); this inconsistency is a reflection of how lazy_look_ahead is currently implemented. `LookAheadSelection` is not cheaply clonable because it owns a `Vec`. In Rust it's not considered bad to leak implementation details about how memory is managed internally. I of course agree that it makes it harder to refactor details later without breaking any public API, but this is just Rust.

juniper/src/executor/look_ahead_lazy.rs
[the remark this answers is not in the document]
tyranron: @audunhalland seems that just `BorrowedSpanning` works totally fine due to `Spanning::as_ref`: https://github.com/graphql-rust/juniper/pull/1212/commits/cecee4b3cc81596ec16e61c63ac99589c69c3c4d

juniper/src/executor/look_ahead_lazy.rs
tyranron: @audunhalland unfortunately we don't have exact this situation described in [guidelines](https://rust-lang.github.io/api-guidelines/naming.html#methods-on-collections-that-produce-iterators-follow-iter-iter_mut-into_iter-c-iter). Your vision that `iter`/`iter_mut` is about notating receiver rather than emitted values seems quite reasonable and logic.

However, it still feels quite odd to me to expect owned values as the result of `.iter()`. I don't know why. Usually, I expect some other naming like `.iter_values()` or just `.values()` notating exactly that some computation happens, and something other is produced. However, here it makes no sense, since the `LookAheadArguments` type itself means iteration over `LookAheadArgument`. There is no "other computed stuff".

Something like `.iter_owned()` is odd and not quite expected from the user-side too.

What if we impl `IntoIterator for &LookAheadArguments` with `type Item = LookAheadArgument`?

It should work relatively the same in method chains. As a bonus, it allows us `for arg in &args {}`.

audunhalland: Suggestion for middle ground: `LookAheadValue`, `LookAheadList` and `LookAheadObject` all implement `Copy`, and there is no `iter(&self)` method at all on these types, only `IntoIterator` implementations for `Self` and `&Self`:

```
match look_ahead_value {
    LookAheadValue::List(list) => {
        list.into_iter().map(|value| { /* read values */ })
    }
    ...
}
```

The drawback is that this style would _likely_ be more surprising to the user. As we can see with the [clippy lint](https://github.com/rust-lang/rust-clippy/issues/1565) that suggests changing `.into_iter()` to `.iter()` if there is no actual ownership transfer going on, users are taught to reach for `.iter()` first, and a _GraphQL look ahead_ context is quite clearly a _borrow only_ context.

I'd think "hmm weird, why do I have to call .into_iter(), my mental model expects `.iter()` here". but I'd probably accept the situation and just move on with my life :)

Iteration methods can always be added at a later stage if users turn out to find this confusing. Naming suggestion in that case:
* `fn LookAheadList::elements(&self)`
* `fn LookAheadObject::entries(&self)` (although this might be confusing as I would think of Rust's Map entry API.)

tyranron:
@audunhalland once you finish adjustments, please, re-request my review via this GitHub button. Thanks!

Screenshot 2024-01-15 at 18 03 06

juniper/src/executor/look_ahead_lazy.rs
audunhalland: Note about this added in docs for the two iterator types in https://github.com/graphql-rust/juniper/pull/1212/commits/c100140091fb8dbf33ec3396ef4fd8b7bc129d4e.

audunhalland:
Ok, rebased on master, overwrote `look_ahead` and updated the tests I found to be relevant (based on grepping for `look_ahead`).

tyranron (approved):
@audunhalland thank you for the hard work on this!

PR #1213, opened 2023-11-14
fix: Three tests because of updated compiler(?)

audunhalland:
In #1212 there are three test failures unrelated to the change. Unsure why they fail, but my guess is that a newer rustc has a change in error reporting.


tyranron:
@audunhalland yeah, sorry. Compiler output is quite unstable even for stable Rust 🙃

Let's consider these changes in https://github.com/graphql-rust/juniper/pull/1212 then. No need for a separate PR here.

PR #1226, opened 2023-11-27
GraphQL over HTTP spec compliance (#1120)

### LINKED ISSUE #1120: Make sure we are compliant with new graphql over http spec
Currently in draft but it is an official project:

https://graphql.github.io/graphql-over-http/draft/

tyranron:
Resolves #1120 

## Synopsis

> Currently in draft but it is an official project:
> 
> https://graphql.github.io/graphql-over-http/draft/

### DIFF
--- a/juniper_axum/README.md
+++ b/juniper_axum/README.md
@@ -4,7 +4,8 @@
 [![Crates.io](https://img.shields.io/crates/v/juniper_axum.svg?maxAge=2592000)](https://crates.io/crates/juniper_axum)
 [![Documentation](https://docs.rs/juniper_axum/badge.svg)](https://docs.rs/juniper_axum)
 [![CI](https://github.com/graphql-rust/juniper/workflows/CI/badge.svg?branch=master "CI")](https://github.com/graphql-rust/juniper/actions?query=workflow%3ACI+branch%3Amaster)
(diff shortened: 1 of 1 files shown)


tyranron:
@LegNeato after investigating it closely, the rabbit hole seems to be way deeper that I thought it would be. There are quite a lot of nuisances and details to be handled properly regarding different media types (for example, response codes for `application/json` and `application/graphql-response+json` are treated very differently), and moreover, spec implies few points for extension by user-side, while the core should remain.

Implementing all this separately for each integration crate would be painful and fragile. Moreover, integration crates are not doing well with extensibility. That's why, to implement this properly, it seems right to introduce a new `juniper_graphql_http` crate, providing a core `http`-crate-based implementation and being framework-agnostic, just as we have the `juniper_graphql_ws` crate for GraphQL over WebSocket specs.

This, however, quite a huge and time-consuming task, so I decided to move it from [0.16 milestone](https://github.com/graphql-rust/juniper/milestone/4) to [0.17 one](https://github.com/graphql-rust/juniper/milestone/5).

LegNeato:
Make sense to me! I've been thinking that we'd probably need something similar, mirroring the http / tower stack

PR #1230, opened 2023-12-01
Rework Book

tyranron:
Resolves #234, #444, #679, #867, #974, #1029, #1056, #1168   
Follows #518

## Synopsis

Juniper Book is outdated and requires massive rework.

## Solution

- [x] Rework and update existing chapters.
    - [x] Mention correct case transformation for GraphQL enums. (#1029)
    - [x] Provide better docs for dataloader and similar techniques. (#444)
    - [x] Cover look-ahead feaures. (#234)
- [x] Mention `juniper_actix` and `juniper_axum` in Book. (#867)
- [x] Fix all links leading to Book. (#974)
- [x] Remove old version of Book from GitHub Pages root and place the redirect instead. (#1168)

### Reworked chapters

- [x] Introduction
- [x] Quickstart
- [x] Type System
  - [x] Objects
    - [x] Complex fields
    - [x] Context
    - [x] Error handling
  - [x] Interfaces
  - [x] Unions
  - [x] Enums
  - [x] Input objects
  - [x] Scalars
- [x] Schemas and mutations
- [x] Adding A Server
  - [x] Official Server Integrations
  - [x] Third Party Integrations
- [x] Advanced Topics
  - [x] Introspection
  - [x] Non-struct objects
  - [x] Implicit and explicit null
  - [x] Objects and generics
  - [x] Multiple operations per request
  - [x] Dataloaders
  - [x] Subscriptions

PR #1232, opened 2023-12-11
Allow top-level `#[graphql]` attribute in attribute macros

tyranron:
Required for #1230 

Allows to use `#[graphql]` attributes near `#[graphql_object]` and similar.


tyranron:
CI failure seems to be `bson` crate regression and is not the topic of this PR.

tyranron:
> CI failure seems to be `bson` crate regression and is not the topic of this PR.

The fix: https://github.com/mongodb/bson-rust/pull/445

PR #1237, opened 2024-01-12
Stable order in generated schema's SDL (#1134)

### LINKED ISSUE #1134: Consistent order of definitions in Schema::as_schema_language output
Definitions in SDL created by `Schema::as_schema_language` appear to be sorted semi randomly. We are using this method to print SDL to a file, and this behaviour makes changes to the schema harder to review, and to navigate the file in general.

**Proposed solution**

Sort definitions in the following order:
1. Query type
2. Mutation type, if any
3. Subscription type, if any
4. All other types, sorted alphabetically

**Additional context**

Currently, the order persists between compilations as long as there are no changes to schema, but a change to one item commonly results in multiple unrelated items changing their position.

tyranron:
Resolves #1134

## Synopsis

See #1134 for details:
> Currently, the order persists between compilations as long as there are no changes to schema, but a change to one item commonly results in multiple unrelated items changing their position.

## Solution

The solution is based on the https://github.com/graphql-rust/juniper/issues/1134#issuecomment-1704425425 approach, proposed and implemented by @michael-groble for the [`graphql-parser`] directly.

We ship it without waiting for a [`graphql-parser`], and later can switch to the solution provided by the [`graphql-parser`] out-of-the-box.

## Additionally

This PR removes the `graphql-parser` Cargo feature of `juniper`, because it effectively exposed the `RootNode::as_parser_document()` method only, hardly overlapping with the `schema-language` Cargo feature exposing `RootNode::as_schema_language()` method only.

Since they're very similar, now it's only the `schema-language` Cargo feature is left, exposing both `RootNode::as_sdl()` and `RootNode::as_document()` methods.

## Checklist

- [x] Documentation is updated.
- [x] Tests are updated.
- [x] Changelog entry is added.

[`graphql-parser`]: https://docs.rs/graphql-parser


tyranron:
@LegNeato please look whether you have any comments on this.

PR #1239, opened 2024-01-17
Stable order in introspection output

audunhalland:
The type list sent from juniper through introspection to `GraphiQL` has random (but deterministic) order currently (based on FNV string hash of the type name).

Inspired by https://github.com/graphql-rust/juniper/pull/1237, I was interested in improving this.

This PR applies a sort order quite similar to the SDL sorter. The introspection data has a slightly different layout, but comparable.

To verify that the order is nice, stable and deterministic I removed the _test sorter_ from the `schema_introspection` test.

### DIFF
--- a/juniper/src/schema/schema.rs
+++ b/juniper/src/schema/schema.rs
@@ -282,11 +282,11 @@ impl<'a, S: ScalarValue + 'a> TypeType<'a, S> {
             TypeType::Concrete(&MetaType::Interface(InterfaceMeta {
                 name: ref iface_name,
                 ..
-            })) => Some(
-                context
-                    .concrete_type_list()
-                    .iter()
-                    .filter_map(|&ct| {
+            })) => {
+                let mut type_names: Vec<_> = context
+                    .types
+                    .values()
+                    .filter_map(|ct| {
                         if let MetaType::Object(ObjectMeta {
                             name,
                             interface_names,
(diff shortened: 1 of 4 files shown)


juniper/src/schema/schema.rs
```diff
@@ -282,11 +282,11 @@ impl<'a, S: ScalarValue + 'a> TypeType<'a, S> {
...
                 ..
-            })) => Some(
-                context
-                    .concrete_type_list()
-                    .iter()
-                    .filter_map(|&ct| {
+            })) => {
+                let mut type_names: Vec<_> = context
+                    .types
+                    .values()
```
audunhalland, line 288: This is an optimization. `concrete_type_list()` collects into a vector first, which is wasteful here.

juniper/src/schema/schema.rs
```diff
@@ -295,15 +295,22 @@ impl<'a, S: ScalarValue + 'a> TypeType<'a, S> {
...
+                                .any(|iname| iname == iface_name)
+                                .then(|| name.as_ref())
                         } else {
                             None
                         }
                     })
-                    .collect(),
-            ),
+                    .collect();
+                type_names.sort();
```
audunhalland, line 305: This just sorts a `Vec<&str>` of type names, since they are all Object types there is no need to care about fundamental type order.

tyranron (approved):
@audunhalland thanks!

PR #1240, opened 2024-01-17
refactor: Use auto_enums instead of futures-enum

audunhalland:
`auto_enums` is from the same author as `futures-enum` ([taiki-e](https://github.com/taiki-e)), and `auto_enums` appears more up-to-date. But the most important thing is that this gets rid of `syn 1` from juniper's non-dev dependencies.

### DIFF
--- a/juniper/Cargo.toml
+++ b/juniper/Cargo.toml
@@ -49,13 +49,13 @@ uuid = ["dep:uuid"]
 [dependencies]
 anyhow = { version = "1.0.47", optional = true }
 async-trait = "0.1.39"
+auto_enums = "0.8"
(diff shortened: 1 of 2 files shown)


audunhalland:
On second thought, `impl Future` for this enum can be done manually in about 20 lines of code. Doing this instead will shrink the number of dependencies. I see juniper already uses some `unsafe`.

tyranron (approved):
@audunhalland thanks! 🍻

I think `auto_enums` is a good enough way to go, at the moment. Shrinking dependencies is a major topic and needs careful revisiting and renewing the whole codebase first. Eventually we'll get there.

PR #1242, opened 2024-01-23
refactor: Avoid copying Span when evaluating rules

audunhalland:
As mentioned in https://github.com/graphql-rust/juniper/pull/1209#discussion_r1388107123, making `Spanning` generic over `Sp` allows avoiding copying 6 `usize`s (the `Span`) in places where a borrowed Spanning needs to be type mapped. I.e. the `Document` contains `Spanning<InputValue>`, but `InputValue` is an enum, so when the spanned item's type gets mapped into a reference to each variant, there's no need to transform `&Spanning<T, Span>` to `Spanning<&U, Span>`. It can instead be transformed to a `Spanning<&U, &Span>`.

This applies to the `Visitor` trait and its methods related to the visitation of `InputValue`s. The `Visitor` trait is public, but `#[doc(hidden)]`.

### DIFF
--- a/juniper/src/validation/multi_visitor.rs
+++ b/juniper/src/validation/multi_visitor.rs
@@ -6,6 +6,7 @@ use crate::{
     parser::Spanning,
     validation::{ValidatorContext, Visitor},
     value::ScalarValue,
+    Span,
 };
 
 #[doc(hidden)]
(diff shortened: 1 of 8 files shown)


tyranron (approved):
@audunhalland thanks!

PR #1243, opened 2024-01-26
Use the prod distribution of React in graphiql.html

kiendang:
to reduce bundle size


tyranron (changes requested):
@kiendang thank you!

However, this file is [auto-downloaded from the upstream repo](https://github.com/graphql-rust/juniper/blob/f213d5193c3c0d3bae194403c9867f077e0b5004/juniper/Makefile#L38-L39), and so:
- either better contribute this change to the upstream repo;
- or [add post-processing commands](https://github.com/graphql-rust/juniper/blob/f213d5193c3c0d3bae194403c9867f077e0b5004/juniper/Makefile#L40-L49) doing the substitution.

kiendang:
Ah I see. I actually did just that a while back, opening a PR in the graphiql repo graphql/graphiql#3402 but it got closed and I didn't really understand the reason.

tyranron:
@kiendang let's adjust post-processing in `Makefile` then.

PR #1256, opened 2024-05-04
Refactor code to remove async move closure which may result in double…

### LINKED ISSUE #820: Create a custom future to prevent double boxing
We should fix this TODO:

https://github.com/graphql-rust/juniper/blob/2c15ea798c1bd7df5bede24495fba541c1c84b2d/juniper/src/types/async_await.rs#L264

jawang94:
Closes #820 

I did not detect any apparent double boxing issues here however I refactored the code to use the custom `AsyncValueFuture` enum and eliminate the unnecessary `async move` closure. This is my first contribution here - all `cargo test`s passed however the `test.book` failed with some strange error about the `dataloader` crate which seems unrelated. 

P.S. I am new to rust and snooping around for real-world practice so please advise me if I've made mistakes or missed something! Thank you in advance and hope to contribute more in the future!

### DIFF
--- a/juniper/src/types/async_await.rs
+++ b/juniper/src/types/async_await.rs
@@ -258,31 +258,29 @@ where
                 let is_non_null = meta_field.field_type.is_non_null();
 
                 let response_name = response_name.to_string();
-                async_values.push_back(AsyncValueFuture::Field(async move {
-                    // TODO: implement custom future type instead of
-                    //       two-level boxing.
-                    let res = instance
-                        .resolve_field_async(info, f.name.item, &args, &sub_exec)
-                        .await;
+                let res = instance
+                    .resolve_field_async(info, f.name.item, &args, &sub_exec)
+                    .await;
 
-                    let value = match res {
-                        Ok(Value::Null) if is_non_null => None,
-                        Ok(v) => Some(v),
-                        Err(e) => {
-                            sub_exec.push_error_at(e, pos);
-
-                            if is_non_null {
-                                None
-                            } else {
-                                Some(Value::null())
-                            }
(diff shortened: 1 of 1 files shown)


juniper/src/types/async_await.rs
```diff
@@ -258,31 +258,29 @@ where
...
-                            sub_exec.push_error_at(e, pos);
-
-                            if is_non_null {
-                                None
-                            } else {
-                                Some(Value::null())
-                            }
+                let res = instance
+                    .resolve_field_async(info, f.name.item, &args, &sub_exec)
+                    .await;
```
audunhalland, line 263: It looks like this `.await` breaks concurrency. The await now happens directly within the `for` loop body, so the future will pause at this point before processing the next select. The intention of `async_values` is that the loop should produce all the futures synchronously up front, then after this loop there is another `while` loop at the bottom of the function which polls the `FuturesOrdered` so that all its contained futures may progress _simultaneously_.

juniper/src/types/async_await.rs
```diff
@@ -258,31 +258,29 @@ where
                 let is_non_null = meta_field.field_type.is_non_null();
 
                 let response_name = response_name.to_string();
-                async_values.push_back(AsyncValueFuture::Field(async move {
-                    // TODO: implement custom future type instead of
-                    //       two-level boxing.
```
audunhalland, line 263: I don't see any boxing at all.. `AsyncValueFuture` is generic so this comment looks outdated.

LegNeato (changes requested):
See the comments from @audunhalland

PR #1263, opened 2024-06-15
Add support for `Request<T>` where `T` is a type that implements `hyper::body::Body` trait

### LINKED ISSUE #1102: Possible API improvements to `juniper_hyper`
While working on #1101 , I thought of a couple more possible improvements. 

For one, I think it would be useful to make `parse_req` and `execute_request[_sync]` public. With that, it would be possible to inspect the `GraphQLRequest` before executing it (e.g. for logging). It's even questionable whether the convenience `graphql[_sync]` functions are even necessary then, as they just call these two other functions. 

All functions taking a `Request<Body>` should also be able to take a `Request<String>`. The body is converted to a string internally anyway and users of the library might already have a `Response<String>`. This also helps with users inspecting the request beforehand, given that `String` is much easier to use than `hyper::Body`. However, extending the API like that is a bit harder than #1101 as we still want to support `Body`.

elcharitas:
Should resolve #1102 

> ## Problem
> Attempts to use any other hyper request type aside `Request<hyper::body::Incoming>` fails with a compile-time error of type mismatch

> ## Solution
> Add generic support for valid supported hyper `Request` or any custom `Request` that has a body which implements `hyper::body::Body`
> 
> ## Checklist
> * [x]  Tests are added
> * [x]  CHANGELOG entry is added
> * [x]  ~Documentation is updated~ (not required)

### DIFF
--- a/juniper_hyper/src/lib.rs
+++ b/juniper_hyper/src/lib.rs
@@ -4,7 +4,7 @@ use std::{error::Error, fmt, string::FromUtf8Error, sync::Arc};
 
 use http_body_util::BodyExt as _;
 use hyper::{
-    body,
+    body::Body,
     header::{self, HeaderValue},
     Method, Request, Response, StatusCode,
 };
@@ -15,14 +15,7 @@ use juniper::{
 use serde_json::error::Error as SerdeError;
 use url::form_urlencoded;
 
-pub async fn graphql_sync<
-    CtxT,
-    QueryT,
-    MutationT,
-    SubscriptionT,
-    S,
-    T: body::Body<Error = impl Error + 'static>,
->(
+pub async fn graphql_sync<CtxT, QueryT, MutationT, SubscriptionT, S, T>(
     root_node: Arc<RootNode<'static, QueryT, MutationT, SubscriptionT, S>>,
     context: Arc<CtxT>,
     req: Request<T>,
@@ -36,6 +29,7 @@ where
     SubscriptionT::TypeInfo: Sync,
     CtxT: Sync,
     S: ScalarValue + Send + Sync,
+    T: Body<Error: fmt::Display>,
 {
     match parse_req(req).await {
         Ok(req) => execute_request_sync(root_node, context, req).await,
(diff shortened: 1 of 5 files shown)


elcharitas:
CI Checks seem to fail and this is weird since the issues arise from the `juniper` crate.

Any pointers?

LegNeato (changes requested):
Thank you for the PR, sorry for the slow response! Are there backward compat considerations here? I see you have a default type but I am not sure. Also, can you add a test so we can be sure this doesn't regress?

elcharitas:
> Thank you for the PR, sorry for the slow response! Are there backward compat considerations here? I see you have a default type but I am not sure. Also, can you add a test so we can be sure this doesn't regress?

Thanks for checking in. To answer your concerns..

> Are there backward compat considerations here?

Yes, this PR is fully backward compatible. How do we know? There's a test already included here and it passes with no need for further configuration.

> can you add a test so we can be sure this doesn't regress

Yes I can add a test for custom `Request<T>` type. However, I must point out that this PR makes no algo change whatsoever, it only allows passing generic request types making juniper compatible with different request types.

juniper_hyper/src/lib.rs
```diff
@@ -419,10 +448,30 @@ mod tests {
...
                                         Ok::<_, Infallible>(if matches {
-                                            if is_sync {
-                                                super::graphql_sync(root_node, db, req).await
+                                            if is_custom_type {
+                                                let (parts, mut body) = req.into_parts();
+                                                let body = {
+                                                    let mut buf = String::new();
+                                                    if let Some(Ok(frame)) = body.frame().await {
+                                                        if let Ok(bytes) = frame.into_data() {
+                                                            buf = String::from_utf8_lossy(&bytes)
```
elcharitas, line 457: I'm choosing to use `String`, since that's the quickest sample that came to mind without having to create something new. However, this would work just about the same for other types.

juniper_hyper/src/lib.rs
```diff
@@ -376,8 +401,12 @@ mod tests {
         }
     }
 
-    async fn run_hyper_integration(is_sync: bool) {
-        let port = if is_sync { 3002 } else { 3001 };
+    static mut PORT: u16 = 3001;
+    async fn run_hyper_integration(is_sync: bool, is_custom_type: bool) {
+        let port = unsafe {
```
LegNeato, line 406: Why this `unsafe`? I don't think this is worth adding more unsafe to the codebase but I could be convinced.

LegNeato (dismissed):

juniper_hyper/src/lib.rs
elcharitas: Never mind why it's there. I've removed it since it can be avoided.

Please check again @LegNeato

tyranron:
@elcharitas @LegNeato I'll look into it too in the next few working days.

tyranron (approved):
@elcharitas I've somewhat relaxed trait bounds. Anything else seems fine. Thanks! 🍻

PR #1277, opened 2024-08-15
Add references to known GraphQL scalars

sgoll:
Similar to #1275

## Description

This PR adds references (links) to GraphQL scalars that are well-known and documented at https://graphql-scalars.dev.

Some of these links were missing because the associated types had not been defined at the time the corresponding feature was added.


tyranron (approved):
@sgoll thanks! This one definitely showed me some stuff I hadn't noticed in https://github.com/graphql-rust/juniper/pull/1275.

PR #1289, opened 2024-10-24
Fix content-header parsing in juniper_axum

### LINKED ISSUE #1288: juniper_axum breaks on valid Content-Type Header
Given a valid POST with Content-Type header like: `application/json; charset=utf-8`.

juniper_axum responds with:

415 Unsupported Media Type

> `Content-Type` header is expected to be either `application/json` or `application/graphql`

This is because we match on the full header value instead of reading substring and matching on that:

https://github.com/graphql-rust/juniper/blob/257bc69dded08423945d848f8775df9461215a68/juniper_axum/src/extract.rs#L125-L132

Axum handles this case by using a `starts_with` (which matches expectations here: https://httpwg.org/specs/rfc9110.html#media.type)

https://github.com/tokio-rs/axum/blob/280d16a61059f57230819a79b15aa12a263e8cca/examples/parse-body-based-on-content-type/src/main.rs#L63-L76

jgrund:
Check media type using `starts_with` instead of full string matching.

Fixes #1288

### DIFF
--- a/juniper_axum/CHANGELOG.md
+++ b/juniper_axum/CHANGELOG.md
@@ -16,6 +16,10 @@ All user visible changes to `juniper_axum` crate will be documented in this file
 
 - Building on `wasm32-unknown-unknown` and `wasm32-wasi` targets. ([#1283], [#1282])
 
+### Fixed
+
+- `Content-Type` header reading full value instead of just the media type. ([#1288])
+
 [#1272]: /../../pull/1272
 [#1282]: /../../issues/1282
 [#1283]: /../../pull/1283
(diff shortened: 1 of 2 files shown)


LegNeato (changes requested):
Thank you for the great report and the PR! 🍻 

Two quick things:
1. Can you add an entry to the `CHANGELOG`?
2. Can you add a test so we don't regress this in the future? (if it is too hard / will take too much time, please feel free to comment as such!)

LegNeato (approved):
Thanks again!

jgrund:
Thank you, can we cut a new release of `juniper_axum` with this fix?

tyranron:
@jgrund thank you for the PR!

Backported and released as 0.1.1: https://crates.io/crates/juniper_axum/0.1.1

PR #1304, opened 2025-01-04
Update `axum` from 0.7 to 0.8 version (breaking)

eternal-flame-AD:
Thanks for the project!

Axum just released 0.8 and changed a couple of things, and I tried upgrading and there are a couple places in `juniper_axum` that need changing (the type of WS message and remove `::async_trait`). This will be a breaking change to your user as the type for ["extract::ws::WebSocket"] is incompatible.


tyranron:
@eternal-flame-AD released in [`juniper_axum` 0.2.0](https://github.com/graphql-rust/juniper/releases/tag/juniper_axum-v0.2.0).

PR #1324, opened 2025-05-28
Allow cheap conversion into`ScalarValue` for custom string types 

tyranron:
Part of #819\
Revealed from https://github.com/graphql-rust/juniper/pull/1247#issuecomment-2049538564

## Synopsis

> an idea: Could we somehow make `From<ArcStr>` a supertrait of `ScalarValue`? Then one could make a custom scalar value that doesn't need to clone string buffers that are already `ArcStr`. Think introspection queries, etc.
>
> (Ideally `From<&ArcStr>` (which doesn't require cloning up front), but that's not possible without making `ScalarValue` lifetime generic. Another idea is to add a method `fn from_arcstr(&ArcStr) -> Self` to `trait ScalarValue` itself.)
>
> edit: `fn from_arcstr(&ArcStr)` could be a method with a default implementation that just delegates to `From<String>`

Indeed, if an `ArcStr` is used (like in `SchemaType` types), and some custom `ScalarValue` is involved, which uses `ArcStr` internally, we want to operate on it without intermediate conversions into/from `String`.

## Solution

To allow other custom string types as well, instead of just `fn from_arcstr(&ArcStr) -> Self` method, we introduce a `fn from_displayable<Str: fmt::Display + Any + ?Sized>(s: &Str) -> Self` method defaulting to `s.to_string().into()`.

However, an author of a custom `ScalarValue` could redefine it to convert efficiently:
```rust
fn from_displayable<Str: fmt::Display + Any + ?Sized>(s: &Str) -> Self {
    use juniper::AnyExt as _;

    if let Some(arc_str) = s.downcast_ref::<ArcStr>() {
        Self::ArcStr(arc_str.clone())
    } else {
        s.to_string().into()
    }
}
```

## Checklist

- [x] Documentation is updated
- [x] Examples are provided
- [x] Tests are added
- [x] CHANGELOG is updated


audunhalland:
@tyranron I like it, it's a nice improvement.

PR #1342, opened 2025-07-29
[AIM-2066] fixing problem with alias counting

leo-vital:
When investigating why we keep having problems with the API Consumers choking when receiving request with lots of aliases, I noticed a major bug in the alias counting, which was the fact that since the number of aliases was defined as an unsigned 8-bit integer, it was overflowing whenever reaching the max value. Therefore, this logic:

```rust
            self.alias_count += 1;
            if self.alias_count > self.max_allowed {
                ctx.report_error(&error_message(&alias.item), &[alias.start]);
            }
```

did not block all aliases from being executed if there were more than 255 aliases. Once it sees the 256th alias, there is an overflow and the count goes back to 0, and therefore it allows the next 4 aliases (from 0 to 3, all are less than or equal to the `max_allowed`, which is 3).

So for 9999 aliases, instead of only allowing 3, it allows more than 100 queries to go through.

As for why the overflow was not caught:
1. No tests existed that tried to go beyond 255 aliases.
2. There is no overflow check when compiling in release mode, so in production we never got an error, the overflow simply happened without us noticing it.

But I did confirm this behavior by looking at the logs of the attack the Security Team performed on staging:

```
RuleError { locations: [SourcePosition { index: 20188, line: 254, col: 4 }], message: "Illegal number of aliases, **a253** is not allowed" },
RuleError { locations: [SourcePosition { index: 20268, line: 255, col: 4 }], message: "Illegal number of aliases, **a254** is not allowed" },
RuleError { locations: [SourcePosition { index: 20668, line: 260, col: 4 }], message: "Illegal number of aliases, **a259** is not allowed" },
```

(formatting done by me to make it easier to parse it)

Notice how a255, a256, a257 and a258 were not included in the error message.

The fix was to simply change the u8 to a u32. I imagine that there's no way to include 2**32 aliases since that would most likely go beyond what our server would even accept in the payload, although I have not really verified this, and HTTP does not by design define a limit for the payload size.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,48 @@
+## About this fork
+
+This is a fork of Juniper v0.15.9.
+
+It restricts the resolving of `__schema` and `__type` fields via the added 
+`disable_introspection` feature.
+
+### Example
+
+#### Query
+
+```graphql
+{
+    __schema {
+        types {
+            name
+        }
+    }
+}
+```
+
+#### Response, with `disable_introspection` on
+
+```json
+{
+  "data": null,
+  "errors": [
+    {
+      "message": "Introspection queries are disabled",
+      "locations": [
+        {
+          "line": 1,
+          "column": 3
+        }
+      ],
+      "path": [
+        "__schema"
+      ]
+    }
+  ]
+}
+```
+
+---
+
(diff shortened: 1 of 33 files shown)


leo-vital:
Sorry about this confusion, I did not intend to create a PR for this, only for the fork we have, my bad.

PR #1355, opened 2025-10-14
feat(integrations): add ruint

### LINKED ISSUE #1350: add optional support for ruint types
it would be nice to have support for [ruint](https://github.com/recmo/uint) types (behind a feature flag)

similar implementation to https://github.com/graphql-rust/juniper/pull/1060
happy to provide a PR for this

siosw:
closes #1350

Ruint uses const generics to set the size of the uint, I chose to include type aliases for `U64`, `U128`, and `U256` which are commonly used but i think arguments can be made to include more or less.

It would be nice to avoid repeating the doc string for these but i'm unsure if that's possible.

Also wasn't sure where to add this in the change log so left it for now.

### DIFF
--- a/juniper/Cargo.toml
+++ b/juniper/Cargo.toml
@@ -61,7 +61,7 @@ itertools = "0.14"
 jiff = { version = "0.2", features = ["std"], default-features = false, optional = true }
 juniper_codegen = { version = "0.17.0", path = "../juniper_codegen" }
 ref-cast = "1.0"
-ruint = { version = "1.2", optional = true }
+ruint = { version = "1.10", optional = true }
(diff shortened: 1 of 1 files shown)


tyranron (approved):
@siosw thank you for your effort!

Don't bother with that test, it's known to be flaky, and there is no way to make it deterministic, unfortunately, since we depend there on nondeterministic compiler output.

Also, I've made few adjustments:
- Exposed the module fed to `#[graphql_scalar]` macro as `pub`, so anyone who desires custom-sized type could just use the module directly (example for doing so is included). 
- Added implementations for `U8`, `U16` and `U32` types.

PR #1371, opened 2026-03-13
Fix errors handling in GraphQL subscriptions

50U10FCA7:
## Synopsis

`juniper_graphql_ws` doesn't allow end-user to handle panics can happen during resolvers/streams execution.

Also, current implementation of `#[graphql_subscription]` maps errors incorrectly, which causes memory leak and completely omits errors in GraphQL subscriptions:

#### Omitted errors in GraphQL subscriptions

Each subscription resolves items of its stream, [expecting the resolving can fail](https://github.com/graphql-rust/juniper/blob/4dc48130dced24759685d010211ff416c2fbf0ab/juniper_codegen/src/common/field/mod.rs#L364) which is [never happens](https://github.com/graphql-rust/juniper/blob/4dc48130dced24759685d010211ff416c2fbf0ab/juniper/src/types/async_await.rs#L108) because of the [GraphQL specification](https://github.com/graphql-rust/juniper/blob/4dc48130dced24759685d010211ff416c2fbf0ab/juniper/src/types/async_await.rs#L90). And because of this, no errors passed to GraphQL subscription leaving them inside the [`Executor`](https://github.com/graphql-rust/juniper/blob/4dc48130dced24759685d010211ff416c2fbf0ab/juniper/src/executor/mod.rs#L73).

#### Memory leak

`juniper::Executor` is shared across all the subscription's items (one executor per subscription). And its errors [shared](https://github.com/graphql-rust/juniper/blob/4dc48130dced24759685d010211ff416c2fbf0ab/juniper/src/executor/mod.rs#L73) across all the types resolution. [`Executor::errors`](https://github.com/graphql-rust/juniper/blob/4dc48130dced24759685d010211ff416c2fbf0ab/juniper/src/executor/mod.rs#L73) in current implementation [only collects errors](https://github.com/graphql-rust/juniper/blob/4dc48130dced24759685d010211ff416c2fbf0ab/juniper/src/executor/mod.rs#L655) but never `Drop`s them, which can lead to a memory leak in long-running subscription regularly emitting errors.

## Solution

- [x] Support panic handling in `juniper_graphql_ws` crate.
- [x] Fix errors mapping in `#[graphql_subscription]` macro expansion.
- [x] Fix memory leak caused by error handling in GraphQL subscriptions.
- [x] Make `juniper::ValuesStream` to support error batching.

## Checklist
- [x] Tests are added
- [x] Documentation is added (including Book)
- [x] CHANGELOG is filled


50U10FCA7:
### FCM
```markdown
Fix errors handling in GraphQL subscriptions (#1371)

- support panic handling in `juniper_graphql_ws` crate
- fix errors mapping in `#[graphql_subscription]` macro expansion in `juniper_codegen` crate
- fix memory leak caused by incorrect errors handling in GraphQL subscriptions machinery
- refactor `juniper::ValuesStream` to support error batching in `juniper` crate
```

tyranron (approved):

PR #1373, opened 2026-04-06
Fix unknown argument validation

zohnannor:
As per [GraphQL S5.4.1 Argument Names](https://spec.graphql.org/October2021/#sec-Argument-Names),

> Every argument provided to a field or directive must be defined in the set of possible arguments of that field or directive

But currently, juniper allows passing the arguments even if the field doesn't have any.

MRE: https://github.com/zohnannor/juniper-mre/tree/unknown-argument-no-error

---

This can be considered a breaking change, however, anyone relying on this was essentially abusing a bug unknowingly. Any way, should probably be mentioned in the changelog.

### DIFF
--- a/juniper/src/tests/query_tests.rs
+++ b/juniper/src/tests/query_tests.rs
@@ -1,5 +1,6 @@
 use crate::{
-    graphql,
+    GraphQLError, RuleError, graphql,
+    parser::SourcePosition,
     schema::model::RootNode,
     tests::fixtures::starwars::schema::{Database, Query},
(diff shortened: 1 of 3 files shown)


LegNeato (changes requested):
Thanks for the PR! Looks like CI is failing.

zohnannor:
You are right, I updated the old tests and added one for regression coverage.

LegNeato:
Great, thank you! Sorry this took so long.

zohnannor:
Awesome! No problem, thank you :)

PR #1376, opened 2026-04-26
default variable fixes

jbellenger:
Hi there! I'm the maintainer of [graphql-conformance](https://jbellenger.github.io/graphql-conformance/#), which verifies spec conformance across the graphql ecosystem. I noticed that juniper has a couple of deviations from the spec and I thought I might be able to contribute a fix for one of the easy ones (see [juniper's conformance page](https://jbellenger.github.io/graphql-conformance/#/impl/juniper) for other conformance gaps)

The issue addressed by this PR is that juniper rejects some allowed patterns with variables. 

For example, this query is valid and should be executable with empty variables:
```graphql
query ($var: Boolean! = true) {
  __typename @skip(if:$var)
}
```

Juniper rejects this query in 2 ways:
```
# default value is not allowed
ValidationError([RuleError {
  locations: [SourcePosition { index: 33, line: 1, col: 32 }],
  message: "Argument \"var\" has type \"Boolean!\" and is not nullable, so it can't have a default value"
}])

# value must be provided
ValidationError([RuleError {
  locations: [SourcePosition { index: 16, line: 1, col: 15 }],
  message: "Variable \"$var\" of required type \"Boolean!\" was not provided."
}])
```

Queries with this shape are allowed by these sections of the spec:
- [Section 5.8.5](https://spec.graphql.org/draft/#sel-JANRLFCFFNCAAEJDGB39c), which allows non-nullable variables to have a default (see also: [this PR](https://github.com/graphql/graphql-spec/pull/418), which explicitly describes that this is supported, "All variables can now have default values ... Previously it was invalid to supply a default value with a non-null variable.")
- [Section 6.1.2 3g](https://spec.graphql.org/draft/#sel-JANRLFCFFNCAAEJDGB39c), which allows a variable to be unspecified in a request if it has a default

Finally, the combination of a non-nullable variable with default, plus an unspecified variable, is implemented by [this graphql-js test case](https://github.com/graphql/graphql-js/blob/next/src/execution/__tests__/variables-test.ts#L714-L726).

Thank you for considering this PR. I'm not familiar with this code-base or its conventions, so please feel free to point out any issue no matter how small. I went through the checklist in CONTRIBUTING.md, everything passed except for the `make test.book` which failed for me in master.


LegNeato:
Awesome project, thank you!

PR #1378, opened 2026-06-06
Return error instead of panicking on unsupported operation type

LegNeato:
Executing a `mutation` against a schema with `EmptyMutation` (or a `subscription` with `EmptySubscription`) panics with `No mutation type found` / `No subscription type found` and aborts the request: the executor `.expect()`s a root type the schema intentionally doesn't expose, and validation can't reject the operation since there's no root type to check against.

Return a new `GraphQLError::NotSupported(OperationType)` request error instead, in the sync and async query executors and the subscription resolver. Adds regression tests.


LegNeato:
CI failures are unrelated and fixed by https://github.com/graphql-rust/juniper/pull/1379

PR #1384, opened 2026-06-30
Fix Juniper Book deploy for mdBook 0.5

LegNeato:
mdBook 0.5 makes `--dest-dir` relative to the CWD, so the book was written to `gh-pages/` (repo root) instead of `book/gh-pages` where the deploy step publishes from.
Pass `out=book/gh-pages` to fix the broken `deploy (Book)` job.


LegNeato:
Can't test this on PRs, merging.

